PR #25, opened 2019-04-22
Kanvi/fix tf python tests

kanvi-nervana:


src/grappler/ngraph_optimizer.cc
```diff
@@ -70,20 +70,38 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+  std::set<string> nodes_to_preserve;
+
+  // Feed Nodes
+  for (int i = 0; i < item.feed.size(); i++) {
+    nodes_to_preserve.insert(item.feed[i].first);
+  }
+  // Keep Ops
+  for (const string& f : item.keep_ops) {
+    nodes_to_preserve.insert(f);
+  }
```
sayantan-nervana: Minor:
`nodes_to_preserve.insert(item.keep_ops.begin(), item.keep_ops.end())`

similarly for `item.init_ops`

kanvi-nervana:
> LGTM
> 
> Can you enable some tests that pass with this change

I did not disable any tests. Without this change, about 70 tf python tests were failing from the groups bias_op_test, cast_op_test, concat_op_test, conv_ops_test, conv_ops_3d_test, nn_fused_batchnorm_test, reduction_ops_test, transpose_op_test, math_grad_test, conv3d_transpose_test.

PR #26, opened 2019-04-22
Sindhu/encap compute refactor

sindhu-nervana:


src/ngraph_encapsulate_op.cc
```diff
@@ -258,28 +258,10 @@ class NGraphEncapsulateOp : public OpKernel {
...
-    NGRAPH_VLOG(4) << "NGraphEncapsulateOp::Compute starting for cluster "
-                   << m_ngraph_cluster;
-
-    NGRAPH_VLOG(4) << "Got backend of type: " << m_op_backend_name;
-    ng::runtime::Backend* op_backend =
-        BackendManager::GetBackend(m_op_backend_name);
-
-    ngraph::Event event_func_maybe_create("FunctionMaybeCreate", name(), "");
-    Timer function_lookup_or_create;
+  Status ComputeSignature(OpKernelContext* ctx, std::stringstream& signature_ss,
```
sayantan-nervana: `ComputeSignature` vs `get_ng_exec`. I think we use `snake_case` in python and `UpperCamelCase` in cpp. Can you please also rename `get_current_ng_tensor` to `UpperCamelCase`.

src/ngraph_encapsulate_op.cc
```diff
@@ -291,20 +273,36 @@ class NGraphEncapsulateOp : public OpKernel {
...
+                     std::vector<const Tensor*>& static_input_map) {
+    std::stringstream signature_ss;
+    string signature;
 
     std::shared_ptr<ngraph::Function> ng_function;
-    std::shared_ptr<ngraph::runtime::Executable> ng_exec;
     std::shared_ptr<ngraph::runtime::Executable> evicted_ng_exec;
-    std::string signature = signature_ss.str();
+
+    NGRAPH_VLOG(4) << "get_ng_exec:Got backend of type: " << m_op_backend_name;
```
sayantan-nervana: Please rename `get_ng_exec` (`UpperCamelCase`).

Also space after colon

src/ngraph_encapsulate_op.cc
```diff
@@ -443,6 +437,37 @@ class NGraphEncapsulateOp : public OpKernel {
...
+
+    ngraph::Event event_func_maybe_create("FunctionMaybeCreate", name(), "");
+    Timer function_lookup_or_create;
+
+    std::vector<TensorShape> input_shapes;
+    std::vector<const Tensor*> static_input_map;
+    std::shared_ptr<ngraph::Function> ng_function;
+    std::shared_ptr<ngraph::runtime::Executable> ng_exec;
+
+    // Get ngraph executable
```
sayantan-nervana:
```suggestion
    // Get nGraph executable and inputs information
```

src/ngraph_encapsulate_op.cc
```diff
@@ -291,20 +273,36 @@ class NGraphEncapsulateOp : public OpKernel {
...
+    std::stringstream signature_ss;
+    string signature;
 
     std::shared_ptr<ngraph::Function> ng_function;
-    std::shared_ptr<ngraph::runtime::Executable> ng_exec;
     std::shared_ptr<ngraph::runtime::Executable> evicted_ng_exec;
-    std::string signature = signature_ss.str();
+
+    NGRAPH_VLOG(4) << "GetNgExec: Got backend of type: " << m_op_backend_name;
+    ng::runtime::Backend* op_backend =
```
shresthamalik: Please pass the backend pointer also. We have already retrieved the backend pointer from the backend manager in the compute function. It can be passed to GetNgExec() function.

src/ngraph_encapsulate_op.cc
```diff
@@ -443,6 +437,37 @@ class NGraphEncapsulateOp : public OpKernel {
...
+    ngraph::Event event_func_maybe_create("FunctionMaybeCreate", name(), "");
+    Timer function_lookup_or_create;
+
+    std::vector<TensorShape> input_shapes;
+    std::vector<const Tensor*> static_input_map;
+    std::shared_ptr<ngraph::Function> ng_function;
+    std::shared_ptr<ngraph::runtime::Executable> ng_exec;
+
+    // Get ngraph executable and inputs information
+    OP_REQUIRES_OK(ctx,
```
shresthamalik: Add log for this function call

sindhu-nervana: I am using `OP_REQUIRES_OK` and not `OP_REQUIRES`, so it will already print logs from the function. Is that enough?
Or should i do try catch and add logs if this function returns bad status?

shresthamalik: I meant just a log like "Getting executable " in lines of this,
https://github.com/NervanaSystems/ngraph-tf/blob/master/src/ngraph_encapsulate_op.cc#L467

To signal where in the compute function we are.

src/ngraph_encapsulate_op.cc
```diff
@@ -291,20 +273,36 @@ class NGraphEncapsulateOp : public OpKernel {
...
       }
     }
+    return Status::OK();
+  }
+
+  Status GetNgExec(OpKernelContext* ctx,
+                   std::shared_ptr<ngraph::runtime::Executable>& ng_exec,
+                   std::vector<TensorShape>& input_shapes,
+                   std::vector<const Tensor*>& static_input_map,
+                   ng::runtime::Backend*& op_backend) {
```
shresthamalik: @avijit-nervana Do you have any thoughts around reference to pointer?

sayantan-nervana: or non-const references generally. I am ok with them.

avijit-nervana: Yes - this should be fine. Thanks for bringing this up.

PR #30, opened 2019-04-24
Sindhu/tf ut runner enhancement

sindhu-nervana:


sindhu-nervana:
This is the new failure output we see! 

Screen Shot 2019-04-24 at 3 20 47 PM

sindhu-nervana:
1. Added verbose option which will print standard out only if set
2. Tqdm - show a progress bar while running tests
3. Pass and Fail summary will be printed only if verbose is set
4. Print failed tests and failure message. 
5. Print total number of tests ran along in the end if verbose is not set.

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -280,35 +305,55 @@ def print_and_check_results(test_result, invalid_list):
...
+            if len(test_case_name.keys()) > 0:
+                if key is "PASSED":
+                    print('Number of tests ' + key + ' ' +
+                          str(len(test_case_name.keys())))
+                if key is "FAILED":
+                    print('Number of tests ' + '\033[91m' + key + '\033[0m' +
+                          ' ' + str(len(test_case_name.keys())))
+                if key is "ERRORS":
+                    print('Number of tests ' + '\033[33m' + key + '\033[0m' +
+                          ' ' + str(len(test_case_name.keys())))
```
sayantan-nervana: lines 346 to 355 can be replaced by:
```
colour = {'PASSED':'\033[98m', 'FAILED:'\033[91m', 'ERRORS':'\033[33m'}[key]
print('Number of tests ' + colour + key + ' ' + '\033[0m' + str(len(test_case_name.keys())))
```

PR #32, opened 2019-04-25
add json file and code change in verify_inference_model.py to use thi…

chliny2016:
…s json file and use args input of json file


chliny2016:
need to login using intel e-mail to create PR and create PR

PR #33, opened 2019-04-25
Sarkars/merge utils

sayantan-nervana:


src/ngraph_utils.cc
```diff
@@ -33,15 +34,67 @@
...
+#include "version.h"
+
 using namespace std;
 namespace ng = ngraph;
 
 namespace tensorflow {
 
 namespace ngraph_bridge {
 
+Status IsCopyLogEnabled(int graph_id, bool& is_copy_log_enabled) {
```
avijit-nervana: Add some comments to illustrate what this function does. The name of this function is a bit ambiguous including the env. variable name. "Copy what?" is the question comes to ming. Please consider renaming. 
For example the word `log` can be replaced by `logging`. Also, `copy` may be replaced by `tensor_copy`?

Also, why do we need an env. variable? What's the impact of this being always done?

src/ngraph_utils.cc
```diff
@@ -33,15 +34,67 @@
...
+
 bool IsNGVariableType(string node_type) {
-  return node_type == "NGraphVariable";
+  if (ngraph_tf_are_variables_enabled())
+    return (node_type == "NGraphVariable" || node_type == "NGraphAssign");
+  else
+    return node_type == "NGraphVariable";
+}
+
+bool IsNGSupportedType(string node_type) {
```
avijit-nervana: Suggest `IsSupportedNGNodeType`

src/ngraph_utils.cc
shresthamalik: @avijit-nervana  This function logs the number of copies done as well as the reason for it, during the computation of Encapsulate, Variable and Assign ops. The reason could by syncing, input from TF op, etc.

The log is in form of cout in lines of the environment variable NGRAPH_TF_LOG_PLACEMENT. There is no impact other than some extra time for couts.

avijit-nervana: In that case why do we need the env variable `NGRAPH_TF_LOG_COPIES`?

shresthamalik: Sorry, this is a utility function to check whether the environment variable NGRAPH_TF_LOG_COPIES is set for the graph that is currently being processed.  For e.g. we can print only the copies in the main Training Graph,etc.
If it is set, then those logs will be cout.

avijit-nervana: Sounds good

shresthamalik: Name Change Suggestions: @avijit-nervana @sayantan-nervana 
1. Env Variable : NGRAPH_TF_LOG_TENSOR_COPIES
2. Function Name: IsNgraphTFLogTensorCopiesEnabled

shresthamalik: I prefer logging over log, to be in line with NGRAPH_TF_LOG_PLACEMENT

src/ngraph_utils.cc
shresthamalik: This we can revisit on Monday, as we have to figure out a corresponding name for IsNGVariableType.

PR #34, opened 2019-04-26
Sarkars/cm vect to map

sayantan-nervana:
Convert clustermamager's main data structure to map (from vector)


sayantan-nervana:
`GetClusterIndexes` is not needed for now

PR #36, opened 2019-04-26
Kanvi/capture pass checks

kanvi-nervana:
In Capture pass, only capture the Assign, and other modifiers if they operate on a Variable Op. Otherwise reject it.

Similar check needs to be done for the Attribute : Validate_shape. Capture only if True. We dont expect the Variable Tensor to change shape.


src/enable_variable_ops/ngraph_replace_op_utilities.cc
```diff
@@ -216,6 +216,44 @@ Status ReplaceOutputEdges(Graph* graph, Node* node, Node* replacement) {
...
+        // So clear the ref list and return.
+        ref_list->clear();
+        return Status::OK();
+      } else {
+        // In all other cases:
+        // 1. Assign, variable_shape = true
+        // 2. AssignAdd
+        // 3. AssignSub
+        // 4. ApplyGradientDescent
+        // which are all ref types, add to the ref list if not already added
```
shresthamalik: It seems all the outputs that are of RefType are added and not just the 4 ops that we capture

sayantan-nervana: We need a check if its one of the 4 types mentioned above. This (List of assign type ops that ng-bridge supports) should be a function/table in utils or somewhere, so that we can keep adding to it.

kanvi-nervana: Yes, we would add all the outputs that are of RefType but only the 4 ops will be captured that are present in the CAPTURE_REPLACE_OP_MAP, but I could also add a check to add only those 4 to the ref list.

sayantan-nervana: I see, so its functionally correct now i guess.

Also `nodes_to_capture` is actually `possible_nodes_to_capture`, right? (because its a necessary but not sufficient condition for a node to be captured is that it ends up in `nodes_to_capture`/`possible_nodes_to_capture`)

src/enable_variable_ops/ngraph_replace_op_utilities.cc
```diff
@@ -216,6 +216,44 @@ Status ReplaceOutputEdges(Graph* graph, Node* node, Node* replacement) {
...
+        // not already added part is taken care of by the list being a set.
+        NGRAPH_VLOG(4) << "Adding " << dst->name() << " to the ref list";
+        ref_list->insert(dst);
+      }
+    }
+    // Recursively go over the outputs of each node.
+    StoreRefTypeOutputs(dst, ref_list);
```
sayantan-nervana: recursion may fail if the graph has cycles (eg, through next iteration etc)

maybe when traversing output edges, only traverse data edges and not control edges. I forget if `node->out_edges()` gives all edges or data edges only

src/enable_variable_ops/ngraph_replace_op_utilities.cc
shresthamalik: The check for being present in CAPTURE_REPLACE_OP_MAP is missing in the main function also

src/enable_variable_ops/ngraph_capture_variables.cc
```diff
@@ -57,34 +57,45 @@ Status CaptureVariables(Graph* graph, std::set<string> skip_these_nodes) {
...
+      NGRAPH_VLOG(4) << "Found Variable: " << node->name();
+      // Add the Variable node to the ref list
+      ref_list.insert(node);
+
+      // go over all the nodes leading from VariableV2 and store them
+      // in a list if they are ref type
+      StoreRefTypeOutputs(node, &ref_list);
+
+      if (ref_list.size()) {
+        nodes_to_capture.insert(nodes_to_capture.end(), ref_list.begin(),
```
shresthamalik: we can add the check here. Enter in nodes_to_capture only if type exists in Capture_OP_Map

src/enable_variable_ops/ngraph_replace_op_utilities.cc
shresthamalik: We should recurse only when the dst is reftype. The recursive call needs to be inside the if_ref check

test/graph_rewrites/capture_variables.cc
```diff
@@ -0,0 +1,185 @@
...
+      ASSERT_NE("NGraphVariable", node->type_string());
+    else if (node_name == "AssignX")
+      ASSERT_EQ("NGraphAssign", node->type_string());
+    else if (node_name == "AssignY")
+      ASSERT_NE("NGraphAssign", node->type_string());
+  }
+}
+
+// Test that an Assign (specifically Assign1) with attribute
+// variable_scope = false and also the Variable is shared with
```
shresthamalik: Typo validate_shape

test/test_variable_ops.cpp
```diff
@@ -119,11 +119,11 @@ TEST(VariableTest, SmallGraph2) {
...
   auto add = ops::Add(root.WithOpName("Add1"), var, c);
 
-  auto assign = ops::AssignAdd(root, var, add);
+  auto assign_add = ops::AssignAdd(root.WithOpName("AssignAdd"), var, add);
 
-  auto add2 = ops::Add(root.WithOpName("Add2"), assign, c);
+  auto add2 = ops::Add(root.WithOpName("Add2"), assign_add, c);
 
-  auto assign2 = ops::AssignSub(root, var, add2);
+  auto assign_sub = ops::AssignSub(root.WithOpName("AssignSub"), var, add2);
```
shresthamalik: Are we using the op names anywhere? Only some ops (Variable related) have names others don't.

src/enable_variable_ops/ngraph_capture_variables.cc
```diff
@@ -57,34 +57,45 @@ Status CaptureVariables(Graph* graph, std::set<string> skip_these_nodes) {
           {"AssignSub", std::make_pair("NGraphAssignSub", ReplaceAssign)},
           {"VariableV2", std::make_pair("NGraphVariable", ReplaceVariable)}};
 
-  std::vector<Node*> replaced_nodes;
+  std::set<Node*> ref_list;
```
shresthamalik: We can move this declaration inside the for loop, as it is only used within

src/enable_variable_ops/ngraph_replace_op_utilities.cc
```diff
@@ -224,32 +224,41 @@ bool IsValidateShape(Node* node) {
...
+          // 1. Assign, validate_shape = true
+          // 2. AssignAdd
+          // 3. AssignSub
+          // 4. ApplyGradientDescent
+          // 5. other ref types
+          // which are all ref types, add to the ref list if not already added
+          if (ref_list->find(dst) == ref_list->end()) {
+            NGRAPH_VLOG(4) << "Adding " << dst->name() << " to the ref list";
+            ref_list->insert(dst);
+          } else {
```
kanvi-nervana: Not sure if we need this log message here, thoughts?

test/test_variable_ops.cpp
kanvi-nervana: We are not, but I found it handy while debugging.

test/graph_rewrites/capture_variables.cc
```diff
@@ -0,0 +1,185 @@
...
+    auto node_name = node->name();
+    if (node_name == "Var")
+      ASSERT_NE("NGraphVariable", node->type_string());
+    else if (node_name == "Assign1")
+      ASSERT_NE("NGraphAssign", node->type_string());
+    else if (node_name == "Assign2")
+      ASSERT_NE("NGraphAssign", node->type_string());
+  }
+}
+
```
shresthamalik: Please add the TempVar Test as discussed

test/test_variable_ops.cpp
```diff
@@ -359,6 +363,149 @@ TEST(VariableTest, SmallGraph4) {
...
+
+  auto attr_validate_shape = ops::Assign::Attrs().ValidateShape(false);
+  auto var_assign = ops::Assign(root.WithOpName("Var_Assign"), var, init_value,
+                                attr_validate_shape);
+
+  auto c = ops::Const(root, {{1.f, 1.f}, {1.f, 1.f}});
+
+  auto add = ops::Add(root.WithOpName("Add"), var, c);
+
+  // auto assign = ops::Assign(root, var, add);
```
sayantan-nervana: stray comment

test/test_variable_ops.cpp
```diff
@@ -359,6 +363,149 @@ TEST(VariableTest, SmallGraph4) {
...
+  std::vector<tensorflow::Tensor> ng_outputs1;
+  std::vector<tensorflow::Tensor> ng_outputs2;
+  std::vector<tensorflow::Tensor> ng_outputs3;
+
+  ASSERT_OK(ng_session.Run(
+      {
+          var_assign,
+      },
+      &ng_outputs1));
+  for (int i = 0; i < 20; i++) {
```
sayantan-nervana: Why are we looping till 20?

PR #45, opened 2019-04-29
add json file and modify verify_inference_model.py to use this json f…

chliny2016:
…ile, and enhance re and exit part


diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -81,31 +58,25 @@ def run_inference(model_name, models_dir):
                     os.chdir(models_dir + "research/slim")
                 command_executor("export PYTHONPATH=$PYTHONPATH:`pwd`:`pwd`")
                 command_executor('git apply ' + pwd +
-                                 '/image_recognition.patch')
+                                 '/iage_recognition.patch')
```
sindhu-nervana:
```suggestion
                                 '/image_recognition.patch')
```

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -134,6 +105,12 @@ def check_accuracy(model, p, tolerance=0.001):
         help=
         'Model name to run inference. Availble models are Inception_v4, MobileNet_v1, ResNet50_v1',
         required=True)
+
+    parser.add_argument(
+        '--json_file_name',
+         help=
+         'json file to use for run inference and check acucuracy',
```
sindhu-nervana: Maybe we could modify --help here to `json file with model parameters to run inference and accuracy values to verify`

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -154,7 +131,11 @@ def check_accuracy(model, p, tolerance=0.001):
     #TODO(Sindhu): Run multiple or ALL models at once and compare accuracy.
 
     try:
-        model_name, p = run_inference(args.model_name, models_dir)
-        check_accuracy(model_name, p)
+        print(args.json_file_name)
```
sindhu-nervana: why are we printing json file name here?

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: Thanks @sindhu-nervana, good catch, I will change it.

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: Thanks @sindhu-nervana, good catch, I will change it.

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: @sindhu-nervana, there isn't this line "print(args.json_file_name)" there, would you pls double check.

chliny2016: @sindhu-nervana , sorry, I saw the line, let me delete it.

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -154,7 +131,10 @@ def check_accuracy(model, p, tolerance=0.001):
     #TODO(Sindhu): Run multiple or ALL models at once and compare accuracy.
 
     try:
-        model_name, p = run_inference(args.model_name, models_dir)
-        check_accuracy(model_name, p)
+        model_name, p = run_inference(args.model_name, models_dir,args.json_file_name)
+        check_accuracy(model_name, p,args.json_file_name)
+        sys.exit(0)
```
crlishka: sys.exit(0) should only occur if check_accuracy() has found that all models pass.  check_accuracy() returns True or False based on whether the models passed or failed.  Therefore, the code here should be:

if check_accuracy(model_name, p,args.json_file_name) :  sys.exit(0)
else:  sys.exit(1)

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -84,28 +61,22 @@ def run_inference(model_name, models_dir):
...
-
-    data = json.loads(accuracy)
 
+    data = parse_json(args.json_file_name)
     for line in p.splitlines():
         print(line.decode())
         if ('eval/Accuracy'.encode() in line):
             accuracy = re.split("eval/Accuracy", line.decode())[1]
-            top1_accuracy = re.search(r'\[(.*)\]', accuracy).group(1)
+            top1_accuracy = re.search(r'\[([0-9.]+)]', accuracy).group(1)
```
crlishka: It is unwise to rely on a split when the regular expression can be better tuned to search for more exact match.  In addition, the group(1) call above will fail with a confusing error message ("IndexError: no such group") if there are no groups found.

It is better to rewrite the two lines:
    accuracy = re.split("eval/Accuracy", line.decode())[1]
    top1_accuracy = re.search(r'\[([0-9.]+)]', accuracy).group(1)

into something like:
    result = re.search(r'eval/Accuracy\[([0-9.]+)]', line.decode())
    if len(result.groups()) > 0 :  top1_accuracy = result.group(1)
    else :  raise Exception('top1_accuracy not found in output of line: %s" % line)

(Note that I wrote the replacement code above off the cuff, so you may need to test the regex and tune it.)

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -84,28 +61,22 @@ def run_inference(model_name, models_dir):
...
         #for now we just validate top 1 accuracy, but calculating top5 anyway.
         if ('eval/Recall_5'.encode() in line):
             accuracy = re.split("eval/Recall_5", line.decode())[1]
-            top5_accuracy = float(re.search("\[(.*?)\]", accuracy).group(1))
+            top5_accuracy = re.search(r'\[([0-9.]+)]', accuracy).group(1)
```
crlishka: See above comment for how to better rewrite the re.search and drop the re.split

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -39,40 +39,17 @@ def download_repo(repo, target_name=None, version='master'):
...
-                --pipeline_config_path=object_detection/samples/configs/ssd_mobilenet_v1_coco.config \
-                --checkpoint_dir=/nfs/site/disks/aipg_trained_dataset/ngraph_tensorflow/fully_trained/ssd_mobilenet_v1_coco_2018_01_28/ \
-                --run_once=True"}]'
+def run_inference(model_name, models_dir,json_file_name):
+    data = parse_json(args.json_file_name)
+    
 
     try:
-        data = json.loads(parameters)
+       data
```
crlishka: In the except: block below, you want to add a sys.exit(1), since not having a valid JSON prevents the rest of the script from working properly.

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: @crlishka Thanks Chris. Changed.

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: @crlishka Thanks Chris. Changed.

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: @crlishka thanks Chris.
Changed to:
is_match = re.search(r'eval/Accuracy\[([0-9.]+)]', line.decode())
            if is_match and len(is_match.groups()) > 0 :
                top1_accuracy = is_match.group(1)

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: @crlishka thanks Chris.
Changed to:
is_match = re.search(r'.+eval/Recall_5[([0-9.]+)]', line.decode())
if is_match and len(is_match.groups()) > 0 :
top5_accuracy = is_match.group(1)

diagnostics/model_accuracy/model-run.json
```diff
@@ -0,0 +1,16 @@
+[
```
crlishka: What happened to entries for models "MobileNet_v1" and "SSD-MobileNet_v1"?  They were present in the original JSON list (in the python script), but are missing in this external JSON file.

chliny2016: will add them today :)

chliny2016:
@crlishka, @sindhu-nervana code change is done, test result at https://aipg-rancher.intel.com/jenkins/algo/job/tf-ng-bridge-run-accuracy/15/.
dashboard result at https://exp-avus.apps1-fm-int.icloud.intel.com/?what=benchmark_accuracy&draw=1&response_type=&is_filtered=filtered&orderby=test_start_time&selectedResultId=&datepicker_benchmark_accuracy=05%2F01%2F2019&results_interval=last_four_days&name_framework%5B%5D=2&name_frontend%5B%5D=6#tabs_tests_results, timestamp at 05-01 03:50, 03:08, 02:43

diagnostics/model_accuracy/model-run.json
```diff
@@ -0,0 +1,29 @@
...
+
+  {
+    "model_name": "ResNet50_v1",
+    "accuracy" : "0.752",
+    "model_type": "Image Recognition",
+    "cmd": "OMP_NUM_THREADS=28 KMP_AFFINITY=granularity=fine,compact,1,0, python eval_image_classifier.py --alsologtostderr --checkpoint_path=/nfs/fm/disks/aipg_trained_dataset/ngraph_tensorflow/fully_trained/resnet50_v1_slim/resnet_v1_50.ckpt --dataset_dir=/mnt/data/TF_ImageNet_latest --dataset_name=imagenet --dataset_split_name=validation --model_name=resnet_v1_50 --labels_offset=1"
+  },
+
+  {
+    "model_name": "SSD-MobileNet_v1",
```
crlishka: There is no "accuracy" for the SSD-MobileNet_v1 model, but there wasn't in the original script either.  @avijit-nervana and @sindhu-nervana : how do you want to handle this?  Should an accuracy be added, or the SSD-MobileNet_v1 model be removed from the JSON file, or simply leave as is (which may cause odd exceptions when the script is run with SSD-MobileNet_v1)?

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -39,40 +39,17 @@ def download_repo(repo, target_name=None, version='master'):
...
-                python object_detection/model_main.py --logtostderr \
-                --pipeline_config_path=object_detection/samples/configs/ssd_mobilenet_v1_coco.config \
-                --checkpoint_dir=/nfs/site/disks/aipg_trained_dataset/ngraph_tensorflow/fully_trained/ssd_mobilenet_v1_coco_2018_01_28/ \
-                --run_once=True"}]'
+def run_inference(model_name, models_dir, json_file_name):
+    data = parse_json(args.json_file_name)
 
     try:
-        data = json.loads(parameters)
+        data
```
crlishka: What is the reason for just testing "data" here in the try block, and not testing "data = parse_json(args.json_file_name)"?  The latter is easier to understand, and matches the original code more closely.

chliny2016: @crlishka . I was following sindhu's original sequence of lines: replace "parameters" with "data = parse_json(args.json_file_name)", and then when it comes to try block, I realized that look like there I can only put something like "data". Yes, agree with you that your recommendation is better. So I will do this: move the line "data = parse_json(args.json_file_name)" to the position of "data", and delete "data" there. and re-do the test.

diagnostics/model_accuracy/model-run.json
sindhu-nervana: @crlishka I was not able to run SSD-MobileNet successfully when i developed this script, so did not mention the model name in the list of available ones in the help here https://github.com/tensorflow/ngraph-bridge/blob/master/diagnostics/model_accuracy/verify_inference_model.py#L132. It needs to be debugged further as we continue to enable more models. Also, something to keep in mind we might have to keep updating the regex and any other dependencies/patches to apply in this script as we enable more models. You could add accuracy and keep it in json for now or remove it entirely, really up to you.

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -87,25 +64,23 @@ def run_inference(model_name, models_dir):
...
     #check if the accuracy of the model inference matches with the published numbers
     #Accuracy values picked up from here https://github.com/tensorflow/models/tree/master/research/slim
-    accuracy = \
-    '[{"model_name" : "Inception_v4", "accuracy" : "0.802"},\
-     {"model_name" : "ResNet50_v1", "accuracy" : "0.752"}, \
-     {"model_name" : "MobileNet_v1", "accuracy" : "0.709"}]'
-
-    data = json.loads(accuracy)
 
+    data = parse_json(args.json_file_name)
```
sindhu-nervana: We could just return the `data` which we already parse in line#45 from `run_inference` function and use it here, instead of parsing it here again in `check_accuracy` function.

diagnostics/model_accuracy/verify_inference_model.py
```diff
@@ -39,40 +39,17 @@ def download_repo(repo, target_name=None, version='master'):
...
-                python object_detection/model_main.py --logtostderr \
-                --pipeline_config_path=object_detection/samples/configs/ssd_mobilenet_v1_coco.config \
-                --checkpoint_dir=/nfs/site/disks/aipg_trained_dataset/ngraph_tensorflow/fully_trained/ssd_mobilenet_v1_coco_2018_01_28/ \
-                --run_once=True"}]'
+def run_inference(model_name, models_dir, json_file_name):
 
     try:
-        data = json.loads(parameters)
+        data = parse_json(args.json_file_name)
+
```
crlishka: I checked the code-format issue, and it looks like you added an line with only spaces (line 46.  You should probably just delete the empty line 46, since it wasn’t in original code.

diagnostics/model_accuracy/model-run.json
crlishka: To keep things simple and remove the SSD-MobileNet model then.  If it is not running, and we do not have an accuracy target, it is probably better to remove it.

diagnostics/model_accuracy/verify_inference_model.py
chliny2016: @crlishka , thanks. Deleted this empty line.

diagnostics/model_accuracy/model-run.json
chliny2016: @crlishka , @sindhu-nervana , I'm removing the entire entry ofSSD-MobileNet  in the json file.

chliny2016:
@crlishka @sindhu-nervana new test log at https://aipg-rancher.intel.com/jenkins/algo/job/tf-ng-bridge-run-accuracy/17/consoleFull, 

Dashboard result at https://exp-avus.apps1-fm-int.icloud.intel.com/?what=benchmark_accuracy&draw=1&response_type=&is_filtered=filtered&orderby=test_start_time&selectedResultId=&datepicker_benchmark_accuracy=05%2F02%2F2019&results_interval=last_four_days#tabs_tests_results, timestamp are 05-02 20:29, 20:04. 
FYI that due to inception4 model run got segmentation fault, so it does not post to dashboard.

Pls re-review.
Thanks.

diagnostics/model_accuracy/verify_inference_model.py
sindhu-nervana: This can be done in a later PR.

chliny2016:
Last night I modified README.md .  would you pls quickly re-review it. jenkins job is at https://aipg-rancher.intel.com/jenkins/algo/job/tf-ng-bridge-run-accuracy/18//consoleFull. Dashboard result is https://exp-avus.apps1-fm-int.icloud.intel.com/?what=benchmark_accuracy&draw=1&response_type=&is_filtered=filtered&orderby=test_start_time&selectedResultId=&datepicker_benchmark_accuracy=05%2F03%2F2019&results_interval=last_four_days#tabs_tests_results,  timestamp is 05-03 07:56, 07:15, 06:50.
Thanks.

PR #47, opened 2019-05-01
Sindhu/Compute - more refactor

sindhu-nervana:


src/ngraph_encapsulate_op.cc
```diff
@@ -900,6 +924,9 @@ class NGraphEncapsulateOp : public OpKernel {
   NgFunctionIOCache m_ng_exec_input_cache_map;
   NgFunctionIOCache m_ng_exec_output_cache_map;
 
+  int number_of_copies = 0;
+  std::stringstream copy_log_str;
```
shresthamalik: initialize to empty string, ""

src/ngraph_encapsulate_op.cc
```diff
@@ -460,78 +460,19 @@ class NGraphEncapsulateOp : public OpKernel {
...
-    NGRAPH_VLOG(4)
-        << "NGraphEncapsulateOp::Compute got freshness tracker for cluster "
-        << m_ngraph_cluster;
-
+  Status AllocateTensorInput(
+      OpKernelContext* ctx,
+      std::shared_ptr<ngraph::runtime::Executable>& ng_exec,
+      std::vector<TensorShape>& input_shapes, ng::runtime::Backend* op_backend,
+      vector<shared_ptr<ng::runtime::Tensor>>& ng_inputs,
+      std::vector<std::unique_ptr<ngraph::Event>>& input_copy_events) {
```
sayantan-nervana: I think that `ng_exec`, `input_shapes`, `op_backend` and `ng_inputs` are the essential inputs to this function.
`input_copy_events` just happens to be the tracing instrumentation. in that case we should just time `AllocateTensorInput` outside at its call site.

```
start_timer(input_copy_events)
AllocateTensorInput()
stop_timer(input_copy_events)
```

instead of:
```
AllocateTensorInput(input_copy_events)
```

Same comment for `AllocateTensorOutput`

Comments @avijit-nervana ?

src/ngraph_encapsulate_op.cc
```diff
@@ -587,37 +528,27 @@ class NGraphEncapsulateOp : public OpKernel {
...
-    ngraph::Event event_alloc_output("Output: maybe create", name(), "");
-    vector<shared_ptr<ng::runtime::Tensor>> ng_outputs;
-    int ng_output_tensor_size_in_bytes = 0;
-    std::vector<std::pair<void*, std::shared_ptr<ng::runtime::Tensor>>>&
-        output_caches = m_ng_exec_output_cache_map[ng_exec];
+  Status AllocateTensorOutput(
+      OpKernelContext* ctx,
+      std::shared_ptr<ngraph::runtime::Executable>& ng_exec,
+      std::vector<TensorShape>& input_shapes, ng::runtime::Backend* op_backend,
+      vector<shared_ptr<ng::runtime::Tensor>>& ng_inputs,
```
sayantan-nervana: Right now, `AllocateTensorOutput` accepts `ng_inputs` and `ng_outputs`. More refactor needed. `ng_inputs` is being populated in the `#if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)` section. This should be moved to `AllocateTensorInput` in that case

src/ngraph_encapsulate_op.cc
```diff
@@ -715,6 +639,106 @@ class NGraphEncapsulateOp : public OpKernel {
...
+    vector<shared_ptr<ng::runtime::Tensor>> ng_outputs;
+    int ng_output_tensor_size_in_bytes = 0;
+
+    // Allocate tensors for the output results.
+    ngraph::Event event_alloc_output("Output: maybe create", name(), "");
+    std::vector<std::pair<void*, std::shared_ptr<ng::runtime::Tensor>>>
+        output_caches;
+    OP_REQUIRES_OK(ctx,
+                   AllocateTensorOutput(ctx, ng_exec, input_shapes, op_backend,
+                                        ng_inputs, ng_outputs, output_caches));
```
shresthamalik: Not sure why we need to pass the output_caches. The function signature for AllocateTensorInputs and AllocateTensorOutputs should be same/similar, i would think

sindhu-nervana: @shresthamalik returning output caches because it is used here https://github.com/tensorflow/ngraph-bridge/blob/master/src/ngraph_encapsulate_op.cc#L774

src/ngraph_encapsulate_op.cc
sindhu-nervana: I removed input_copy_events as input to this function and moved the for loop inside the function as @shresthamalik suggested instead of doing it in Compute.

src/ngraph_encapsulate_op.cc
```diff
@@ -661,60 +643,100 @@ class NGraphEncapsulateOp : public OpKernel {
...
+    NGRAPH_VLOG(4)
+        << "NGraphEncapsulateOp::Compute got freshness tracker for cluster "
+        << m_ngraph_cluster;
+
+#if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
+    bool log_copies = false;
+    OP_REQUIRES_OK(ctx, IsCopyLogEnabled(m_graph_id, log_copies));
+    copy_log_str << "KERNEL[" << type_string() << "]: " << name()
+                 << " ,GraphID " << m_graph_id << "\n";
 #endif
```
shresthamalik: Once this PR goes in , https://github.com/tensorflow/ngraph-bridge/pull/33, we can remove the #if defined around this log. We will have less if-defs in compute

PR #48, opened 2019-05-02
sarkars/fix ng tensor dumping

sayantan-nervana:


src/ngraph_utils.h
```diff
@@ -184,8 +184,61 @@ void print_node_histogram(const std::unordered_map<string, int>&,
...
+  bool T_is_integral = std::is_integral<T>::value;
+  bool type_is_integral = type.is_integral();
+  if (type_is_integral != T_is_integral) {
+    std::stringstream err_msg;
+    err_msg << "Tensor type " << type << " is"
+            << (type_is_integral ? " " : " not ")
+            << "integral but passed template is"
+            << (T_is_integral ? " " : " not ") << "integral";
+    throw std::invalid_argument(err_msg.str());
+  }
```
sayantan-nervana: Added some checks to prevent accidental foot-shooting in some cases, but not perfectly type safe

shresthamalik: It might be more useful if we do not template-ize DumpNGTensor. We can map the ng-element-type to int/float etc inside it and Call GetScalarFromTensor with that type. That way we remove the burden from the user and it is easier to use it as a utility in a loop. For eg. encapsulate has many outputs each can be of a different type. User can directly call this in loop. rather than type checking before calling

We can add it as a todo here, if this is beyond the scope of this PR. Otherwise looks good to me

sayantan-nervana: Yes, I thought of that. but could not think of any easy way of converting from ng::Type to cpp types.
One solution would be sort of a manually specified map (like `MakeConstOp` in `ngraph_builder.cc`), but for that we'd need to know the map of all ng element types to cpp types. Will add a TODO.

Atleast now, users will be made aware by the compiler that they need to supply a cpp datatype and they'd presumably take care of that instead of `DumpNgTensor` running but failing silently.

shresthamalik: Yes, no easy way

PR #53, opened 2019-05-03
Shrestha/cpp example loading data to device

shresthamalik:


src/ngraph_catalog.h
```diff
@@ -35,8 +36,39 @@ namespace tensorflow {
 
 namespace ngraph_bridge {
 
+// TODO(malikshr, sarkars) : Create a template class "TMap" for map that can
+// take arbitary key and value types
+// With Functions Add, Get, Exists etc
+// Catalog will be a collection of objects of sub-classes that implement this
+// class (TMap)
+
```
shresthamalik: Added @sayantan-nervana 's name as he had some WIP on this

src/ngraph_encapsulate_op.cc
```diff
@@ -171,6 +170,11 @@ class NGraphEncapsulateOp : public OpKernel {
                    ctx->GetAttr<string>("_ngraph_backend", &m_op_backend_name));
     NGRAPH_VLOG(4) << "NGraphEncapsulateOp::Create backend " << def().name();
     BackendManager::CreateBackend(m_op_backend_name);
+    m_number_inputs = num_inputs();
+    m_number_outputs = num_outputs();
+    NGRAPH_VLOG(4) << "No of inputs " << m_number_inputs << " ,No of outputs "
```
sayantan-nervana:
```suggestion
    NGRAPH_VLOG(4) << "No of inputs " << m_number_inputs << ", No of outputs "
```

src/ngraph_encapsulate_op.cc
```diff
@@ -543,9 +567,23 @@ class NGraphEncapsulateOp : public OpKernel {
...
         continue;
       }
-
       NGRAPH_VLOG(4) << "NGraphEncapsulateOp:: Input from non Variable Node";
 #endif
+
+      bool has_data_input =
+          NGraphCatalog::ExistsInInputDataTensorMap(def().input(i));
+      if (has_data_input) {
+        NGRAPH_VLOG(4) << " Input data already in device. For Input Index " << i
```
sayantan-nervana:
```suggestion
        NGRAPH_VLOG(4) << "Input data already in device. For Input Index " << i
```

src/ngraph_encapsulate_op.cc
```diff
@@ -543,9 +567,23 @@ class NGraphEncapsulateOp : public OpKernel {
...
+                       << endl;
+
+        auto data_input_tensor =
+            NGraphCatalog::GetTensorFromInputDataTensorMap(def().input(i));
+        OP_REQUIRES(ctx, data_input_tensor != nullptr,
+                    errors::Internal("Not found input data on Device.\n"));
+        ng_inputs.push_back(data_input_tensor);
+        continue;
```
sayantan-nervana: i'd prefer relying in if-else than in if-continue. if-else makes it easier to read i think. it makes it clear there is only 2 ways to get data... from`GetTensorFromInputDataTensorMap` or the usual way `GetCurrentNgTensor`

src/ngraph_input_data_pipeline.cc
```diff
@@ -0,0 +1,84 @@
...
+    for (int j = 0; j < tf_shape.dims(); ++j) {
+      ng_shape[j] = tf_shape.dim_size(j);
+    }
+
+    // Create nGTensor
+    auto ng_tensor = op_backend->create_tensor(ng_element_type, ng_shape);
+
+    // Load to Device
+    void* current_src_ptr = (void*)DMAHelper::base(input_tf_tensors[i]);
+    ng_tensor->write(current_src_ptr, 0,
```
sayantan-nervana: we might want to do a `if cpu` here. write is not needed for CPU

test/test_data_loading.cpp
```diff
@@ -0,0 +1,93 @@
...
+  // TF replaces them with _Arg nodes and renames them
+  vector<string> input_node_names = {"_arg_x_0_0", "_arg_y_0_1"};
+  vector<Tensor*> input_tensors = {&x, &y};
+  ASSERT_OK(NGraphInputDataPiepline::LoadInputDataOnDevice(input_node_names,
+                                                           input_tensors));
+
+  std::unique_ptr<Session> ng_session(NewSession(options));
+  ng_session->Create(gdef);
+  std::vector<Tensor> ng_outputs;
+  ng_session->Run({{"x", x}, {"y", y}}, {"mul", "add"}, {}, &ng_outputs);
```
sayantan-nervana: we can pass a new uninitialized tensor here (because it does not matter)

Also lets call `LoadInputDataOnDevice` twice. to test that the input is changing

test/test_data_loading.cpp
```diff
@@ -0,0 +1,93 @@
...
+
```
sayantan-nervana: In case that an input feeds to TF node? Can we assume the graph has already been ngraph-optimizer grappler processed.

src/ngraph_input_data_pipeline.cc
shresthamalik: Good point. As the API is called LoadInputDataOnDevice, we might not want to use it for CPU. We can discuss

src/ngraph_encapsulate_op.cc
shresthamalik: There are 3 ways right now as we also have the if for Variables also, but it is in if-def. I can put this in if-else and let that one (Variable) be as it is, with continue.

shresthamalik: Added comments for now

PR #55, opened 2019-05-05
Added option to build the INTELGPU backend

avijit-nervana:


build_ngtf.py
```diff
@@ -57,6 +57,11 @@ def main():
         help="nGraph backends will include PlaidML bckend\n",
         action="store_true")
 
+    parser.add_argument(
+        '--build_intelgpu_backend',
+        help="nGraph backends will include Intel GPU bckend\n",
```
sindhu-nervana: Since we have two arguments related to GPU now, should we rename the other argument we have to build Nvidia GPU to ```--build_nvidiagpu_backend```

build_ngtf.py
```diff
@@ -233,6 +238,11 @@ def main():
     else:
         ngraph_cmake_flags.extend(["-DNGRAPH_PLAIDML_ENABLE=NO"])
 
+    if arguments.build_intelgpu_backend:
+        ngraph_cmake_flags.extend(["-DNGRAPH_INTELGPU_ENABLE=YES"])
+    else:
+        ngraph_cmake_flags.extend(["-DNGRAPH_INTELGPU_ENABLE=NO"])
+
```
sayantan-nervana: Might i suggest:
```
flag_string_map = {True:'YES', False:'NO'}

ngraph_cmake_flags.extend(["-DNGRAPH_INTELGPU_ENABLE=" + flag_string_map[arguments.build_intelgpu_backend]])
ngraph_cmake_flags.extend(["-DNGRAPH_UNIT_TEST_ENABLE =" + flag_string_map[not arguments.use_prebuilt_tensorflow]])
#etc for all the ngraph_cmake_flags.extends we have that needs YES/NO
```

avijit-nervana: Agree. This makes sense.

build_ngtf.py
avijit-nervana: Good point. Will do.

avijit-nervana: Actually - changed my mind. I will keep this as is - because the name of the backend is encoded in the option itself. The bacends are GPU (NVidia GPU), PLAIDML etc. and their options are build_<backend_name>. I have added some help string to indicate how to use the specific backend.

PR #65, opened 2019-05-10
Kanvi/Redo BatchMatMul

kanvi-nervana:
Redo BatchMatMul since nGraph has the op support for it now
Add tests


test/test_math_ops.cpp
```diff
@@ -513,18 +514,100 @@ TEST(MathOps, BatchMatMul2D) {
...
+}
+
+// BatchMatMul2D
+// AdjX = true
+// AdjY = true
+TEST(MathOps, BatchMatMul2DAdjXY) {
+  Scope root = Scope::NewRootScope();
+
+  Tensor A(DT_FLOAT, TensorShape({2, 3}));
+  Tensor B(DT_FLOAT, TensorShape({3, 2}));
```
sayantan-nervana: since adj is true for both, what we are actually doing here is 3x2 times 2x3
the number `2` matches for both, `3` does not need to be same in the 2 matrices. so lets make the sizes: `2x4` and `3x2` say

test/test_math_ops.cpp
```diff
@@ -513,18 +514,100 @@ TEST(MathOps, BatchMatMul2D) {
...
+  opexecuter.RunTest();
+}
+
+// BatchMatMul2D
+// AdjX = true
+// AdjY = true
+TEST(MathOps, BatchMatMul2DAdjXY) {
+  Scope root = Scope::NewRootScope();
+
+  Tensor A(DT_FLOAT, TensorShape({2, 3}));
```
sayantan-nervana:
```suggestion
  Tensor A(DT_FLOAT, TensorShape({2, 4}));
```

test/test_math_ops.cpp
kanvi-nervana: Thanks :)

PR #67, opened 2019-05-10
Sarkars/backend setting

sayantan-nervana:
Absorb looking at env var inside the backend manager api


src/ngraph_backend_manager.cc
```diff
@@ -40,6 +40,8 @@ std::map<std::string, int> BackendManager::ref_count_each_backend_;
 
 Status BackendManager::SetBackendName(const string& backend_name) {
   std::lock_guard<std::mutex> lock(BackendManager::ng_backend_name_mutex_);
+  // TODO: Should NGRAPH_TF_BACKEND be read here as well and should it override
+  // backend_name?
```
shresthamalik: I would think No, if we look at environment variables as means to override some settings. GetCurrentlySetBackendName() returns the backend that is requested by the environment variable.

src/ngraph_backend_manager.cc
```diff
@@ -107,7 +109,28 @@ bool BackendManager::IsSupportedBackend(const string& backend_name) {
...
+}
+
+Status BackendManager::GetCurrentlySetBackendName(string* current_backend) {
+  string possible_backend;
+  const char* ng_backend_env_value = std::getenv("NGRAPH_TF_BACKEND");
+  // Environment variable has higher priority than API set value
+  if (ng_backend_env_value != nullptr) {
+    possible_backend = std::string(ng_backend_env_value);
+  } else {
+    possible_backend = BackendManager::ng_backend_name_;
```
shresthamalik: We dont need to check if the possible_backend is empty or is_supported in the case when NGRAPH_TF_BACKEND is nullptr. As we are doing that check in SetBackendName().

I was actually working on something similar
https://github.com/tensorflow/ngraph-bridge/pull/53/files#diff-73f0d2f88e27f31832d455ac7a525189R112

src/ngraph_backend_manager.cc
```diff
@@ -107,7 +109,28 @@ bool BackendManager::IsSupportedBackend(const string& backend_name) {
...
+  }
+  // Check if possible_backend is valid
+  if (possible_backend.empty() ||
+      !BackendManager::IsSupportedBackend(possible_backend)) {
+    current_backend = nullptr;
+    return errors::Internal("NGRAPH_TF_BACKEND: ", possible_backend,
+                            " is not supported");
+  } else {
+    *current_backend = possible_backend;
+    return SetBackendName(*current_backend);
```
shresthamalik: This is tricky. The API says GetCurrentlySetBackendName(), and we also SetBackendName() in it. It might be ok to not SetBackend Here.

test/opexecuter.cpp
```diff
@@ -332,6 +332,8 @@ void OpExecuter::ExecuteOnNGraph(vector<Tensor>& ngraph_outputs,
   // If NGRAPH_TF_BACKEND is set create that backend
   // Else create backend of type ng_backend_name
   string ng_backend_type = ng_backend_name;
+  // Note in unit tests opexecuter, backend manager is not used. So
+  // NGRAPH_TF_BACKEND is the only way to set backend
```
shresthamalik: or we can pass as an argument to the function ExecuteOnNGraph().

test/graph_rewrites/backend_manager_test.cc
```diff
@@ -40,21 +40,23 @@ namespace testing {
...
+For e.g. "BackendCLustering" test would fail if the NGRAPH_TF_BACKEND is set to
 INTERPRETER the nodes going to CPU also go to interpreter
 putting all the nodes in the same cluster.
 */
 
 // Test SetBackendAPI
 TEST(BackendManager, SetBackend) {
   ASSERT_OK(BackendManager::SetBackendName("CPU"));
-  string cpu_backend = BackendManager::GetCurrentlySetBackendName();
+  string cpu_backend;
```
shresthamalik: We would want to disable these tests when BackendManager::GetCurrentlySetBackendName() is not CPU

src/ngraph_backend_manager.cc
sayantan-nervana: yes probably should not set in get

We can discuss this scenario on monday:


initially backend='cpu'
env var is set to GPU
someone calls getbackend. it tries to return cpu, but sees envvar = gpu, so returns GPU

now env var is unset

next time someone queries getbackend, it would return "cpu". if i had called set in the get earlier it would return "gpu"
We want to decide if it should return cpu or gpu at this point.

if think we should return "cpu" (hence there should be no set in get)

examples/tf_cpp_examples/hello_tf.cpp
```diff
@@ -78,10 +78,13 @@ void RunSimpleNetworkExample() {
       ->mutable_rewrite_options()
       ->set_constant_folding(tensorflow::RewriterConfig::OFF);
 
-  std::cout
-      << "Currently selected backend: "
-      << tensorflow::ngraph_bridge::BackendManager::GetCurrentlySetBackendName()
-      << std::endl;
+  string backend;
+  if (tensorflow::ngraph_bridge::BackendManager::GetCurrentlySetBackendName(
```
sindhu-nervana: can we use this here? `ngraph_bridge_lib.ngraph_get_currently_set_backend_name`

PR #68, opened 2019-05-10
sarkars/Ensure grappler is enabled, and attach device

sayantan-nervana:


tools/tf2ngraph.py
```diff
@@ -177,6 +177,11 @@ def save_model(gdef, format, location):
     }[format](gdef, location)
 
 
+def attach_device(gdef):
+    for n in gdef.node:
+        n.device = "/job:localhost/replica:0/task:0/device:cpu:0"
```
avijit-nervana: Loogs good. But we need to ensure that if a node already has the device attached - then we skip that - no?

PR #74, opened 2019-05-13
Sarkars/enc to enc shared tensors

sayantan-nervana:
Looking for these 2 patterns:
E1<--E3-->E2
E1<--TF-->E2

Pre-req for enc-to-enc tensor sharing


src/enable_variable_ops/ngraph_rewrite_pass.cc
```diff
@@ -282,7 +282,7 @@ class NGraphEncapsulationPass : public NGraphRewritePass {
     // 4. Encapsulate clusters then, if requested, dump the graphs.
     FunctionDefLibrary* fdeflib_new = new FunctionDefLibrary();
     TF_RETURN_IF_ERROR(
-        EncapsulateClusters(options.graph->get(), idx, fdeflib_new));
+        EncapsulateClusters(options.graph->get(), idx, fdeflib_new, nullptr));
```
sayantan-nervana: Passing in `nullptr`: Right now we are not using the information we collect (Only unit test is using this info)

src/ngraph_encapsulate_clusters.cc
```diff
@@ -506,7 +507,105 @@ Status EncapsulateClusters(Graph* graph, int graph_id,
...
+                if (in_neighbour_out_edge->src_output() ==
+                    in_neighbour_edge->src_output()) {
+                  in_neighbour_out_src_slot =
+                      in_neighbour_out_edge->src_output();
+                  // Add the source of the shareable group
+                  add_to_group(in_neighbour, true, in_neighbour_out_src_slot,
+                               new_group);
+                }
+                // Add the sinks of the shareable group
+                add_to_group(in_neighbour_out, false, in_neighbour_out_dst_slot,
```
shresthamalik: The sink should only be added to the group if the 
```
in_neighbour_out_edge->src_output() ==
                    in_neighbour_edge->src_output()) 
```
ie the sink should be also part of the if? no?

shresthamalik: For a graph like below, it seems E2 and E3 are put in the same shared group, when they should not be
```
E1 (output index 1) -----> E2
(output index 0)
|
|
E3
```

sayantan-nervana: Will check this, and add a UT. Thanks

shresthamalik: Would this pseudo-algo be simpler?
```
 For each node:
          create new_grp
          Look at all output edges with src_output_index, i
                     if dst node is enc
                              Add the dstNode_input_id to the new_grp
         if node is enc
                    Add the node_output_id to the new_grp
         if new_grp is trivial 
                   Discard it
```

test/graph_rewrites/encapsulate_clusters_test.cc
```diff
@@ -113,6 +113,210 @@ TEST(EncapsulateClusters, PopulateLibrary) {
...
+  ASSERT_OK(EncapsulateClusters(&g, 0, fdeflib_new, &shared_tensors));
+
+  // There is only 1 group
+  ASSERT_EQ(shared_tensors.size(), 1);
+  // That group contains 2 shareable tensors
+  ASSERT_EQ(shared_tensors[0].size(), 3);
+
+  // Input slot 0 of encapsulate 0, 1 and 2 are shareable
+  ASSERT_EQ(shared_tensors[0], (std::set<string>{"0_0_0", "1_0_0", "2_0_0"}));
+}
```
shresthamalik: We can add tests with encapsulates with multiple inputs and outputs

PR #75, opened 2019-05-14
Sindhu/RsqrtGrad

sindhu-nervana:


src/ngraph_mark_for_clustering.cc
```diff
@@ -530,6 +531,7 @@ Status MarkForClustering(Graph* graph,
       type_constraint_map["Reshape"]["T"] = NGraphDTypes();
       type_constraint_map["Reshape"]["Tshape"] = NGraphIndexDTypes();
       type_constraint_map["Rsqrt"]["T"] = NGraphDTypes();
+      type_constraint_map["RsqrtGrad"]["T"] = NGraphDTypes();
```
sayantan-nervana: Probably should be realtypes. But did not check. Can you please check and confirm

test/python/test_rsqrtgrad.py
```diff
@@ -0,0 +1,61 @@
...
+from tensorflow.python.framework import constant_op
+from tensorflow.python.ops.gen_math_ops import rsqrt_grad
+
+from common import NgraphTest
+import numpy as np
+
+
+class TestRsqrtGrad(NgraphTest):
+
+    def test_rsqrtgrad_2d(self):
```
sayantan-nervana: Please use `parameterize` to avoid 2 separate tests. See https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/test_prod.py for an example

test/python/test_rsqrtgrad.py
```diff
@@ -0,0 +1,61 @@
...
+        y = constant_op.constant(
+            self.generate_random_numbers(6, 1.0, 10.0), shape=[2, 3])
+        dy = constant_op.constant(
+            self.generate_random_numbers(6, 0.0, 100.0), shape=[2, 3])
+
+        out = rsqrt_grad(y, dy)
+
+        def run_test(sess):
+            return sess.run(out)
```
sayantan-nervana: Using constants might constant fold. Please run `NGRAPH_TF_LOG_PLACEMENT=1` to see if it constant-folds. We can use placeholders for `y` and `dy`

sindhu-nervana: I checked the placement already :) 
```OP_placement:	Host	_SOURCE (NoOp)
OP_placement:	Host	_SINK (NoOp)
OP_placement:	Host	_retval_RsqrtGrad_0_0 (_Retval)
OP_placement:	nGraph[0]	Const (Const)
OP_placement:	nGraph[0]	Const_1 (Const)
OP_placement:	nGraph[0]	RsqrtGrad (RsqrtGrad)```

PR #78, opened 2019-05-15
Sarkars/unify rewrite pass

sayantan-nervana:
Merge `ngraph_rewrite_pass` from vars-and-optimizers path to normal path


src/ngraph_rewrite_pass.cc
```diff
@@ -209,6 +214,15 @@ class NGraphEncapsulationPass : public NGraphRewritePass {
       return Status::OK();
     }
 
+#if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
```
sayantan-nervana: Technically we dont need the ifdef perhaps? `ReplaceModifiers` should not have any effect in grappler or normal paths

src/ngraph_rewrite_pass.cc
```diff
@@ -248,6 +262,15 @@ class NGraphEncapsulationPass : public NGraphRewritePass {
                  "Graph with Variables Rewritten for Tracking");
     }
 
+#if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
```
sayantan-nervana: Similarly, this could run in the normal/grappler path without any ill-effects right? so the ifdef could be removed

src/ngraph_rewrite_pass.cc
kanvi-nervana: Yes, perhaps but I would suggest to keep the ifdef for clarity.

src/ngraph_rewrite_pass.cc
kanvi-nervana: Sam as above.

src/ngraph_encapsulate_op.cc
```diff
@@ -44,8 +44,8 @@
 #endif
 
 #if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
-#include "enable_variable_ops/ngraph_catalog.h"
-#include "enable_variable_ops/ngraph_var.h"
+#include "ngraph_catalog.h"
+#include "ngraph_var.h"
```
shresthamalik: Why were these moved out?

src/ngraph_rewrite_pass.cc
```diff
@@ -29,6 +29,11 @@
 #include "ngraph_utils.h"
 #include "tf_graph_writer.h"
 
+#if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
+#include "ngraph_enter_in_catalog.h"
+#include "ngraph_replace_variable_modifiers.h"
```
shresthamalik: They can be included from outside the enable_variable_ops/ ?

sayantan-nervana: Generally i was trying to move out everything I could from the `src/enable_variable_ops` folder to `src` in the general effort to merge normal + enable_variable_ops

Are you saying we merge files in normal + enable_variable_ops that already existed in normal, but for new files (like `ngraph_enter_in_catalog.h`) we keep them in `src/enable_variable_ops`. That is a good idea too, in case we want to sort of logically partition `src` so that `src` doesn't become a huge file dump?

PR #80, opened 2019-05-16
Sarkars/tf2ngraph out savedmodel master

sayantan-nervana:
Without loading the .so library through load_library, the ngencapsulate op is not registered in op_def_registry. That causes savedmodel generation to fail. Fixing that

### DIFF
--- a/python/ngraph_bridge/__init__.in.py
+++ b/python/ngraph_bridge/__init__.in.py
@@ -36,7 +36,6 @@ from tensorflow.python.framework import ops
 from tensorflow.core.protobuf import rewriter_config_pb2
 from tensorflow.python.framework import load_library
 
-
 import ctypes
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


python/ngraph_bridge/__init__.in.py
```diff
@@ -92,8 +93,9 @@
    (TF_INSTALLED_VER[1] == TF_NEEDED_VER[1]) and \
    ((TF_INSTALLED_VER[2].split('-'))[0] == (TF_NEEDED_VER[2].split('-'))[0]):
     libpath = os.path.dirname(__file__)
-    ngraph_bridge_lib = ctypes.cdll.LoadLibrary(
-        os.path.join(libpath, 'libngraph_bridge.' + ext))
+    full_lib_path = os.path.join(libpath, 'libngraph_bridge.' + ext)
+    _ = load_library.load_op_library(full_lib_path)
```
avijit-nervana: Good find! This makes sense.

PR #83, opened 2019-05-22
Shrestha/variables_share_buffer_on_cpu

shresthamalik:


src/enable_variable_ops/ngraph_var.h
```diff
@@ -76,17 +85,73 @@ class NGraphVar : public ResourceBase {
...
+    event_sync_ng_tf_tensors.Stop();
+    ngraph::Event::write_trace(event_sync_ng_tf_tensors);
+    return 1;
+  }
+
+  // If the NGTensor is behind TF Tensor (ie if NGTensor is out-of-date)
+  // It updates ng_tensor by copy_tf_to_ng
+  // Returns the number of copies made (0 or 1)
+  int sync_ng_tensor() {
+    if (ng_tf_share_buffer_) {
```
kanvi-nervana: May be you can remove this check here since you already have it in **copy_tf_to_ng**

src/enable_variable_ops/ngraph_var.h
```diff
@@ -76,17 +85,73 @@ class NGraphVar : public ResourceBase {
...
+      return 0;
+    }
+    if (sync_ng_tensor_) {
+      return copy_tf_to_ng();
+    }
+    return 0;
+  }
+
+  // updates the NGTensor with the new value
+  int update_ng_tensor(shared_ptr<ngraph::runtime::Tensor> new_value) {
```
kanvi-nervana: Assuming this is CPU specific and the other one is for other backends, correct?
Can you please add a comment?

test/models/mnist/mnist_deep_simplified.py
```diff
@@ -0,0 +1,258 @@
...
+import time
+
+from tensorflow.examples.tutorials.mnist import input_data
+
+import tensorflow as tf
+import ngraph_bridge
+
+seed = 1
+tf.random.set_random_seed(seed)
+
```
shresthamalik: It is the same mnist deep_simplified.py that we have in examples/mnist. Added some changes to make the script deterministic so that every run is same and the accuracy and loss printed in the logs can be compared. 

Changes:
1. Setting random seed

test/models/mnist/mnist_deep_simplified.py
```diff
@@ -0,0 +1,258 @@
...
+    train_writer.add_graph(tf.get_default_graph())
+
+    saver = tf.train.Saver()
+
+    with tf.Session(config=config) as sess:
+        sess.run(tf.global_variables_initializer())
+        train_loops = FLAGS.train_loop_count
+        loss_values = []
+        for i in range(train_loops):
+            batch = mnist.train.next_batch(FLAGS.batch_size, shuffle=False)
```
shresthamalik: shuffle = False is set for the same reason

test/models/mnist/mnist_deep_simplified_gradient_descent.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # Build the graph for the deep net
+    y_conv, keep_prob = deepnn(x)
+
+    with tf.name_scope('loss'):
+        cross_entropy = tf.nn.softmax_cross_entropy_with_logits(
+            labels=y_, logits=y_conv)
+    cross_entropy = tf.reduce_mean(cross_entropy)
+
+    with tf.name_scope('gradient_descent_optimizer'):
```
shresthamalik: Changes to make deterministic script + Gradient Descent Optimizer

src/enable_variable_ops/ngraph_var.h
shresthamalik: One is when the new_value is coming from nGraphEncapsulate Op and has been saved in Catalog. In that case we use copy_from on all backends

kanvi-nervana: Got it, please add this info as a comment for future refs.

test/models/mnist/mnist_deep_simplified.py
sayantan-nervana: If the only changes are setting the seed and shuffle = False, we can add these changes in the main file and not create a new file

test/models/mnist/validate_mnist.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+
+
+echo "**************************************************************************"
+echo "Run MNIST ADAM OPTIMIZER on NGRAPH"
+echo "**************************************************************************"
+python mnist_deep_simplified.py --train_loop_count=20
```
sayantan-nervana: Are we comparing against known values of accuracy etc? Or Is this script meant to be a one-stop runme for running the models. If we want a test, we can wrap this in a pytest

shresthamalik: one-stop runme for running the models with and without nGraph. I will open another PR for testing.

test/models/mnist/validate_mnist.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Run MNIST GRADIENT DESCENT on NGRAPH"
+echo "**************************************************************************"
+
+python $mnist_training_example --train_loop_count=20 --make_deterministic --optimizer sgd
+
+echo "**************************************************************************"
+echo "Run MNIST GRADIENT DESCENT on TF"
+echo "**************************************************************************"
+
+NGRAPH_TF_DISABLE=1 python $mnist_training_example --train_loop_count=20 --make_deterministic --optimizer sgd
```
shresthamalik: Currently this is something you can quickly run and verify the accuracy that gets printed. I will change this to an automated test in another PR. It's a handy script for now.

PR #84, opened 2019-05-24
Sarkars/Add file for batchmatmul, to fix interpreter build. Also new release

sayantan-nervana:
1. Updates ngraph version
2. Fix bazel build of `libinterpreter_backend.so`
3. Remove provenance tagging for now

### DIFF
--- a/python/setup.in.py
+++ b/python/setup.in.py
@@ -59,7 +59,7 @@ package_data_dict['ngraph_bridge'] = include_list
 
 setup( 
     name='ngraph_tensorflow_bridge',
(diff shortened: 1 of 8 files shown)


python/setup.in.py
```diff
@@ -59,7 +59,7 @@ def get_tag(self):
 
 setup( 
     name='ngraph_tensorflow_bridge',
-    version='0.12.0',
+    version='0.14.0-rc1',
```
avijit-nervana:
```suggestion
    version='0.14.0rc1',
```

PR #88, opened 2019-05-28
Kanvi/freshness tracker

kanvi-nervana:


src/enable_variable_ops/ngraph_assign_op.cc
```diff
@@ -48,9 +48,10 @@ namespace ngraph_bridge {
 // Computes *input[0] = input[1]
 class NGraphAssignOp : public OpKernel {
  private:
-  bool just_looking_;
+  bool is_tf_modifying_;
   bool copy_to_tf_;
   int ng_graph_id_;
+  bool just_looking_;
```
kanvi-nervana: Should I change the variable name from just_looking to something else? May be is_not_changing or is_tracked or will_not_be_modified or ?

avijit-nervana: Yes - please do so.

src/enable_variable_ops/ngraph_variable_modifiers.cc
```diff
@@ -76,7 +76,8 @@ REGISTER_OP("NGraphApplyGradientDescent")
     .Output("out: Ref(T)")
     .Attr("T: numbertype")
     .Attr("use_locking: bool = false")
-    .Attr("just_looking: bool = false")
+    .Attr("just_looking: bool = true")
```
shresthamalik: Let the default be false, for consistency with other ops

src/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -67,29 +69,36 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
-          // turn off just_looking
+          // turn off is_tf_modifying
           if (!IsNGVariableType(edge->dst()->type_string())) {
-            NGRAPH_VLOG(1) << DebugNode(edge->dst())
-                           << "needs reference, setting just_looking to false";
-            just_looking = false;
+            NGRAPH_VLOG(1)
+                << DebugNode(edge->dst())
+                << "needs reference, setting is_tf_modifying to false";
+            is_tf_modifying = false;
```
shresthamalik: Either change the name 'is_tf_not_modifying' or  set  'is_tf_modifying = true'
In the 2nd case you will need to change the checks for (!is_tf_modifying) to (is_tf_modifying)

src/enable_variable_ops/ngraph_assign_op.cc
shresthamalik: is_not_modified, is_not_modified_by_tf

shresthamalik: or just_looking, tf_just_looking

sayantan-nervana: Can you please add a comment in the file here what `just_looking_` and `is_tf_modifying_` would mean, and why we need 2 flags instead of just one (normal path vs var-opt path etc)

sayantan-nervana: `flib_graph_2.pbtxt` might need changes (adding `is_tf_modifying_` attribute): https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/flib_graph_2.pbtxt

src/enable_variable_ops/ngraph_rewrite_for_tracking.cc
kanvi-nervana: Changed it to is_tf_just_looking

PR #90, opened 2019-05-29
Sarkars/mark node with specific deviceid

sayantan-nervana:
Core change:
1. Call `ngraph_bridge.set_backend(X)`, where `X` is a ng backend string such as `CPU` or `GPU:0` etc 

Other changes:
1. Changed optional `_ngraph_backend` attribute of EncapsulateOp to compulsory `ngraph_backend`, to avoid confusion between the tag in normal TF nodes and EncapsulateOp nodes
2. Added a unit test for `mark_for_clustering`
3. Ectended tf2ngraph's pytest to include 2 backends


src/ngraph_encapsulate_clusters.cc
```diff
@@ -330,7 +330,7 @@ Status EncapsulateClusters(Graph* graph, int graph_id,
     Node* n;
     Status status = NodeBuilder(ss.str(), "NGraphEncapsulate")
                         .Attr("ngraph_cluster", cluster_idx)
-                        .Attr("_ngraph_backend", cluster_backend)
+                        .Attr("ngraph_backend", cluster_backend)
```
sayantan-nervana: Normal TF nodes are marked with `_ngraph_backend`, since underscore is needed to name an undeclared field.
Before this `NGraphEncapsulate`, our custom tf node was also using `_ngraph_backend`. In this PR, we make `ngraph_backend` a declared field, to avoid confusion.

test/graph_rewrites/encapsulate_clusters_test.cc
```diff
@@ -55,7 +55,7 @@ TEST(EncapsulateClusters, PopulateLibrary) {
   Node* node2;
   ASSERT_OK(NodeBuilder("node2", "Const")
                 .Attr("dtype", DT_FLOAT)
-                .Attr("value", t_shape)
+                .Attr("value", t_input_1)
```
sayantan-nervana: Cosmetic changes in this file

test/graph_rewrites/backend_manager_test.cc
```diff
@@ -111,6 +111,10 @@ TEST(BackendManager, BackendAssignment) {
   ASSERT_EQ(bA, bR);
   ASSERT_EQ(bA, backend1);
 
+  for (auto node : graph.op_nodes()) {
+    node->ClearAttr("_ngraph_backend");
+  }
```
sayantan-nervana: TODO: revisit

src/ngraph_mark_for_clustering.cc
```diff
@@ -630,7 +630,40 @@ Status MarkForClustering(Graph* graph,
...
+      }
+      if (backend_name_from_nodes.empty()) {
+        backend_name_from_nodes = backend_name_from_this_node;
+      }
+    }
+  }
+
+  if (!backend_name_from_nodes.empty()) {
+    current_backend = backend_name_from_nodes;
+  }
```
sayantan-nervana: So this thing is not needed now that we use the SetBackend API in `tf2ngraph`

test/graph_rewrites/mark_for_clustering_test.cc
```diff
@@ -0,0 +1,87 @@
...
+  if (ng_backend_env_value != nullptr) {
+    expected_backend = std::string(ng_backend_env_value);
+  }
+
+  string backend;
+  for (auto node : g.op_nodes()) {
+    ASSERT_OK(GetNodeBackend(node, &backend));
+    ASSERT_EQ(backend, expected_backend);
+  }
+}
```
sayantan-nervana: General unit test for mark_for_clustering. Does not test this current feature specifically

PR #96, opened 2019-06-04
Shrestha/mnist test

shresthamalik:
Added MNIST Training Test
- Runs end-to-end training of mnist dataset as in examples/mnist/mnist_deep_simplified.py
for 20 iterations. Gets loss value for each iteration + final test accuracy (accuracy on validation test).
- Compares the values between  TF-CPU backend and nGraph backend


test/python/test_mnist_training.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # disable ngraph-tf
+        ngraph_bridge.disable()
+        tf_loss_values, tf_test_accuracy = train_mnist_cnn(FLAGS)
+        tf_values = tf_loss_values + [tf_test_accuracy]
+
+        # compare values
+        assert np.allclose(
+            ng_values, tf_values,
+            atol=1e-3), "Loss or Accuracy values don't match"
```
shresthamalik: Please note, I have added  tolerance of 1e-3. For eg. on CPU we see values like 0.3973 and 0.3974, which dont pass with the default tolerance values

PR #99, opened 2019-06-05
Sindhu/gathernd

sindhu-nervana:


test/test_array_ops.cpp
```diff
@@ -316,6 +316,55 @@ TEST(ArrayOps, ExpandDims) {
...
+
+  OpExecuter opexecuter(root, "GatherNd", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+
+}  // end of test op GatherND
+
+// Test op:GatherND
+TEST(ArrayOps, GatherNd_3D) {
```
sayantan-nervana: Please remove underscore from gtest names. apparently gtest reserves the right to use underscores willy-nilly

sindhu-nervana:
Two gather_nd tf python tests are failing, so disabled those. Debugging in parallel.

test/test_array_ops.cpp
```diff
@@ -316,6 +316,55 @@ TEST(ArrayOps, ExpandDims) {
...
+  std::vector<Output> sess_run_fetchoutputs = {R};
+
+  OpExecuter opexecuter(root, "GatherNd", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+
+}  // end of test op GatherND
+
 // Test op: Gather. vector indices
```
sindhu-nervana: These C++ tests will fail on GPU as GatherNd is not supported on GPU. `buildkite/ngtf-r0-dot-13-cpu-ubuntu-18-04`

src/ngraph_mark_for_clustering.cc
```diff
@@ -439,6 +440,9 @@ Status MarkForClustering(Graph* graph,
       // DT_FLOAT
       type_constraint_map["FusedBatchNormV2"]["T"] = {DT_FLOAT};
       type_constraint_map["FusedBatchNormGrad"]["T"] = NGraphNumericDTypes();
+      type_constraint_map["GatherNd"]["Tparams"] = {
+          DT_FLOAT};  // NGraphDTypes();
```
sayantan-nervana: Can we expand this to more types?

test/test_array_ops.cpp
sayantan-nervana: To pass CI on GPU, please add this test to the filter here: https://github.com/tensorflow/ngraph-bridge/blob/master/test/ci/buildkite/test_runner.py#L97

src/ngraph_mark_for_clustering.cc
sindhu-nervana: TF python tests were throwing unsupported CPU builder errors for more types.

sindhu-nervana: TODO:   {DT_FLOAT} passes, but the full list `NGraphDTypes()` fails, so may need to check what other datatypes will pass

sindhu-nervana:
@avijit-nervana Looks like the failure was introduced after TF 1.14 change and further. I am bisecting now to figure out which commit might've introduced it to fix.

PR #100, opened 2019-06-06
Sarkars/merge r013

sayantan-nervana:


python/setup.in.py
```diff
@@ -72,6 +72,7 @@ def get_tag(self):
     package_data= package_data_dict,
     cmdclass={'bdist_wheel': BinaryBdistWheel},
     extras_require={
+        'plaidml': ["tensorflow==1.13.0"],
```
sayantan-nervana: TODO check: now that master is at 1.14, this will not work?

test/ci/buildkite/ngtf-cpu_centos-grappler.yaml
```diff
@@ -17,7 +17,7 @@
       python3 build_ngtf.py --use_grappler --artifacts /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID
       
     label: ":hammer_and_wrench: Build"
-    timeout_in_minutes: 40
+    timeout_in_minutes: 60
```
sayantan-nervana: Increasing buildtimes, as CI seems to timeout during building with the new upgrade to tf1.14

test/opexecuter.h
```diff
@@ -89,8 +89,6 @@ class OpExecuter {
 
   shared_ptr<ng::Function> get_ng_function() { return ng_function; }
 
-  // TODO(malikshr) : Overload RunTest() to take in tolerance
-
```
sayantan-nervana: This TODO is now a BEENDONE

python/setup.in.py
avijit-nervana: It should. Will find out :)

PR #103, opened 2019-06-07
Kanvi/Config_support

kanvi-nervana:
1. Added capability to pass backend configurations through Grappler's Rewriter Config
2. Added BackendConfig class to capture backend specific attributes
3. BackendManager interfaces for BackendConfig


src/enable_variable_ops/ngraph_rewrite_pass.cc
```diff
@@ -249,17 +249,39 @@ class NGraphEncapsulationPass : public NGraphRewritePass {
...
+          !BackendManager::IsSupportedBackend(backend_env)) {
+        return errors::Internal("NGRAPH_TF_BACKEND: ", backend_env,
+                                " is not supported");
+      }
+      backend_name = backend_env;
+    }
+    config_map = BackendManager::GetBackendAttributes(
+        backend_name);  // SplitBackendConfig
+    backend_name = config_map.at("ngraph_backend");
+    config_map.erase("ngraph_backend");
```
sayantan-nervana: Why do we `erase` the attribute?

test/graph_rewrites/backend_manager_test.cc
```diff
@@ -159,29 +162,87 @@ TEST(BackendManager, BackendClustering) {
   ASSERT_NE(A_cluster, B_cluster);
 }
 
-// Test Backend Run
-TEST(BackendManager, BackendRun) {
```
shresthamalik: Removed the BackendRun test, was running end to end and checking for failures in general. Wasnt testing anything specific

src/enable_variable_ops/ngraph_rewrite_pass.cc
```diff
@@ -249,17 +249,39 @@ class NGraphEncapsulationPass : public NGraphRewritePass {
...
       DumpGraphs(options, idx, "replaced_modifier",
                  "Graph with Modifiers replaced");
     }
 
     // 1. Mark for clustering then, if requested, dump the graphs.
     std::set<string> skip_these_nodes = {};
-    TF_RETURN_IF_ERROR(
-        MarkForClustering(options.graph->get(), skip_these_nodes));
+    TF_RETURN_IF_ERROR(MarkForClustering(options.graph->get(), skip_these_nodes,
+                                         backend_name));
```
sayantan-nervana: Note to self: `MarkForClustering` and `EncapsulateClusters` are now explicitly passed the `backend`
`backend_name` is expected to be only the name `CPU` and not `CPU:0` etc.

test/graph_rewrites/backend_manager_test.cc
```diff
@@ -138,7 +141,7 @@ TEST(BackendManager, BackendClustering) {
 
   std::set<string> skip_these_nodes = {};
 
-  ASSERT_OK(MarkForClustering(&graph, skip_these_nodes));
+  ASSERT_OK(MarkForClustering(&graph, skip_these_nodes, "CPU"));
```
shresthamalik: Most of the backend tests were calling MarkForClustering and checking if the right backend (the backend that was set) was getting attached as attribute. Now MarkForClustering takes backend name as input, so the test is kind of trivial.

The optimizer pass already has backend_name as data member, if these functions can be made member functions of the rewrite pass/optimizer classes we wont need to pass these as arguments. we are incrementally having more arguments being passed around for different functions :)

src/grappler/ngraph_optimizer.cc
```diff
@@ -166,8 +191,30 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
     DumpGraphs(graph, idx, "unmarked", "Unmarked Graph");
   }
 
+  // Get backend + its configurations, to be attached to the nodes
```
sayantan-nervana: Can you please add a comment about the order or precedence:

BM < Env var <  rewriter config

Is the precedence intentional, or do you want to refactor the if-else to make the precedence BM < rewriter config < Env var

src/enable_variable_ops/ngraph_rewrite_pass.cc
shresthamalik: backend_name is required by mark for clustering. We use the use backend_name to assign_clusters (ops marked with different backends are not clustered together).  

The other attributes are added directly in encapsulate clusters through the config_map.

src/enable_variable_ops/ngraph_rewrite_pass.cc
shresthamalik: EncapsulateClusters are passed the additional attributes (_ngraph_device_config etc)
MarkForClustering is passed the backend name

src/enable_variable_ops/ngraph_rewrite_pass.cc
sayantan-nervana: Oh, I see. `config_map` in `EncapsulateClusters` is not expected to contain `ngraph_backend`

src/ngraph_backend_config.h
```diff
@@ -0,0 +1,57 @@
...
+using namespace std;
+
+namespace tensorflow {
+
+namespace ngraph_bridge {
+
+class BackendConfig {
+ public:
+  BackendConfig() = delete;
+  BackendConfig(string backend_name);
```
sayantan-nervana: Minor
We can make this constructor private and `BackendManager` a friend class, so no one will be able to use `BackendConfig` other than `BackendManager`. `BackendConfig` could even be a nested class in `BackendManager`

src/ngraph_backend_manager.cc
```diff
@@ -109,5 +112,47 @@ bool BackendManager::IsSupportedBackend(const string& backend_name) {
   return true;
 };
 
+// Backend Config functions
+BackendConfig* BackendManager::GetBackendConfig(const string& backend_name) {
+  std::lock_guard<std::mutex> lock(BackendManager::ng_backend_map_mutex_);
```
sayantan-nervana: Maybe add that comment that BackendConfig is expected to be a readonly class, hence only locked at creation, but not during later access

src/ngraph_backend_manager.cc
```diff
@@ -109,5 +112,47 @@ bool BackendManager::IsSupportedBackend(const string& backend_name) {
...
+  if (itr == BackendManager::ng_backendconfig_map_.end()) {
+    BackendConfig* bconfig = nullptr;
+    if (backend_name == "NNPI") {
+      NGRAPH_VLOG(3) << "Creating NNPI Backend config";
+      bconfig = new BackendNNPIConfig();
+    } else {
+      NGRAPH_VLOG(3) << "Creating default Backend config";
+      bconfig = new BackendConfig(backend_name);
+    }
+    BackendManager::ng_backendconfig_map_[backend_name] = bconfig;
```
sayantan-nervana: We may want to free `bconfig` in `~BackendManager` (by freeing all the values of ng_backendconfig_map_) or save it as smart pointers for auto-freeing

src/ngraph_encapsulate_op.cc
```diff
@@ -168,8 +169,21 @@ class NGraphEncapsulateOp : public OpKernel {
...
+    std::string backend_name;
+    OP_REQUIRES_OK(ctx, ctx->GetAttr<string>("ngraph_backend", &backend_name));
+    // Get the optional attributes
+    std::vector<std::string> optional_attributes =
+        BackendManager::GetOptionalAttributes(backend_name);
+    std::unordered_map<std::string, std::string> optional_attribute_map;
+    for (int i = 0; i < optional_attributes.size(); i++) {
+      std::string val;
+      // If an attribute does not exist, TF will return a non-ok status
+      OP_REQUIRES_OK(ctx, ctx->GetAttr<string>(optional_attributes[i], &val));
```
sayantan-nervana: These are optional attributes. So we should not be `OP_REQUIRES_OK`-ing here right?

Also do we have a unit test where we do not pass in optional attributes?

src/ngraph_backend_manager.h
```diff
@@ -72,6 +74,46 @@ class BackendManager {
...
+  GetBackendAttributes(  // SplitBackendConfig
+      string backend_config);
+
+  // Given a backend name and list of attributes
+  // joins them into a string to create ngraph backend
+  // For e.g.
+  // 1. GetBackendCreationType("GPU", {"ngraph_device_config", "2"})
+  // returns "GPU:2"
+  // throws an error if the required attributes are not present in the map
+  static string GetBackendCreationType(
```
sayantan-nervana: `GetBackendCreationType` or `GetBackendCreationString`?

src/ngraph_encapsulate_op.cc
sayantan-nervana: Note: this is by design. Its designed to fail if an attribute is not present to prevent any accidental foot-shooting.

src/enable_variable_ops/ngraph_rewrite_pass.cc
```diff
@@ -249,17 +249,41 @@ class NGraphEncapsulationPass : public NGraphRewritePass {
...
+    string backend_name = BackendManager::GetCurrentlySetBackendName();
+    const char* ng_backend_env_value = std::getenv("NGRAPH_TF_BACKEND");
+    if (ng_backend_env_value != nullptr) {
+      string backend_env = std::string(ng_backend_env_value);
+      if (backend_env.empty() ||
+          !BackendManager::IsSupportedBackend(backend_env)) {
+        return errors::Internal("NGRAPH_TF_BACKEND: ", backend_env,
+                                " is not supported");
+      }
+      backend_name = backend_env;
```
avijit-nervana: Add a V(0) log item indicating that we what backend is specified and whether we are overriding using the environment variable.

src/grappler/ngraph_optimizer.cc
```diff
@@ -38,6 +39,31 @@ using namespace std;
 namespace tensorflow {
 namespace ngraph_bridge {
 
+Status NgraphOptimizer::Init(
+    const tensorflow::RewriterConfig_CustomGraphOptimizer* config) {
+  if (config == nullptr) {
```
avijit-nervana: No need for checking null ness

src/grappler/ngraph_optimizer.cc
```diff
@@ -62,15 +88,14 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
-  // passes become a no-op.
+  // we will not do anything; all subsequent passes become a no-op.
   bool ngraph_not_enabled =
       (!config::IsEnabled()) || (std::getenv("NGRAPH_TF_DISABLE") != nullptr);
   bool already_processed = IsProcessedByNgraphPass(&graph);
   if (ngraph_not_enabled || already_processed) {
-    NGRAPH_VLOG(0) << "Not running through nGraph. nGraph not enabled: "
-                   << ngraph_not_enabled
-                   << " Already processed: " << already_processed;
+    NGRAPH_VLOG(0)
```
avijit-nervana: Split this into two whether nGraph is enabled or not and then if enabled, whether we are skipping the step due to already processed. Then use VLOG level 1 to report these. Not 0. As we do anticipate workloads in which nGraph is disabled for certain parts of the computation. So it's not a Level 0 candidate.

src/grappler/ngraph_optimizer.cc
```diff
@@ -166,8 +191,32 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
   }
 
+  // Get backend + its configurations, to be attached to the nodes
+  // Precedence Order: RewriteConfig> Env Variable > BackendManager
+  string backend_name = BackendManager::GetCurrentlySetBackendName();
+  if (!config_backend_name.empty()) {
+    backend_name = config_backend_name;
+  } else {
+    const char* ng_backend_env_value = std::getenv("NGRAPH_TF_BACKEND");
+    if (ng_backend_env_value != nullptr) {
```
avijit-nervana: So - is the backend selection logic common? may be we can refactor this into a function and use for both tracks?

src/ngraph_backend_config.cc
```diff
@@ -0,0 +1,100 @@
...
+namespace ngraph_bridge {
+
+BackendConfig::BackendConfig(string backend_name) {
+  NGRAPH_VLOG(3) << "BackendConfig() ";
+  backend_name_ = backend_name;
+  optional_attributes_ = {"_ngraph_device_config"};
+}
+
+string BackendConfig::join(unordered_map<string, string> optional_parameters) {
+  NGRAPH_VLOG(3) << "JOIN";
```
avijit-nervana: If you add a log message - add more information such as the length of the map may be?

src/ngraph_backend_config.cc
```diff
@@ -0,0 +1,100 @@
...
+  try {
+    optional_parameters.at("_ngraph_device_config");
+  } catch (std::out_of_range e1) {
+    throw std::out_of_range("Attribute _ngraph_device_config not found");
+  }
+  return backend_name_ + ":" + optional_parameters.at("_ngraph_device_config");
+}
+
+unordered_map<string, string> BackendConfig::split(string backend_config) {
+  NGRAPH_VLOG(3) << "SPLIT";
```
avijit-nervana: Remove these logs - they don't produce any meaningful information.

src/ngraph_backend_config.cc
```diff
@@ -0,0 +1,100 @@
...
+  // throw an error
+  try {
+    optional_parameters.at("_ngraph_device_config");
+  } catch (std::out_of_range e1) {
+    throw std::out_of_range("Attribute _ngraph_device_config not found");
+  }
+  return backend_name_ + ":" + optional_parameters.at("_ngraph_device_config");
+}
+
+unordered_map<string, string> BackendConfig::split(string backend_config) {
```
avijit-nervana: Since this is a user facing function - need to add a unit test that checks for corner cases.

src/ngraph_backend_config.h
avijit-nervana: I like the nested class idea. `Friend` - no :)

src/grappler/ngraph_optimizer.cc
shresthamalik: Sometimes we have expectation that it will run through nGraph, but it does not. Having it at Vlog0 is helpful then.  
The log message does indicate whether running through nGraph or not, using the booleans ngraph_not_enabled
and already_processed (0/1). Is that ok or we want to break it further?
```<< "NGTF_OPTIMIZER: Not running through nGraph. nGraph not enabled: "
        << ngraph_not_enabled << " Already processed: " << already_processed;```

File: src/ngraph_backend_config.cc (hunk 17 above)
Comment (shresthamalik) [reply src/ngraph_backend_config.cc]: We have some tests here:
https://github.com/tensorflow/ngraph-bridge/pull/103/files#diff-1daf047044b07a8ed776c373202925d6R179

File: src/ngraph_backend_config.cc [hunk 18]
```diff
@@ -0,0 +1,100 @@
...
+BackendConfig::BackendConfig(string backend_name) {
+  NGRAPH_VLOG(3) << "BackendConfig() ";
+  backend_name_ = backend_name;
+  optional_attributes_ = {"_ngraph_device_config"};
+}
+
+string BackendConfig::join(unordered_map<string, string> optional_parameters) {
+  NGRAPH_VLOG(3) << "JOIN";
+  // If _ngraph_device_config is not found
+  // throw an error
```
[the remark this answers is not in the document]
shresthamalik: We do not have an assert on the optional_parameters size, as it can be greater than 1. For eg in NNPI we have 4 attributes being passed though only one is being used to create the string.

We are throwing an error if the attribute we are interested in is not passed.

src/enable_variable_ops/ngraph_rewrite_pass.cc
kanvi-nervana: We do have log indicating which backend was set here: https://github.com/tensorflow/ngraph-bridge/pull/103/files/5a85aded0c924619dab17f22ccf11a3a9706a84b#diff-f3f34e20694334a7d37b19708a2ef258R272
I will change the log level and add the log for whether it was set using env variable.

kanvi-nervana: @avijit-nervana V(0) might be too much, do you think we need that or is V(1) ok?

src/grappler/ngraph_optimizer.cc
```diff
@@ -192,10 +190,16 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
   }
 
   // Get backend + its configurations, to be attached to the nodes
-  // Precedence Order: RewriteConfig> Env Variable > BackendManager
+  // Precedence Order: RewriteConfig > Env Variable > BackendManager
   string backend_name = BackendManager::GetCurrentlySetBackendName();
   if (!config_backend_name.empty()) {
+    if(!BackendManager::IsSupportedBackend(backend_name)) {
```
kanvi-nervana: Check to see if the requested backend through the config is supported.

test/graph_rewrites/config_for_grappler_test.cc
```diff
@@ -70,13 +70,9 @@ TEST(GrapplerConfig, RConfig1) {
...
   ConfigProto config_proto;
   auto backend_name = AttrValue();
-  backend_name.set_s("NNPI");
-  auto device_id = AttrValue();
-  device_id.set_s("1");
-  auto num_ice_cores = AttrValue();
-  num_ice_cores.set_s("4");
-  auto max_batch_size = AttrValue();
-  max_batch_size.set_s("64");
+  backend_name.set_s("CPU");
```
kanvi-nervana: Had to redo the test to be for CPU backend since a check was added to see if the requested backend is supported and the earlier test would fail.

src/enable_variable_ops/ngraph_rewrite_pass.cc
shresthamalik: V(0) for backend_name, and v(1) for who is setting the backend

src/grappler/ngraph_optimizer.cc
shresthamalik: Only print whether we are running through nGraph finally V(0)
And V(1)

src/ngraph_backend_manager.h
```diff
@@ -83,8 +125,15 @@ class BackendManager {
   // set of backends supported by nGraph
   static unordered_set<string> ng_supported_backends_;
 
+  // map of cached backend config objects
+  static unordered_map<string, BackendConfig*> ng_backendconfig_map_;
```
shresthamalik: change to unique ptr

src/grappler/ngraph_optimizer.cc
shresthamalik: We will change the BackendManager::GetCurrentlySetBackendName() function to return the backend set by the environment variable in a future PR.

src/ngraph_backend_config.h
shresthamalik: Decided to keep as it is

src/ngraph_backend_config.cc
```diff
@@ -0,0 +1,98 @@
...
+  optional_attributes_ = {"_ngraph_device_id", "_ngraph_ice_cores",
+                          "_ngraph_max_batch_size"};
+}
+
+string BackendNNPIConfig::join(
+    unordered_map<string, string> optional_parameters) {
+  // If _ngraph_device_id is not found
+  // throw an error
+  try {
+    optional_parameters.at("_ngraph_device_id");
```
sindhu-nervana: This join is similar to BackendConfig::join above with only difference being config and id right. So what if we have one `join` function and use for both BackendConfig and BackendNNPIConfig?

src/grappler/ngraph_optimizer.cc
sayantan-nervana: in terms of the booleans: `already_processed` and `ngraph_not_enabled`

Will finally run thru ngraph = `already_processed || ! ngraph_not_enabled`

will finally NOT run thru ngraph: negate above: `!already_processed && ngraph_not_enabled`

will run not rewrite pass if `ngraph_not_enabled || already_processed`

will run rewrite pass (negate above line):  `!ngraph_not_enabled && !already_processed`

```
if (!already_processed && ngraph_not_enabled) {
  VLOG(0) << "Graph will not run through ngraph"
}

if (ngraph_not_enabled || already_processed) {
VLOG(1) << "Rewrite pass will not run because " + (already_processed ? "graph is already preprocessed" : "ngraph is disabled");
}

```

i'd say we can have `ngraph_enabled` instead of `ngraph_not_enabled` (and have appropriate boolean expressions after that.). 


Also, this can be in a separate PR, because this PR does not change this prints, it just shows up in the diff list because this change just adds `NGTF_OPTIMIZER` to the string

src/grappler/ngraph_optimizer.cc
```diff
@@ -166,8 +189,42 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+                        "NGRAPH_TF_BACKEND = "
+                     << backend_name;
+    } else {
+      NGRAPH_VLOG(1) << "Setting backend from the BackendManager ";
+    }
+    // splits into {"ngraph_backend", "_ngraph_device_config"}
+    config_map = BackendManager::GetBackendAttributes(
+        backend_name);  // SplitBackendConfig
+    backend_name = config_map.at("ngraph_backend");
+    config_map.erase("ngraph_backend");
```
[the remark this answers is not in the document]
kanvi-nervana: Yes, we don't need this map entry for later. The ngraph_backend is sent to mark_for_clustering as the backend_name parameter where it is added as a node attr. The rest of the config options are sent to encapsulate_clusters as a map and that does not need ngraph_backend since it will already be an attr of the node.

src/ngraph_backend_config.cc
shresthamalik: For backends that are not NNPIConfig we have a different attribute as it is expected to be more than just the device id. So keeping them separate. We also have separate checks.

src/ngraph_backend_manager.h
```diff
@@ -83,8 +125,17 @@ class BackendManager {
...
+  // map of cached backend config objects
+  static unordered_map<string, std::shared_ptr<BackendConfig>>
+      ng_backendconfig_map_;
+  static mutex ng_backendconfig_map_mutex_;
+
   // Map of backends and their reference counts
   static std::map<std::string, int> ref_count_each_backend_;
+
+  // utility functions
+  static std::shared_ptr<BackendConfig> GetBackendConfig(
```
shresthamalik: on using unique_ptr, i get this error
`make_unique is not a member of ‘std’`
It seems our compiler doesn't support this
https://stackoverflow.com/questions/24609271/errormake-unique-is-not-a-member-of-std

Using shared_ptr for now

python/ngraph_bridge/__init__.in.py
```diff
@@ -210,6 +210,18 @@ def update_config(config):
...
+        # rewriter_options = rewriter_config_pb2.RewriterConfig()
+        # rewriter_options.meta_optimizer_iterations=(rewriter_config_pb2.RewriterConfig.ONE)
+        # rewriter_options.min_graph_nodes=-1
+        # ngraph_optimizer = rewriter_options.custom_optimizers.add()
+        # ngraph_optimizer.name = "ngraph-optimizer"
+        # ngraph_optimizer.parameter_map["ngraph_backend"].s = b'NNPI'
+        # ngraph_optimizer.parameter_map["_ngraph_device_id"].s = b'1'
+        # ngraph_optimizer.parameter_map["_ngraph_max_batch_size"].s = b'64'
+        # ngraph_optimizer.parameter_map["_ngraph_ice_cores"].s = b'12'
+        # config.MergeFrom(tf.ConfigProto(graph_options=tf.GraphOptions(rewrite_options=rewriter_options)))
```
sayantan-nervana: This could be a Pytest

sayantan-nervana: Can be added in the PR that modifies tf2ngraph based on this

src/enable_variable_ops/ngraph_rewrite_pass.cc
```diff
@@ -241,25 +241,58 @@ class NGraphEncapsulationPass : public NGraphRewritePass {
...
+    } else {
+      NGRAPH_VLOG(1) << "Setting backend from the BackendManager ";
+    }
+
+    // splits into {"ngraph_backend", "_ngraph_device_config"}
+    config_map = BackendManager::GetBackendAttributes(
+        backend_name);  // SplitBackendConfig
+    backend_name = config_map.at("ngraph_backend");
+    config_map.erase("ngraph_backend");
+    NGRAPH_VLOG(0) << "NGraphEncapsulationPass: backend_name " << backend_name;
```
avijit-nervana: This is VLOG(0) that means this will be received by data scientists etc. They don't need to know or should care about `NGraphEncapsulationPass:`. Change this to more user friendly message. For example: `nGraph using backend: CPU:0`

src/grappler/ngraph_optimizer.cc
```diff
@@ -62,15 +86,17 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
 
   // If ngraph is disabled via ngraph_bridge api or NGRAPH_TF_DISABLE is set
-  // we will not do anything; all subsequent
-  // passes become a no-op.
+  // we will not do anything; all subsequent passes become a no-op.
   bool ngraph_not_enabled =
       (!config::IsEnabled()) || (std::getenv("NGRAPH_TF_DISABLE") != nullptr);
   bool already_processed = IsProcessedByNgraphPass(&graph);
+  if (!already_processed && ngraph_not_enabled) {
+    NGRAPH_VLOG(0) << "Graph will not run through ngraph";
```
avijit-nervana: Since this is a VLOG(0) level message - need to be a bit more user friendly. Need to specify why the graph won't run through nGraph (think of a message that an user would get and wonder - what can I do to fix/change this). 
So - use something like: `nGraph is available but disabled`. 
If the graph is alreasy transformed ti nGraph then it won't be further transformed but `WILL run` through nGraph. So this message `Graph will not run through ngraph` is misleading - no?

src/grappler/ngraph_optimizer.cc
```diff
@@ -166,8 +192,44 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+    const char* ng_backend_env_value = std::getenv("NGRAPH_TF_BACKEND");
+    if (ng_backend_env_value != nullptr) {
+      string backend_env = std::string(ng_backend_env_value);
+      if (backend_env.empty() ||
+          !BackendManager::IsSupportedBackend(backend_env)) {
+        return errors::Internal("NGRAPH_TF_BACKEND: ", backend_env,
+                                " is not supported");
+      }
+      backend_name = backend_env;
+      NGRAPH_VLOG(1) << "Setting backend from the enviornment variable "
```
avijit-nervana: Rephrase as: `Overriding backend using environment variable to GPU:2`

src/grappler/ngraph_optimizer.cc
```diff
@@ -166,8 +192,44 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+    }
+    // splits into {"ngraph_backend", "_ngraph_device_config"}
+    config_map = BackendManager::GetBackendAttributes(
+        backend_name);  // SplitBackendConfig
+    backend_name = config_map.at("ngraph_backend");
+    // config_map in EncapsulateClusters is not expected to contain
+    // ngraph_backend
+    config_map.erase("ngraph_backend");
+  }
+  NGRAPH_VLOG(0) << "NGTF_OPTIMIZER: backend_name " << backend_name;
```
avijit-nervana: Since this is VLOG(0) - need to use more user friendly and consistent message. See the above comments for example.

src/ngraph_backend_config.cc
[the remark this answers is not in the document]
avijit-nervana: Since this is an optional attribute - why are we then throwing exception when the attribute not present?

src/ngraph_backend_config.h
```diff
@@ -0,0 +1,57 @@
...
```
avijit-nervana: Use a `const string&`

src/ngraph_backend_manager.h
avijit-nervana: You don't need to use the `make_unnique` API for creating a unique_ptr. Please don't use shared_ptr as this Ptr is not really shared. Use a unique ptr instead.

src/ngraph_encapsulate_op.cc
avijit-nervana: The term `optional_parameter` i misleading as it's not really an optional!

src/ngraph_mark_for_clustering.cc
```diff
@@ -136,8 +136,8 @@ static ConfirmationFunction SimpleConfirmationFunction() {
 //
 // Main entry point for the marking pass.
 //
-Status MarkForClustering(Graph* graph,
-                         const std::set<string> skip_these_nodes) {
+Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
+                         std::string current_backend) {
```
avijit-nervana: Always use `const string&` when passing values

src/ngraph_backend_config.cc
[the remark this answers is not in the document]
shresthamalik: This is more like additional attributes that are backend specific. I can change the name.

src/grappler/ngraph_optimizer.cc
sayantan-nervana: `!already_processed && ngraph_not_enabled` implies will not run thru nGraph
Its not processed AND ngraph is disabled, so no chance of forming encapsulates. So no running thru nGraph.


Comment about helpfulness of message stands

kanvi-nervana: How about `nGraph is disabled, so graph will not be transformed`.

src/ngraph_backend_manager.h
shresthamalik: That's correct. I did change it to create a unique_ptr without using make_unique. We have a map that holds onto the backend-config object ptr, and a utility function GetBackendConfig(string backend_name) that returns the ptr by querying the map. Was running into issues around it. Let me try again.

PR #105, opened 2019-06-10
sarkars/Adding mnist file to build dir

sayantan-nervana:
Assumes that `build_cmake` folder is inside the `ngraph_bridge` dir, and `./test_ngtf.py` is run from there


sayantan-nervana:
https://buildkite.com/ngraph/ngtf-cpu-centos-grappler/builds/1008

PR #92, opened 2019-06-03
Avijit/cpp app

avijit-nervana:
* This PR creates a C++ example of using a pre-trained network to classify images (based on TensorFlow label_image example)
* The single model example (`infer_single`) will be extended to support multiple inference threads in a separate PR.
* This PR also creates a placeholder for a C++ inference example application with multiple inference models.

PR #106, opened 2019-06-11
sarkars/Add back provenance tagging

sayantan-nervana:


CMakeLists.txt
```diff
@@ -291,6 +291,7 @@ get_filename_component(
     ABSOLUTE
 )
 
+add_subdirectory(examples)
```
sayantan-nervana: Adding back this line which was deleted here: https://github.com/tensorflow/ngraph-bridge/pull/92/files#diff-af3b638bc2a3e6c650974192a53c7291R283

src/ngraph_builder.cc
```diff
@@ -95,8 +95,7 @@ std::shared_ptr<TOpType> ConstructNgNode(const std::string& op_name,
                                          TArg&&... Args) {
   auto ng_node = std::make_shared<TOpType>(std::forward<TArg>(Args)...);
   ng_node->set_friendly_name(op_name);
-  // Fails for now, so commenting out for now
-  // ng_node->add_provenance_tag(op_name);
+  ng_node->add_provenance_tag(op_name);
```
avijit-nervana: Do we have enough tests (on other backends) to ensure that this won't crash?

sayantan-nervana: This line gets added in `ngraph_builder.cc` which means all the tests use this. gtests (opexecutor), pytests, tf python tests and resnet tests. Also in CI we run this on CPU and GPU 2 backends.

The issue was more related to running models with `pip install tensorflow` + `build ngraph bridge`
iIf TF is built from source then there is no problem. In fact the crash ws observed even with CPU backend in debian (where they had pip installed tensorflow). So its more a matter of gcc compatibility than backend

PR #107, opened 2019-06-12
Sarkars/build refactor

sayantan-nervana:
1. Introduces `build_tf.py`
2. Changes `build_ngtf` to include a new flag `use_tensorflow_from_location`


test_ngtf.py
```diff
@@ -86,6 +86,9 @@ def main():
         # Run the C++ example build/run test
         run_cpp_example_test('build')
 
+    if (not os.path.isdir(build_dir + '/tensorflow')):
+        download_repo(build_dir + "/tensorflow", "https://github.com/tensorflow/tensorflow.git", "v1.14.0-rc0")
+
```
sayantan-nervana: TODO: verify `test_ngtf` is working. Seems to hang now in tf python tests.

build_ngtf.py
```diff
@@ -173,39 +180,45 @@ def main():
...
-        command_executor(["pip", "install", "-U", "tensorflow==" + tf_version])
-
-        import tensorflow as tf
-        print('Version information:')
-        print('TensorFlow version: ', tf.__version__)
-        print('C Compiler version used in building TensorFlow: ',
-              tf.__compiler_version__)
-        cxx_abi = str(tf.__cxx11_abi_flag__)
+    if arguments.use_tensorflow_from_location != "":
+        print("Using TensorFlow from " + arguments.use_tensorflow_from_location)
```
shresthamalik: assert that use_tensorflow_from_location is a valid path

shresthamalik: Please add a comment on what is being done (if arguments.use_tensorflow_from_location != "":)

sayantan-nervana: Added comments here: https://github.com/tensorflow/ngraph-bridge/pull/107/files#diff-955b7e25a6094527d84683bcc9da8af9R73
and expanded help string of the option: https://github.com/tensorflow/ngraph-bridge/pull/107/files#diff-ece698560cf09ce24af8979586c76638R119

sayantan-nervana:
It seems .py files at the base location (build_ngtf and test_ngtf) were not being formatted or checked by the scripts in `maint`. So added some lines to include those files as well

build_ngtf.py
sayantan-nervana: Added some comments about the mechanics: https://github.com/tensorflow/ngraph-bridge/pull/107/commits/d0b7ae4b0aeb762dfd18529b8a9d0e70e05c47a2

PR #110, opened 2019-06-18
Shrestha/backend manager fixes

shresthamalik:
Changes: 
-  IsSupportedBackend tries to create a backend rather than querying nGraph for registered backends. There were some error conditions arising due to static initialization of the supported_backends map and late registration of backends, which led to this change
- GetCurrentlySetBackendName accounts for the NGRAPH_TF_Backend set
- Update BackendManager Tests


src/ngraph_backend_manager.h
```diff
@@ -47,25 +47,25 @@ struct Backend {
 class BackendManager {
  public:
   // Returns the backend name currently set
-  static string GetCurrentlySetBackendName() {
-    return BackendManager::ng_backend_name_;
-  };
+  // If env variable NGRAPH_TF_BACKEND is set it has precedence
```
kanvi-nervana: It has precedence only over the BackendManager and not the config, correct?

test/python/common.py
```diff
@@ -91,3 +91,29 @@ def generate_random_numbers(self,
         if datatype == "DTYPE_INT":
             return [random.randint(start, end) for i in range(vector_length)]
         return [random.uniform(start, end) for i in range(vector_length)]
+
+    # returns true if the the env variable is set
+    def is_env_variable_set(self, env_var):
+        return env_var in os.environ
+
+    # sets the env variable
+    def set_env_variable(self, env_var, env_var_val):
```
kanvi-nervana: Is this only for setting the backend env variable? If yes, then please change the name to project that.
If not, then please change the `print message`

test/python/test_set_backend.py
```diff
@@ -19,54 +19,68 @@
...
-    ngraph_bridge.enable()
-    backend_cpu = 'CPU'
-    backend_interpreter = 'INTERPRETER'
+class TestSetBackend(NgraphTest):
+
+    def test_set_backend(self):
+        # store env variables
+        # when testing on backends like GPU the tests are run with NGRPAH_TF_BACKEND
+        # by storing and restoring the env_variables we run the tests independent of the backend set
+        # currently we only and restore only NGRPAH_TF_BACKEND
```
kanvi-nervana: typo: ` # currently we only and restore only NGRPAH_TF_BACKEND` ---> ` # currently we only restore NGRPAH_TF_BACKEND`

src/ngraph_backend_manager.cc
```diff
@@ -101,22 +107,52 @@ void BackendManager::UnlockBackend(const string& backend_name) {
...
+
+  // NGRAPH_TF_BACKEND is set
+  string backend_env = std::string(ng_backend_env_value);
+  if (backend_env.empty() || !BackendManager::IsSupportedBackend(backend_env)) {
+    return errors::Internal("NGRAPH_TF_BACKEND: ", backend_env,
+                            " is not supported");
+  }
+
+  *backend_name = backend_env;
+  NGRAPH_VLOG(1) << "Overriding backend using the enviornment variable "
```
sayantan-nervana:
```suggestion
  NGRAPH_VLOG(1) << "Overriding backend using the environment variable "
```

src/ngraph_backend_manager.h
shresthamalik: Yes.

test/python/common.py
shresthamalik: This is for setting any env variable.

PR #116, opened 2019-06-21
Kkasravi/Add support for docker build system

kkasravi:
fixes NGTF-2093

Suggested follow-on improvement:
- allow the base docker image to be pushed to a registry (eg gcr.io/constant-cubist-173123/ngtf) 
  (then build artifacts can be shared across the team.)


kkasravi:
moving this to local branch from forked branch

PR #119, opened 2019-06-21
Add dockerfiles for builds with Debian Stretch python3

crlishka:
Add Debian 9 "Stretch" as a platform for builds and unit-tests.

Successful testing:
tf-ng-bridge-daily-build: https://aipg-rancher.intel.com/jenkins/algo/job/tf-ng-bridge-daily-build/130/


test/ci/docker/docker-scripts/run-build-test-ngtf.sh
```diff
@@ -88,17 +88,29 @@ echo ' '
...
+    echo 'python version is:'
+    python --version
+else
+    echo '/usr/bin/python was not found'
+fi
 
-echo ' '
-echo 'python2 version is:'
-python2 --version
+if [ -f /usr/bin/python2 ] ; then
```
avijit-nervana: Why are we still supporting Python2?

crlishka: It is old code that was already there.  It is simply a check to tell if python2 is installed in the Docker container -- it is useful to know that python2 is *not* installed.  I added a guard around it because python2 support was removed.

The presence of this code does not indicate that we are supporting python2.  I no longer run any python2 builds.

PR #121, opened 2019-06-21
Add HorovodBroadcast op

ami-GS:


src/ngraph_builder.cc
```diff
@@ -1810,6 +1810,18 @@ static Status TranslateDepthwiseConv2dNativeOp(
   return Status::OK();
 }
 
+static Status TranslateBroadcastDistributedOp(
```
avijit-nervana: This should be guarded by `#if defined (NGRAPH_DISTRIBUTED)`

src/ngraph_builder.cc
```diff
@@ -4833,6 +4845,7 @@ const static std::map<
         {"Greater", TranslateBinaryOp<ngraph::op::Greater>},
         {"GreaterEqual", TranslateBinaryOp<ngraph::op::GreaterEq>},
         {"HorovodAllreduce", TranslateAllreduceOp},
```
avijit-nervana: These Horovod ops should be guarded by `#if defined (NGRAPH_DISTRIBUTED)`

src/ngraph_utils.cc
```diff
@@ -372,26 +372,26 @@ bool DumpTrackedGraphs() {
          std::getenv("NGRAPH_TF_DUMP_TRACKED_GRAPHS") != nullptr;
 }
 
-void AllreduceOpControlOrder(
-    const std::shared_ptr<ngraph::Function>& ng_function) {
+void OpControlOrder(const std::shared_ptr<ngraph::Function>& ng_function,
```
avijit-nervana: This should be guarded by `#if defined (NGRAPH_DISTRIBUTED)`

src/ngraph_utils.h
```diff
@@ -322,7 +322,8 @@ bool DumpEncapsulatedGraphs();
 bool DumpTrackedGraphs();
 
 // Insert constrol dependency for AllReduce ops to ensure execution order
-void AllreduceOpControlOrder(const std::shared_ptr<ngraph::Function>&);
+void OpControlOrder(const std::shared_ptr<ngraph::Function>&,
```
avijit-nervana: This should be guarded by `#if defined (NGRAPH_DISTRIBUTED)`

src/ngraph_builder.cc
```diff
@@ -1810,6 +1810,18 @@ static Status TranslateDepthwiseConv2dNativeOp(
...
+static Status TranslateBroadcastDistributedOp(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_input;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input));
+
+  auto ng_broadcast_distributed =
+      ConstructNgNode<ng::op::BroadcastDistributed>(op->name(), ng_input);
+  SaveNgOp(ng_op_map, op->name(), ng_broadcast_distributed);
+  return Status::OK();
```
sayantan-nervana: since this looks like a direct one-to-one map, please consider using `TranslateUnary`. Please see "Abs" translation for an example. https://github.com/tensorflow/ngraph-bridge/blob/master/src/ngraph_builder.cc#L4794

ami-GS: great idea, used

src/ngraph_builder.cc
ami-GS: fixed

PR #122, opened 2019-06-26
Kanvi/config support II

kanvi-nervana:


src/ngraph_backend_manager.h
```diff
@@ -65,6 +65,10 @@ class BackendManager {
 
   static void ReleaseBackend(const string& backend_name);
 
+  static void SetConfig(
+      const string& backend_name,
+      std::unordered_map<std::string, std::string> additional_attributes_map);
```
shresthamalik: map can be made const & too

test/graph_rewrites/config_for_grappler_test.cc
```diff
@@ -116,6 +115,73 @@ TEST(GrapplerConfig, RConfig1) {
...
+    node->set_requested_device("CPU");
+  }
+
+  // Create GraphDef and Grappler
+  grappler::GrapplerItem item;
+  graph.ToGraphDef(&item.graph);
+  ConfigProto config_proto;
+  auto backend_name = AttrValue();
+  backend_name.set_s("INTERPRETER");
+  auto device_config = AttrValue();
```
shresthamalik: The device_config is not used right?

test/graph_rewrites/config_for_grappler_test.cc
```diff
@@ -116,6 +115,73 @@ TEST(GrapplerConfig, RConfig1) {
...
+  // There is only one node in the graph
+  // And it is an NGraphEncapsulateOp
+  ASSERT_EQ(output_graph.num_op_nodes(), 1);
+  Node* ng_encap = nullptr;
+
+  for (auto node : output_graph.op_nodes()) {
+    ng_encap = node;
+  }
+  ASSERT_NE(ng_encap, nullptr);
+  string ng_backend, ng_device_config, ng_test_echo;
```
shresthamalik: same comment here. We can remove it.

test/graph_rewrites/config_for_grappler_test.cc
kanvi-nervana: Right, missed removing it.

PR #125, opened 2019-06-28
Kanvi/Move the registrations for ops to a separate location

kanvi-nervana:


src/ops/ngraph_tracked_variable.cc
```diff
@@ -0,0 +1,38 @@
...
+    .Output("ref: Ref(dtype)")
+    .Attr("shape: shape")
+    .Attr("dtype: type")
+    .Attr("just_looking: bool = false")
+    .Attr("container: string = ''")
+    .Attr("shared_name: string = ''")
+    .SetIsStateful()
+    .SetShapeFn(shape_inference::ExplicitShape);
+
+}  // namespace ngraph_bridge
```
sayantan-nervana: Another option is: we could have a single file for all ngraph ops. For example in TF, all nn ops are in a single file etc.

src/CMakeLists.txt
```diff
@@ -40,11 +40,13 @@ set(SRC
    ngraph_cluster_manager.cc
    ngraph_deassign_clusters.cc
    ngraph_encapsulate_clusters.cc
+   ops/ngraph_encapsulate_op.cc
    ngraph_encapsulate_op.cc
    ngraph_freshness_tracker.cc
    ngraph_mark_for_clustering.cc
    ngraph_rewrite_for_tracking.cc
    ngraph_rewrite_pass.cc
+   ops/ngraph_tracked_variable.cc
```
sayantan-nervana: Also do we need to update this as well? https://github.com/tensorflow/ngraph-bridge/blob/master/bazel/BUILD#L18

src/ops/ngraph_tracked_variable.cc
kanvi-nervana: Yes, we can have something like ngraph_ops.cc and move registrations there.

src/enable_variable_ops/ngraph_tracked_variable.cc
```diff
@@ -283,19 +283,6 @@ void NGraphVariableOp::Compute(OpKernelContext* ctx) {
...
-    .Attr("dtype: type")
-    .Attr("just_looking: bool = false")
-    .Attr("is_tf_just_looking: bool = false")
-    .Attr("copy_to_tf: bool = false")
-    .Attr("container: string = ''")
-    .Attr("shared_name: string = ''")
-    .Attr("ngraph_graph_id: int")
-    .SetIsStateful()
-    .SetShapeFn(shape_inference::ExplicitShape);
-
```
sayantan-nervana: we could remove `#include "tensorflow/core/framework/op_kernel.h"` from this file.

and maybe in other places too wherever applicable.

src/CMakeLists.txt
```diff
@@ -65,6 +66,9 @@ if(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
     list(REMOVE_ITEM SRC ngraph_rewrite_pass.cc)
     list(APPEND SRC enable_variable_ops/ngraph_rewrite_pass.cc)
 
+    list(REMOVE_ITEM SRC ops/ngraph_ops.cc)
+    list(APPEND SRC enable_variable_ops/ops/ngraph_ops.cc)
```
shresthamalik: We have some registerations here too, if you want to take care of them too
https://github.com/tensorflow/ngraph-bridge/blob/master/src/enable_variable_ops/ngraph_variable_modifiers.cc

src/CMakeLists.txt
avijit-nervana: Yes

PR #126, opened 2019-07-01
Sarkars/model test upgrades

sayantan-nervana:
1. Remove comments from `test_main.py` that are already present in the README.
2. Return appropriate exit code in case some tests fail

### DIFF
--- a/test/model_level_tests/test_main.py
+++ b/test/model_level_tests/test_main.py
@@ -17,6 +17,7 @@
 import pdb, time
 from subprocess import check_output, call, Popen, PIPE
 import json, os, argparse, sys
+from functools import reduce
 import sys
 # expects tools to be present at this relative location. Need access to build_utils
 sys.path.insert(0, os.path.abspath('../../tools'))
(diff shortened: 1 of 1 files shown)


sayantan-nervana:
https://buildkite.com/ngraph/ngtf-cpu-model-test/builds/176

test/python/test_conv2D_KernelChecks.py
```diff
@@ -63,7 +63,11 @@ def make_filter_and_backprop_args(self):
 
         return x1, x2
 
-    def test_conv2d_stride_in_batch_not_supported(self):
+    @pytest.mark.parametrize(("strides",), (
+        ([2, 1, 1, 1],),
+        ([2, 1, 1, 1],),
```
sayantan-nervana:
```suggestion
        ([1, 1, 1, 2],),
```

test/python/test_modeltester.py
```diff
@@ -35,8 +36,18 @@ class TestModelTester(NgraphTest):
     def test_MLP(self):
         cwd = os.getcwd()
         os.chdir('../model_level_tests/')
+        grappler = ngraph_bridge.is_grappler_enabled()
+        varopts = ngraph_bridge.are_variables_enabled()
+        if grappler:
+            if varopts:
+                assert False, "Varopts and grappler does not build together right now"
```
sindhu-nervana: Q/Todo: Is this if condition still valid once we merge grappler + varopts

sayantan-nervana: Once varopts and grappler are merged, we should have a CI pipeline for that. then this will fail this assert and remind us to enable this then.

PR #132, opened 2019-07-09
kanvi/implement_encap_output_info_map

kanvi-nervana:


src/enable_variable_ops/ngraph_catalog.cc
```diff
@@ -111,5 +113,53 @@ bool NGraphCatalog::ExistsInInputVariableSharedNameMap(int graphid,
...
+                                               int input_index) {
+  std::string key =
+      NGraphCatalog::CreateNodeKey(graphid, node_name, input_index);
+  auto itr = NGraphCatalog::encap_output_info_map_.find(key);
+  return itr != NGraphCatalog::encap_output_info_map_.end();
+}
+
+tuple<string, bool, bool> NGraphCatalog::GetInfoFromEncapOutputInfoMap(
+    string key) {
+  return NGraphCatalog::encap_output_info_map_[key];
```
shresthamalik: optional: we can change all the Get api's to use map.at(key) instead of map[key] as they expect the key to be present

PR #138, opened 2019-07-11
Sarkars/merge grappler varopts

sayantan-nervana:
**Change summary**
[Struck out sections were merged in another PR: https://github.com/tensorflow/ngraph-bridge/pull/254]

~1. Fix i'th pytest running ngoptimizer i times:~
    ~1.a `update_config` checks if ngraph optimizer is enabled, if so then simply return (instead of appending another request for ngraph optimizer)~
    ~1.b fixed `common.py` by removing mutable default arguments to functions~

2. Make grappler and var/opt work together:
    2.a Add `ReplaceModifiers` and `EnterInCatalog`, `RemoveNGraphAssigns` in `src/grappler/ngraph_optimizer.cc`
    2.b `AddIdentityN` has been modified to also add `IdentityN` to ref outputs
    2.c If a variable has been replaced before, dig up its old shared name and use that for later replacements. Take care that `AddIdentityN` changes variables names (if they are output nodes), so take care to use the name of the newly added `IdentityN` node to find the shared name from the catalog (see `FindSharedNameOfPreviouslyReplacedVariable`)

3. Turn axpy variable update into a pytest

~4. Fix learning rate for SGD so that the `mnist_deep_simplified.py` produces reasonable results (approx 95%) as opposed to around 20% it is doing now~

**Limitations of current implementation**
have a failing test (currently disabled) in `test/python/test_assign.py`. In normal path, we are registered at `POST_PLACEMENT` to `CaptureVariables`. We can see the whole graph there. So we can choose to not accept variables that are touched (directly or indirectly) by assigns that can change shape. But in grappler, we only register at a single point, and at that point we only see partial graphs. So the unit test fails (and is skipped). In the future, when var/opts code is modified to accept variables with assigns that change shape (by maybe throwing away the underlying ng tensor and replacing it with a new one), this test will pass.

**Tests**
1. CI
- [x] Normal CI
2. Local MNIST convergence tests
2.1 Varopt 0, grappler 0
- [x] CPU with SGD (supported optimizer)
- [x] GPU with SGD (supported optimizer)
- [x] CPU with adam (unsupported optimizer)
- [x] GPU with adam (unsupported optimizer)
2.2 Varopt 1, grappler 0
- [x] CPU with SGD (supported optimizer) with varopts only
- [x] GPU with SGD (supported optimizer) with varopts only
- [x] CPU with adam (unsupported optimizer) with varopts only
- [x] GPU with adam (unsupported optimizer) with varopts only
2.3 Varopt 0, grappler 1
- [x] CPU with SGD (supported optimizer) with grappler only
- [x] GPU with SGD (supported optimizer) with grappler only
- [x] CPU with adam (unsupported optimizer) with grappler only
- [x] GPU with adam (unsupported optimizer) with grappler only
2.4 Varopt 1, grappler 1
- [x] CPU with SGD (supported optimizer) with both grappler and varopts
- [x] GPU with SGD (supported optimizer) with both grappler and varopts
- [x] CPU with adam (unsupported optimizer) with both grappler and varopts
- [x] GPU with adam (unsupported optimizer) with both grappler and varopts
3. Spectral lines
- [ ] Check spectral lines
4. Personal CI builds
- [x] GPU ci with varopts only
- [x] GPU ci with varopts + grappler


examples/mnist/mnist_deep_simplified.py
```diff
@@ -172,7 +172,7 @@ def train_mnist_cnn(FLAGS):
         if FLAGS.optimizer == "adam":
             train_step = tf.train.AdamOptimizer(1e-4).minimize(cross_entropy)
         elif FLAGS.optimizer == "sgd":
-            train_step = tf.train.GradientDescentOptimizer(1e-4).minimize(
+            train_step = tf.train.GradientDescentOptimizer(5e-2).minimize(
```
sayantan-nervana: earlier with sgd, we were converging to 20%. Increased LR to converge to >95% now. Just so that the sample script provides some reasonable mnist numbers.

python/ngraph_bridge/__init__.in.py
```diff
@@ -201,13 +201,21 @@ def is_grappler_enabled():
...
     #updating session config if grappler is enabled
+    opt_name = 'ngraph-optimizer'
     if(ngraph_bridge_lib.ngraph_tf_is_grappler_enabled()):
+        # If the config already has ngraph-optimizer, then do not update it
+        if config.HasField('graph_options'):
+            if config.graph_options.HasField('rewrite_options'):
+                custom_opts = config.graph_options.rewrite_options.custom_optimizers
+                for i in range(len(custom_opts)):
+                    if custom_opts[i].name == opt_name:
+                        return config
```
sayantan-nervana: in update_config, check if ngraph-optimizer is already requested.

test/python/common.py
```diff
@@ -47,8 +47,10 @@ def import_pbtxt(self, pb_filename):
             tf.import_graph_def(graph_def)
         return graph
 
-    def with_ngraph(self, l, config=tf.ConfigProto()):
+    def with_ngraph(self, l, config=None):
```
sayantan-nervana: mutable default values in python function caused `update_config` to be called on the same config object multiple times, causing the i'th test to run ngraph optimizer i times

examples/test_2_assigns.py
```diff
@@ -0,0 +1,23 @@
...
+config = tf.ConfigProto()
+sess = tf.Session(config=ngraph_bridge.update_config(config))
+
+sess.run(v.initializer)
+
+sess.run(new_v_0)
+assert (10 == v.eval(session=sess))
+
+sess.run(new_v_1)
+assert (([10, 20] == v.eval(session=sess)).all())
```
sayantan-nervana: This will fail for now. TODO: figure this out

src/grappler/ngraph_add_identityn.cc
```diff
@@ -24,37 +24,30 @@ namespace ngraph_bridge {
...
+              node->output_type(i) -
+              (IsRefType(node->output_type(i)) ? kDataTypeRefOffset : 0)));
           inputs.push_back(NodeBuilder::NodeOut(node, i));
         }
 
-        if (ref_type) {
-          NGRAPH_VLOG(5)
-              << "NGTF_OPTIMIZER: Cannot construct an IdentityN node";
-          continue;
-        }
```
sayantan-nervana: now we can attach identityN, even on ref type outputs.

sayantan-nervana: if a ref type is detected subtract 100 (kDataTypeRefOffset) from it to get the corresponding type

src/grappler/ngraph_add_identityn.cc
```diff
@@ -24,37 +24,24 @@ namespace ngraph_bridge {
...
-                           << " at index " << i << " is ref type";
-            ref_type = true;
-            break;
-          }
-          input_types.push_back(node->output_type(i));
+          // DT_X_REF = DT_X + 100. kDataTypeRefOffset = 100 here
+          // https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/types.proto
+          input_types.push_back(static_cast<DataType>(
+              node->output_type(i) -
+              (IsRefType(node->output_type(i)) ? kDataTypeRefOffset : 0)));
```
sayantan-nervana: if a ref type is detected subtract 100 (kDataTypeRefOffset) from it to get the corresponding type

sayantan-nervana:
GPU ci: https://buildkite.com/ngraph/ngtf-gpu-ubuntu-18-04/builds/1268

sayantan-nervana:
Updated resnet numbers here: https://jira.devtools.intel.com/browse/NGTF-1964

ngraph_bridge/CMakeLists.txt
```diff
@@ -86,7 +86,12 @@ endif()
 
 
 if(NGRAPH_TF_USE_GRAPPLER_OPTIMIZER)            
-    list(REMOVE_ITEM SRC ngraph_rewrite_pass.cc)
+    if(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
+        list(REMOVE_ITEM SRC enable_variable_ops/ngraph_rewrite_pass.cc)
+    else()
+        list(REMOVE_ITEM SRC ngraph_rewrite_pass.cc)
+    endif()
+    list(APPEND SRC enable_variable_ops/ngraph_catalog.cc)
```
kanvi-nervana: Shouldn't this be only appended `if(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)` ?

ngraph_bridge/ngraph_utils.cc
```diff
@@ -408,6 +426,51 @@ bool IsProcessedByNgraphPass(Graph* g) {
...
+  } else {
+    original_var_node_name = node->name();
+  }
+  std::tie(has_been_replaced_before, shared_name_of_replacement) =
+      NGraphCatalog::HasTFVarBeenReplacedBefore(original_var_node_name);
+  if (has_been_replaced_before) {
+    shared_name = shared_name_of_replacement;
+  }
+  TF_RETURN_IF_ERROR(
+      NGraphCatalog::RegisterTFVarReplacement(node->name(), shared_name));
```
sayantan-nervana: Should `node->name()` be `original_var_node_name` ?

ngraph_bridge/ngraph_capture_variables.cc
```diff
@@ -61,17 +61,27 @@ Status CaptureVariables(Graph* graph, const std::set<string> skip_these_nodes) {
...
         }
 
+        string shared_name;
+        shared_name = shared_name_from_current_var.empty()
+                          ? node->name()
+                          : shared_name_from_current_var;
+
         Node* replacement;
+#if (NGRAPH_TF_USE_GRAPPLER_OPTIMIZER) && \
+    (NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
```
sayantan-nervana: `NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS` can be removed

ngraph_bridge/CMakeLists.txt
kanvi-nervana: No, since ngraph_catalog is also needed for the grappler path.

ngraph_bridge/ngraph_capture_variables.cc
```diff
@@ -61,17 +61,26 @@ Status CaptureVariables(Graph* graph, const std::set<string> skip_these_nodes) {
...
+          shared_name_from_current_var = "";
         }
 
+        string shared_name;
+        shared_name = shared_name_from_current_var.empty()
+                          ? node->name()
+                          : shared_name_from_current_var;
+
         Node* replacement;
+#if (NGRAPH_TF_USE_GRAPPLER_OPTIMIZER)
```
kanvi-nervana: Explain why this is needed for just the `grappler` option?

ngraph_bridge/enable_variable_ops/ngraph_catalog.h
```diff
@@ -38,6 +38,12 @@ namespace ngraph_bridge {
 
 class NGraphCatalog {
  private:
+#if (NGRAPH_TF_USE_GRAPPLER_OPTIMIZER)
+  // Map of tensorflow variable name that has been replaced to a NGV to shared
```
shresthamalik:
```suggestion
  // Map of tensorflow variable name that has been replaced to a NGraphVariable, and its shared name
```

ngraph_bridge/enable_variable_ops/ngraph_capture_variables.cc
```diff
@@ -40,7 +43,13 @@ static bool NGraphPlacementRequested(const Node* node) { return true; }
 // Main entry point for the variable-capture.
 //
 Status CaptureVariables(Graph* graph, std::set<string> skip_these_nodes) {
-  const static std::map<
+  auto ReplaceVariableWithIdentityInfo =
+      std::bind(ReplaceVariable, _1, _2, _3, _4, _5, _6, _7, _8, _9, _10,
+                skip_these_nodes);
+  // This map is no longer static, since we need to do a partial application on
+  // ReplaceVariable. Note it is static in RewriteForTracking, since there the
+  // partial application is done with a constant argument that will not change
```
shresthamalik: Does this comment have to do with binding?

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -31,6 +35,11 @@ namespace ngraph_bridge {
 // Main entry point for rewrite-for-tracking.
 //
 Status RewriteForTracking(Graph* graph, int graph_id) {
+  // The last argument is not used when a NG var is replaced by another NG var,
```
shresthamalik: Requires "NGRAPH_USE_GRAPPLER_OPTIMIZER" flag

ngraph_bridge/grappler/ngraph_add_identityn.cc
```diff
@@ -24,37 +24,24 @@ namespace ngraph_bridge {
...
         for (int i = 0; i < node->num_outputs(); i++) {
-          if (IsRefType(node->output_type(i))) {
-            NGRAPH_VLOG(5) << "NGTF_OPTIMIZER: "
-                           << "Datatype for the node output"
-                           << " at index " << i << " is ref type";
-            ref_type = true;
-            break;
-          }
-          input_types.push_back(node->output_type(i));
+          // DT_X_REF = DT_X + 100. kDataTypeRefOffset = 100 here
```
shresthamalik:
```suggestion
           // If the input-type is ref type, making it non ref type
          // DT_X_REF = DT_X + 100. kDataTypeRefOffset = 100 here
```

ngraph_bridge/enable_variable_ops/ngraph_capture_variables.cc
sayantan-nervana: bind creates a std::function that is resolvable only at runtime (since `skip_these_nodes` is unknown before hand), hence we cannot use it to initialize the map

parallel situation is:
```
x = rand()
static map<int, int> x = {1:2, 3:x};
```

bazel/BUILD
```diff
@@ -70,6 +71,7 @@ cc_binary(
         "ngraph_bridge/tf_deadness_analysis.cc",
         "ngraph_bridge/tf_graphcycles.cc",
         "ngraph_bridge/version.cc",
+        "ngraph_bridge/enable_variable_ops/ngraph_catalog.cc",
```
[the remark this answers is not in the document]
sayantan-nervana: These files are added as only grappler build is possible to build using bazel build. we do not build varopts with bazel afaik. Currently we need catalog to build grappler build as well, hence only adding that file (but not replace modifiers etc, which are not needed for pure grappler build)

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -195,27 +195,42 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
     DumpGraphs(graph, idx, "captured", "Graph With Variables Captured");
   }
 
+#if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)
+  // 0. Replace optimizers then, if requested, dump the graphs.
+  TF_RETURN_IF_ERROR(ReplaceModifiers(&graph, idx));
```
sayantan-nervana: see if addidentityn can be added after capturevar to reduce tackiness

ngraph_bridge/enable_variable_ops/ngraph_capture_variables.cc
```diff
@@ -84,6 +95,19 @@ Status CaptureVariables(Graph* graph, std::set<string> skip_these_nodes) {
...
+// In case of grappler check if we have already captured a variable of this name
+#if (NGRAPH_TF_USE_GRAPPLER_OPTIMIZER)
+          auto tmp_pair =
+              NGraphCatalog::HasTFVarBeenReplacedBefore(node->name());
+          if (get<0>(tmp_pair)) {
+            // return errors::Internal(node->name(),
+            //                       "(VariableV2) was captured in an earlier "
+            //                     "graph, but in the current graph ngraph "
+            //                   "was unable to capture it");
+          }
```
sayantan-nervana: TODO, fix formatting

ngraph_bridge/enable_variable_ops/ngraph_catalog.h
```diff
@@ -38,6 +38,12 @@ namespace ngraph_bridge {
...
+  // name
+  static unordered_map<string, string> tf_var_name_to_shared_name_map_;
```
sayantan-nervana: Discuss: Perhaps we do not need a map of var name -> shared_name
All we need is a set of varnames
(because the current variable under consideration can tell us its shared name (assuming shared names do not change))

PR #141, opened 2019-07-15
Kanvi/fix regression some models

kanvi-nervana:
Perf regression was see in Unet, Draw and Yolo. This fixes it.

### DIFF
--- a/src/ngraph_backend_config.cc
+++ b/src/ngraph_backend_config.cc
@@ -36,7 +36,12 @@ string BackendConfig::Join(
   } catch (std::out_of_range e1) {
     throw std::out_of_range("Attribute device_config not found");
   }
-  return backend_name_ + ":" + additional_parameters.at("device_config");
+  string backend_name = backend_name_;
+  if (additional_parameters.at("device_config") != "") {
+    backend_name =
+        backend_name + ":" + additional_parameters.at("device_config");
+  }
+  return backend_name;
 }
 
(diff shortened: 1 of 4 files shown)


src/ngraph_backend_config.cc
```diff
@@ -36,7 +36,12 @@ string BackendConfig::Join(
   } catch (std::out_of_range e1) {
     throw std::out_of_range("Attribute device_config not found");
   }
-  return backend_name_ + ":" + additional_parameters.at("device_config");
+  string backend_name = backend_name_;
+  if (additional_parameters.at("device_config") != "") {
+    backend_name =
+        backend_name + ":" + additional_parameters.at("device_config");
+  }
+  return backend_name;
```
shresthamalik: I would think the backend should be able to take in the string we were sending without any issue, but seems not.
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/backend.hpp#L56

src/ngraph_backend_manager.cc
```diff
@@ -39,7 +39,6 @@ unordered_set<string> BackendManager::ng_supported_backends_(
 
 map<std::string, int> BackendManager::ref_count_each_backend_;
 
-mutex BackendManager::ng_backendconfig_map_mutex_;
```
avijit-nervana: This was causing regressions?

kanvi-nervana: No, this is not used so removed it.

PR #143, opened 2019-07-17
Move the RemoveEdge() call outside the loop to avoid crash in debug builds

aaroey:
TF disallows modifying the edgeset when iterating it, see:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/graph/edgeset.h#L117


sayantan-nervana:
> LGTM. There are other places we do something similar. We can take an action items to correct those.

https://github.com/tensorflow/ngraph-bridge/blob/a4171750638ecc7cf0c700ef3b7d55f6138eb35d/src/enable_variable_ops/ngraph_replace_op_utilities.cc#L197

https://github.com/tensorflow/ngraph-bridge/blob/2a8c066912ffff0c5afdc1c27c94ba0f288748d7/src/ngraph_rewrite_for_tracking.cc#L95

https://github.com/tensorflow/ngraph-bridge/blob/2a8c066912ffff0c5afdc1c27c94ba0f288748d7/src/ngraph_rewrite_for_tracking.cc#L103


These ones are done right:
https://github.com/tensorflow/ngraph-bridge/blob/2a8c066912ffff0c5afdc1c27c94ba0f288748d7/src/ngraph_rewrite_for_tracking.cc#L103

https://github.com/tensorflow/ngraph-bridge/blob/a4171750638ecc7cf0c700ef3b7d55f6138eb35d/src/enable_variable_ops/ngraph_replace_op_utilities.cc#L211

https://github.com/tensorflow/ngraph-bridge/blob/6a542fbd938167ebcff8730ce416792817f2b84a/src/ngraph_capture_variables.cc#L103

PR #146, opened 2019-07-18
sarkars/destructor order

sayantan-nervana:
The bridge holds on to executable in 4 places. When release backend is called, backend releases its shared ptr to executables, and then releases backend. But that doesn't destruct the executable since the bridge is still holding on the the shared pointers.

Hence clearing the maps to make sure when release backend call is made, the backend's shared ptr to executable has `use_count()==1`


ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -230,6 +230,9 @@ class NGraphEncapsulateOp : public OpKernel {
     }
 #endif
```
sayantan-nervana:
```suggestion
    m_ng_exec_input_cache_map.clear();
    m_ng_exec_output_cache_map.clear();
```

PR #117, opened 2019-06-21
Kasravi/Add support for docker build system

kkasravi:
fixes NGTF-2093

Suggested follow-on improvement:

- allow the base docker image to be pushed to a registry (eg gcr.io/constant-cubist-173123/ngtf)
(then build artifacts can be shared across the team.)

PR #148, opened 2019-07-18
Extend docker build automation to support passing build-options to build_ngtf.py

crlishka:
Add a method of passing optional parameters for build_ngtf.py and ng_tfp.y into the automation running it in a docker container.  Also, add an optional run of unit-tests with PlaidML activated.  This is needed so that we can support building with CPU-backend only, or CPU+INTELGPU+PLAIDML backends in one wheel.

The changes in this PR ...
- ... are used in cje-algo PR#665, which adds switches to the daily-build jobs to optionally build and unit-test with the PlaidML backend;
- ... are for implementing stories NGVALID-996 and NGVALID-1014.


kkasravi:
@crlishka I think this has some overlap with #117 - perhaps you can take a look?
Seems like ci tooling and dev tooling should converge

crlishka:
> @crlishka I think this has some overlap with #117 - perhaps you can take a look?
> Seems like ci tooling and dev tooling should converge

@kkasravi  This PR is a targeted change to allow the validation team's daily-build automation to pass command-line options to build_ngtf.py.  The changes in this PR should be orthogonal to your PR#117.  While I agree it might be good to coverge the tooling, it is outside the scope of this particular PR.

PR #150, opened 2019-07-22
Kanvi/remove additional attr check

kanvi-nervana:
Make ngraph_backend & device_id as a mandatory attribute to be supplied using rewriter_config.
Make all other backend related attributes optional.
All attributes thhat will have '_ngraph_' appended will be passed to the specified backend and it is upto the backend to do the checks for the one's it expects.
Remove ngraph_backend_config class as it is no longer required. We can now use common Split and Join functions for all backends.
When using the grappler, rewriter_config is the only way to specify backend, env variable and backend manager will not work. Hence, removed code for the same from ngraph_optimizer.cc
Fix tests.


ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -44,21 +44,23 @@ Status NgraphOptimizer::Init(
...
-        BackendManager::GetBackendAdditionalAttributes(config_backend_name);
-    for (size_t i = 0; i < additional_attributes.size(); i++) {
-      if (params.count(additional_attributes[i])) {
-        config_map["_ngraph_" + additional_attributes[i]] =
-            params.at(additional_attributes[i]).s();
-        NGRAPH_VLOG(3) << additional_attributes[i] << " "
-                       << config_map["_ngraph_" + additional_attributes[i]];
+    if (params.count("device_id")) {
+      config_backend_name = params.at("ngraph_backend").s();
+      NGRAPH_VLOG(3) << config_backend_name;
```
sayantan-nervana:
```suggestion
      NGRAPH_VLOG(3) << "Backend name from config: " << config_backend_name;
```

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -44,21 +44,23 @@ Status NgraphOptimizer::Init(
...
+      for (auto i : params) {
+        if (i.first != "ngraph_backend") {
+          config_map["_ngraph_" + i.first] = i.second.s();
+          NGRAPH_VLOG(3) << "Attribute: " << i.first
+                         << " Value: " << config_map["_ngraph_" + i.first];
+        }
       }
+    } else {
+      return errors::Internal(
+          "NGTF_OPTIMIZER: device_id is a required parameter");
```
sayantan-nervana: even `ngraph_backend` is a compulsory attr right?

Lets structure this section like this, so its clear what our contract is. Something like:

```
compulsory_attrs = ["ngraph_backend", "devide_id"]
for (auto itr : compulsory_attrs) {
  if params.count("device_id") == 0: {
      return error_status;
  }
}

compulsory_config_map = {}
optional_extra_config_map = {}
for (auto itr : params) {
if (itr in compulsory_attrs) {
  compulsory_config_map[itr] = params[itr];
} else {
  optional_extra_config_map[itr] = params[itr];
}
}

// Note we split config_map into compulsory_config_map and optional_extra_config_map
// So later on we will not need this line
// config_map.erase("ngraph_backend");
```

with this one can add/remove from the list of compulsory attrs easily later on

sayantan-nervana: we'd need to make device_id compulsory:
https://github.com/tensorflow/ngraph-bridge/blob/kanvi/remove_additional_attr_check/ngraph_bridge/ops/ngraph_ops.cc#L24

which means when creating the encapsulate op here, we have to pass the device id.
https://github.com/tensorflow/ngraph-bridge/blob/kanvi/remove_additional_attr_check/ngraph_bridge/ngraph_encapsulate_clusters.cc#L338
Note right now device is populated using `cluster_backend`, lets discuss tomorrow how we are going to pass the other compulsory attr `device-id`.

test/graph_rewrites/config_for_grappler_test.cc
```diff
@@ -73,8 +71,8 @@ TEST(GrapplerConfig, RConfig1) {
   ConfigProto config_proto;
   auto backend_name = AttrValue();
   backend_name.set_s("CPU");
-  auto device_config = AttrValue();
-  device_config.set_s("1");
+  auto device_id = AttrValue();
```
shresthamalik: General: If time permits we can add a test checks for failure if backend or device_id is not specified. The current test check agreement with implementation.

test/python/test_updateconfig.py
```diff
@@ -0,0 +1,46 @@
...
+        config.allow_soft_placement = True
+        config_new = ngraph_bridge.update_config(config)
+        rewriter_options = config_new.graph_options.rewrite_options
+        ngraph_optimizer_name = rewriter_options.custom_optimizers[0].name
+        assert ngraph_optimizer_name == 'ngraph-optimizer'
+        ngraph_optimizer = rewriter_options.custom_optimizers[0]
+        ngraph_optimizer.parameter_map["max_batch_size"].s = b'64'
+        ngraph_optimizer.parameter_map["ice_cores"].s = b'12'
+        assert config_new.__str__(
+        ) == 'allow_soft_placement: true\ngraph_options {\n  rewrite_options {\n    meta_optimizer_iterations: ONE\n    min_graph_nodes: -1\n    custom_optimizers {\n      name: "ngraph-optimizer"\n      parameter_map {\n        key: "device_id"\n        value {\n          s: ""\n        }\n      }\n      parameter_map {\n        key: "ice_cores"\n        value {\n          s: "12"\n        }\n      }\n      parameter_map {\n        key: "max_batch_size"\n        value {\n          s: "64"\n        }\n      }\n      parameter_map {\n        key: "ngraph_backend"\n        value {\n          s: "CPU"\n        }\n      }\n    }\n  }\n}\n'
```
shresthamalik: this needs formatting

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -43,22 +43,25 @@ namespace ngraph_bridge {
...
-      if (params.count(additional_attributes[i])) {
-        config_map["_ngraph_" + additional_attributes[i]] =
-            params.at(additional_attributes[i]).s();
-        NGRAPH_VLOG(3) << additional_attributes[i] << " "
-                       << config_map["_ngraph_" + additional_attributes[i]];
+  for (size_t i = 0; i < compulsory_attrs.size(); i++) {
+    if (params.count(compulsory_attrs[i]) == 0) {
+      NGRAPH_VLOG(0) << "NGTF_OPTIMIZER: Compulsory attribute "
+                     << compulsory_attrs[i] << " not found.";
+      return errors::Internal("NGTF_OPTIMIZER: Missing compulsory attributes.");
```
shresthamalik: Might not need the log as we are throwing the error. The error can have the missing attribute name

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -159,25 +160,33 @@ class NGraphEncapsulateOp : public OpKernel {
...
+    for (auto itx : additional_attributes) {
+      // Find the optional attributes to be sent to the backend.
+      // The optional attributes have '_ngraph_' appended to the start
+      // so we need to get rid of that and only send the remaining string
+      // since the backend will only look for that.
+      // '_ngraph_' is only appended for the bridge.
+      // For e.g. _ngraph_ice_cores --> ice_cores
+      if (itx.first.find("_ngraph_") != std::string::npos) {
+        // find the index of the second '_'
+        int delimiter_index = itx.first.find('_', 6);
```
shresthamalik: minor you dont need the delimiter

itx.first.substr(strlen("_ngraph_")) should give you the attribute name, and you get rid of the hardcoded 6

test/python/test_updateconfig.py
```diff
@@ -0,0 +1,48 @@
+# ==============================================================================
+#  Copyright 2018-2019 Intel Corporation
```
kanvi-nervana:
```suggestion
#  Copyright 2019 Intel Corporation
```

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -43,22 +43,25 @@ namespace ngraph_bridge {
...
+    }
+  }
+  config_backend_name = params.at("ngraph_backend").s();
+  NGRAPH_VLOG(3) << "Backend name from config: " << config_backend_name;
+  for (auto i : params) {
+    if (i.first != "ngraph_backend") {
+      if (i.first == "device_id") {
+        config_map["ngraph_" + i.first] = i.second.s();
+      } else {
+        config_map["_ngraph_" + i.first] = i.second.s();
```
sayantan-nervana: `config_map[(i.first == "device_id" ? "" : "_") + "ngraph_" + i.first] = i.second.s();`

ngraph_bridge/ngraph_backend_manager.h
```diff
@@ -122,9 +121,8 @@ class BackendManager {
   // 1. GetBackendCreationString("GPU", {"_ngraph_device_config", "2"})
   // returns "GPU:2"
   // throws an error if the required attributes are not present in the map
-  static string GetBackendCreationString(
```
sayantan-nervana: the comment above can go away now:
```
// 1. GetBackendCreationString("GPU", {"_ngraph_device_config", "2"})
  // returns "GPU:2"
  // throws an error if the required attributes are not present in the map
```

test/python/test_updateconfig.py
kanvi-nervana: I did run formatting, but it was left unchanged

shresthamalik: might have to do it manually. python strings are not auto formatted

sayantan-nervana: yapf did not break it down since its a single string. lets leave it this way for now since its sort of a "binary"

sayantan-nervana: whoa, i did not notice how loooong it was.. maybe we should break it down by hand

sayantan-nervana: After some experimentation we see some problems with breaking up the string, so keeping it as is right now

test/graph_rewrites/config_for_grappler_test.cc
kanvi-nervana: Added a test for the failure case

PR #151, opened 2019-07-23
sarkars/Upgrade ngraph version to 24rc0

sayantan-nervana:


sayantan-nervana:
varopt + gpu build: https://buildkite.com/ngraph/ngtf-gpu-ubuntu-18-04/builds/1327

PR #161, opened 2019-07-26
sarkars/Prep for 0.17.0rc0

sayantan-nervana:


python/setup.in.py
```diff
@@ -59,7 +59,7 @@ def get_tag(self):
 
 setup( 
     name='ngraph_tensorflow_bridge',
-    version='0.16.0rc3',
+    version='0.17.0rc0',
```
shresthamalik: anyone knows why we dont have a dash here like this ? 0.17.0-rc0

PR #167, opened 2019-07-26
Log parser fix to add the last subgraph to the results

dmsuehir:
As the log parser loops through lines in the log, it adds a subgraph's info to the results dictionary when it comes across the start of a new subgraph. The last subgraph was getting missed, since it wasn't getting added to the results dictionary. This PR fixes that issue by adding the last subgraph to the results dictionary at the end.


dmsuehir:
I'll resubmit this PR after I get the CLA

PR #179, opened 2019-07-31
Vishakha/softplus

vishakha-nervana:
Added Softplus Op

### DIFF
--- a/ngraph_bridge/ngraph_builder.cc
+++ b/ngraph_bridge/ngraph_builder.cc
@@ -2513,6 +2513,22 @@ static Status TranslateLogSoftmaxOp(
   return Status::OK();
 }
 
+static Status TranslateSoftplusOp(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_inp;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_inp));
+  auto ng_exp = ConstructNgNode<ng::op::Exp>(
+      op->name(), ng_inp);
+  auto constant_1 = ConstructNgNode<ng::op::Constant>(
+      op->name(), ng_inp->get_element_type(), ng_inp->get_shape(),
+      std::vector<std::string>(ng::shape_size(ng_inp->get_shape()), "1"));
+  auto ng_output = ConstructNgNode<ng::op::Log>(
+      op->name(), ConstructNgNode<ng::op::Add>(op->name(), ng_exp, constant_1));
+  SaveNgOp(ng_op_map, op->name(), ng_output);
+  return Status::OK();
+}
+
 static Status TranslateMatMulOp(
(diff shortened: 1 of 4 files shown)


ngraph_bridge/ngraph_builder.cc
```diff
@@ -4870,7 +4885,7 @@ const static std::map<
       {"Sigmoid", TranslateSigmoidOp}, {"SigmoidGrad", TranslateSigmoidGradOp},
       {"Size", TranslateSizeOp}, {"Sign", TranslateUnaryOp<ngraph::op::Sign>},
       {"Slice", TranslateSliceOp}, {"Snapshot", TranslateIdentityOp},
-      {"Softmax", TranslateSoftmaxOp},
+      {"Softmax", TranslateSoftmaxOp}, {"Softplus", TranslateSoftplusOp},
```
sayantan-nervana: Shift to new line

test/test_nn_ops.cpp
```diff
@@ -1575,6 +1575,30 @@ TEST(NNOps, SoftmaxZeroDimTest2) {
...
+    vector<DataType> output_datatypes = {DT_FLOAT};
+    std::vector<Output> sess_run_fetchoutputs = {R};
+
+    OpExecuter opexecuter(root, "Softplus", static_input_indexes,
+                          output_datatypes, sess_run_fetchoutputs);
+
+    opexecuter.RunTest();
+  }
+}
+
```
sayantan-nervana: I think u had a pytest too, but that was not committed

test/python/test_softplus.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def test_softplus(self):
+        x = tf.placeholder(tf.float32, shape=(2, 3))
+        y = tf.placeholder(tf.float32, shape=(2, 3))
+        z = tf.placeholder(tf.float32, shape=(2, 3))
+
+        a = x + y + z
+        b = x + y + z
+        c = a * b
+        d = tf.nn.softplus(c)
```
sayantan-nervana: Lets run this pytest locally with `NGRAPH_TF_LOG_PLACEMENT=1` to ensure that a cpp softplus is produced (instead of breaking down into other ops like exp/log etc).

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2513,6 +2513,21 @@ static Status TranslateLogSoftmaxOp(
   return Status::OK();
 }
 
+static Status TranslateSoftplusOp(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_inp;
```
sayantan-nervana: https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/softplus.html

`log(exp(features) + 1)`

kanvi-nervana: May be add this - `log(exp(features) + 1)` as a comment for future reference.

PR #183, opened 2019-08-05
Sindhu/bfloat16 op tests

sindhu-nervana:
This PR implements new tests listed below to test bfloat16 conversion
1. test_conv2d.py 
2. test_conv2dbackpropfilter_nchw.py
3. test_conv2dbackpropfilter_nhwc.py
4. test_conv2dbackpropinput_nchw.py
5. test_conv2dbackpropinput_nhwc.py


test/python/bfloat16/test_conv2d.py
```diff
@@ -0,0 +1,99 @@
...
+W = 5
+C = 1
+
+filter = np.random.rand(1, 1, 1, 2)
+input_size_nhwc = [N, H, W, C]
+input_size_nchw = [N, C, H, W]
+input_nhwc = tf.placeholder(tf.float32, shape=input_size_nhwc, name='x')
+input_nchw = tf.placeholder(tf.float32, shape=input_size_nchw, name='x')
+
+n_np = np.random.rand(1, 1, 3, 5).astype('f')
```
shresthamalik: use N,H,W,C

test/python/bfloat16/test_conv2d.py
```diff
@@ -0,0 +1,99 @@
...
+# Filter shape is allowed
+# If any fail, then we should place Op on CPU for compute
+
+#Inputs
+N = 1
+H = 3
+W = 5
+C = 1
+
+filter = np.random.rand(1, 1, 1, 2)
```
shresthamalik: Please check if filter needs to be casted to bfloat

test/python/bfloat16/test_conv2dbackpropfilter_nchw.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+#Ngraph Graph
+def ng_model(padding):
+    t1 = tf.placeholder(dtype=tf.float32, shape=input_sizes_nchw, name='t1')
+    t2 = tf.constant(filter_size_hwio, dtype=tf.int32, name='t1')
+    t3 = tf.placeholder(
+        dtype=tf.float32, shape=out_backprop_in_sizes[padding], name='t3')
+
+    filt = nn_ops.conv2d_backprop_filter(
```
shresthamalik: do we need to transpose the output of either tf or ngraph before comparing?

test/python/bfloat16/test_conv2dbackpropinput_nhwc.py
```diff
@@ -0,0 +1,107 @@
...
+np.random.seed(5)
+#Inputs
+N = 1
+H = 7
+W = 6
+C = 2
+
+I = C
+O = 2
+FW = 3
```
shresthamalik: Please change the variable names like filter/input/FW etc

shresthamalik:
Tests are quite neat, Thanks! 👍

test/python/bfloat16/test_conv2dbackpropfilter_nchw.py
sindhu-nervana: No, output is same shape as filter which is a constant. the shapes are same for both TF and Ngraph graph

test/python/bfloat16/test_conv2d.py
sindhu-nervana: thanks for pointing this out. yes, it does. changed it

test/python/bfloat16/test_conv2d.py
```diff
@@ -0,0 +1,105 @@
...
+import os
+
+import ngraph_bridge
+
+#Test Ngraph Op Convolution, TF Op:conv2d
+# Implemented based on NNP's unit test TEST(test_assign_layout, convolution_special_case)
+
+np.random.seed(5)
+
+# Colvolution Op is placed on NNP and conerted to bfloat16 only for the special case below, otherwise it falls back to CPU for compute
```
kanvi-nervana:
```suggestion
# Colvolution Op is placed on NNP and conerted to bfloat16 only for the special case below, 
# otherwise it falls back to CPU for compute
```

sindhu-nervana:
```suggestion
# Colvolution Op is placed on NNP and conerted to 
# bfloat16 only for the special case below, otherwise it falls 
# back to CPU for compute
```

test/python/bfloat16/test_conv2d.py
```diff
@@ -0,0 +1,105 @@
...
+t_np = np.transpose(n_np, (0, 2, 3, 1))
+
+
+#TF graph
+def tf_model():
+    stride_nhwc = [1, 2, 2, 1]
+    x = tf.cast(input_nhwc, dtype=tf.bfloat16)
+    filter_cast = tf.cast(filter_size, dtype=tf.bfloat16)
+    m = tf.nn.conv2d(
+        x, filter_cast, stride_nhwc, "SAME", data_format="NHWC", name="m")
```
kanvi-nervana: Should we also have a test where `padding = VALID` ?

test/python/bfloat16/test_conv2d.py
```diff
@@ -0,0 +1,107 @@
...
+
+#Ngraph graph
+def ng_model():
+    stride_nchw = [1, 1, 2, 2]
+    m = tf.nn.conv2d(
+        input_nchw,
+        filter_size,
+        stride_nchw,
+        "SAME",
+        data_format="NCHW",
```
kanvi-nervana: Do we need a test for `data_format=NHWC`?

test/python/bfloat16/test_conv2dbackpropinput_nchw.py
```diff
@@ -0,0 +1,115 @@
...
+    log_device_placement=False,
+    inter_op_parallelism_threads=1)
+
+
+@pytest.mark.parametrize("padding", ("VALID", "SAME"))
+def test_conv2dbackpropinput_nchw(padding):
+    np_filter = np.random.rand(*filter_size_hwio).astype('f')
+    n_np_out = np.random.rand(*out_backprop_in_sizes[padding]).astype('f')
+    #Reshape to NHWC for TF
+    t_np_out = np.transpose(n_np_out, (0, 2, 3, 1))
```
kanvi-nervana: You are transposing `out_backprop` to `NHWC` here as well https://github.com/tensorflow/ngraph-bridge/pull/183/files#diff-d7fb93f2ec39b43164446ba0e32fec3fR59
So this can be removed.

test/python/bfloat16/test_conv2d.py
sindhu-nervana: This is a special case which backend has so does not work for VALID or rather does not convert.

test/python/bfloat16/test_conv2d.py
sindhu-nervana: same comment as above

test/python/bfloat16/test_conv2dbackpropinput_nchw.py
sindhu-nervana: no, both are different. one is input shape and the other is np array shape with random data. so if i remove one of the transpose, feed_dict will fail with incorrect shapes.

PR #197, opened 2019-08-09
sarkars/Upgrade master to 25rc3, bring back failing tests, and add ignore

sayantan-nervana:


tools/test_utils.py
```diff
@@ -124,7 +124,9 @@ def run_ngtf_pytests(venv_dir, build_dir):
     command_executor(["pip", "install", "-U", "pytest"])
     command_executor(["pip", "install", "-U", "psutil"])
 
-    cmd = 'python -m pytest ' + ('--junitxml=%s/xunit_pytest.xml' % build_dir)
+    cmd = 'python -m pytest ' + (
```
avijit-nervana: What's the plan on including this when we need the float tests?

sayantan-nervana: right now on master we have `--ignore` in `run_ngtf_pytests_from_artifacts`. to prevent these tests from running (and failing) on CPU ci

This PR adds `--ignore` to `run_ngtf_pytests`, which is used in `test_ngtf.py`. So this just brings `run_ngtf_pytests` up to speed with `run_ngtf_pytests_from_artifacts`.

regd plans of removing the `--ignore` for CI: @shresthamalik might be able to comment on that. For now these tests are used by stakeholders who are interested in them, but we ignore them in CI

README.md
```diff
@@ -88,7 +88,7 @@ Once TensorFlow's dependencies are installed, clone the `ngraph-bridge` repo:
 
         git clone https://github.com/tensorflow/ngraph-bridge.git
         cd ngraph-bridge
-        git checkout v0.18.0-rc1
+        git checkout v0.18.0-rc2
```
shresthamalik: our rc-2 uses ngraph rc3?

sayantan-nervana: yes. we started with ngrc1 (ngrc0 never had a tag from us)

PR #199, opened 2019-08-09
sarkars/Fix extra_params

sayantan-nervana:
Earlier `extra_params` was parsed by `parse_extra_params_string` to make it accept simple strings like `--extra_params {a:1}`. But due to how argparse handles string, more than 1 entry dictionaries would not go through, ie, `--extra_params {a:1,b:2}` throws an error in argparse

hence updating `parse_extra_params_string` to use json. 

Also since we are using json and not parsing the string ourselves, removing unit tests that tested the parsing mechanism.

Also `extra_params` is called `optional_params`


sayantan-nervana:
rename to `--optional_params` and pass a json file

test/python/test_tf2ngraph_functions.py
```diff
@@ -58,35 +58,3 @@ def test_config_updater_api(self):
             'ngraph_backend': 'CPU',
             'device_id': '0'
         }
-
-    # In this test we test that the spaces are handled correctly
-    @pytest.mark.parametrize(('ng_device',),
-                             (('{abc:1,def:2}',), ('{abc:1, def:2}',),
-                              ('{abc :1,def: 2}',), ('{abc:1,def: 2} ',)))
-    def parse_extra_params_string(self, raw_string):
```
sayantan-nervana: removing these chunk of tests, sicne now we do not string parse, instead we use the json library

test/python/test_tf2ngraph_script.py
```diff
@@ -73,20 +73,21 @@ def test_command_line_api(self, inp_format, inp_loc, out_format,
             pass
         conversion_successful = False
         try:
-            extra_params = {
-                'CPU': '{device_config:0}',
-                'INTERPRETER': '{test_echo:1}'
+            optional_params_json = {
+                'CPU': 'sample_optional_params.json',
+                'INTERPRETER': 'sample_optional_params.json'
```
sayantan-nervana: replaced string dictionaries with jsons

PR #205, opened 2019-08-13
Fix resource leaking due to undeleted ngraph::Executable

XWangIntel:
There must be some kind of cyclic reference for shared_ptr\<ngraph::Executable\> and I'm still looking for the method to break it.
Before that the resource leaking can be fixed by replacing shared_ptr\<ngraph::Executable\> which raw pointer. It is not just a workaround but also reduce the overhead of incrementing shared_ptr use count, which can reach several hundreds and totally unnecessary. Therefore should be good to have anyway.

### DIFF
--- a/ngraph_bridge/ngraph_freshness_tracker.cc
+++ b/ngraph_bridge/ngraph_freshness_tracker.cc
@@ -36,7 +36,7 @@ void NGraphFreshnessTracker::MarkFresh(
 
 bool NGraphFreshnessTracker::IsFresh(
     const void* base_pointer,
-    std::shared_ptr<ngraph::runtime::Executable> user) {
+    const std::shared_ptr<ngraph::runtime::Executable>& user) {
   mutex_lock l(mu_);
   auto it = freshness_map_.find(base_pointer);
   if (it == freshness_map_.end()) {
@@ -69,9 +69,9 @@ void NGraphFreshnessTracker::RemoveTensor(const void* base_pointer) {
 }
 
 void NGraphFreshnessTracker::RemoveUser(
-    std::shared_ptr<ngraph::runtime::Executable> user) {
+    const std::shared_ptr<ngraph::runtime::Executable>& user) {
   mutex_lock l(mu_);
-  for (auto kv : freshness_map_) {
+  for (auto& kv : freshness_map_) {
     kv.second.erase(user.get());
   }
 }
--- a/ngraph_bridge/ngraph_freshness_tracker.cc
+++ b/ngraph_bridge/ngraph_freshness_tracker.cc
@@ -42,7 +42,7 @@ bool NGraphFreshnessTracker::IsFresh(
   if (it == freshness_map_.end()) {
     return false;
   } else {
-    return true; 
+    return (it->second.count(user.get() > 0);
   }
 }
 
--- a/ngraph_bridge/ngraph_freshness_tracker.cc
+++ b/ngraph_bridge/ngraph_freshness_tracker.cc
@@ -26,11 +26,11 @@ namespace ngraph_bridge {
 
 void NGraphFreshnessTracker::MarkFresh(
     const void* base_pointer,
-    std::shared_ptr<ngraph::runtime::Executable> user) {
+    std::shared_ptr<ngraph::runtime::Executable>& user) {
   mutex_lock l(mu_);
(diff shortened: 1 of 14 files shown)


ngraph_bridge/ngraph_freshness_tracker.cc
```diff
@@ -69,10 +69,10 @@ void NGraphFreshnessTracker::RemoveTensor(const void* base_pointer) {
 }
 
 void NGraphFreshnessTracker::RemoveUser(
-    std::shared_ptr<ngraph::runtime::Executable> user) {
+    const std::shared_ptr<ngraph::runtime::Executable>& user) {
   mutex_lock l(mu_);
-  for (auto kv : freshness_map_) {
-    kv.second.erase(user);
+  for (auto& kv : freshness_map_) {
```
XWangIntel: change `kv` to `& kv` is the fix for this problem. The rest of change is just for performance and simplicity.
In short, if using `kv`, we just erase user from the copy of `ngraph::Executable` set. The original freshness_map is untouched. Period.

XWangIntel:
@googlebot I signed it!

ngraph_bridge/ngraph_freshness_tracker.cc
shresthamalik: Good Catch! :)

ngraph_bridge/ngraph_freshness_tracker.h
```diff
@@ -114,15 +114,15 @@ class NGraphFreshnessTracker : public ResourceBase {
...
+  void RemoveUser(const std::shared_ptr<ngraph::runtime::Executable>& user);
 
  private:
   // mutex protecting the freshness_map_
   mutex mu_;
   // for each base pointer (of tensor), maintains a set of ng::Functions that
   // use it
   // Each ng function in the set is then a user of the base_pointer
-  std::map<const void*, std::set<std::shared_ptr<ngraph::runtime::Executable>>>
+  std::map<const void*, std::set<ngraph::runtime::Executable*>>
```
avijit-nervana: We don't want to save raw pointers - as they may become null under the hood. We need to save a unique_ptr or shared_ptr. But in this case we are getting a shared_ptr from nGraph - so we are going to stay with shared_ptr.

ngraph_bridge/ngraph_freshness_tracker.cc
avijit-nervana: Good catch! Thanks for debugging this :)

ngraph_bridge/ngraph_freshness_tracker.h
XWangIntel: @avijit-nervana My concern is about performance. I see the use count of ngraph::Executable goes up to hundreds. And the increment/decrement is implemented with atomic operation and therefore the overhead is quit significant.
If we are sure that the lifespan of executable is longer than freshness tracker, it is still worth the change. You are definitely more familiar with that and can make the decision.

avijit-nervana: @XWangIntel I agree with you on this. I personally do not like shared ptrs just for this reason. We will do this clean up a few weeks later - so will track this definitely.

avijit-nervana:
There are compilation errors. See here: https://buildkite.com/ngraph/external-fork/builds/2#dab399bd-427a-4da4-a272-4d06d19707c2/6-6206

Please run build_ngtf.py and test_ngtf.py to ensure that it builds and sanity tests pass.

XWangIntel:
@googlebot I consent

XWangIntel:
> There are compilation errors. See here: https://buildkite.com/ngraph/external-fork/builds/2#dab399bd-427a-4da4-a272-4d06d19707c2/6-6206
> 
> Please run build_ngtf.py and test_ngtf.py to ensure that it builds and sanity tests pass.

Sorry this change is only verified on full stack build of NNP-transformer, and then moved to the new repo. Will verify that in standalone repo.

XWangIntel:
@googlebot I consent

XWangIntel:
@googlebot I consent!

XWangIntel:
cla:yes

XWangIntel:
@googlebot

XWangIntel:
@googlebot I verified it!

XWangIntel:
Submit the same PR from the repo instead of forked one.

PR #209, opened 2019-08-16
Fixed the memory leak

avijit-nervana:
Replaced the raw pointer with a uique_ptr.


shresthamalik:
In grappler build C++ test was failing. triggered it again

PR #215, opened 2019-08-27
sarkars/Fix TF folder

sayantan-nervana:
Some changes to make the model testing pipeline work
1. Fix prebuilt TF path
2. Fix data paths for nmt
3. Remove unet from tests and filed a ticket for it: https://jira.devtools.intel.com/browse/NGTF-2257


sayantan-nervana:
https://buildkite.com/ngraph/ngtf-cpu-model-test/builds/325

PR #218, opened 2019-08-28
vishakha/Lock1 Experiment

vishakha-nervana:
This branch contains first experimental changes, in Compute() of graph_encapsulate_op.cc,
Modifying lock to optimize  perf.


avijit-nervana:
OBE.

PR #219, opened 2019-08-28
Abhiram/ngraph viz merge master

abhiram-nervana:
Changes made for a grappler build

Changes in ngraph_optimizer.cc to be able to dump NGraphs
Changes in ngraph_utils.cc/h to be able to save NGraphs and scalar graphs (for statistics) to specified directory
Changes in ngraph_encapsulate_op.cc to be able to save statistics


abhiram-nervana:
@googlebot I fixed it.

abhiram-nervana:
> maybe add a unit test (which would also double as a code example).
> 
> see pytests in test/python. Maybe the unit test starts a training and u can set os.environ['NGRAPH_TF_TB_LOGDIR'] through python before starting the training.
> 
> In the end u can check if the LOGDIR is populated with event files etc
> 
> also think about if the functions u introduced, eg `DumpNGraph` etc can be unit tested by gtests (see in test folder)

Done

abhiram-nervana:
@googlebot I fixed it.

ngraph_bridge/ngraph_utils.cc
```diff
@@ -463,6 +475,201 @@ bool IsProcessedByNgraphPass(Graph* g) {
...
+  {
+    mkdir(str_path.c_str(), 0777);
+  }
+
+  std::string dname = GetSessionName(file_idx, nodes);
+  str_path += (dname.insert(0, "ngraph") + "/");
+
+  if (stat(str_path.c_str(), &buffer) != 0)  // path doesn't exist
+  {
+    mkdir(str_path.c_str(), 0777);
```
sayantan-nervana: minor:
is 0777 the appropriate mode? Can we have a less permissive mode?

same comment for other 0777s

sayantan-nervana: I think I see this code snippet repeated. Can we move it to a function?
```
if (stat(str_path.c_str(), &buffer) != 0)  // path doesn't exist
  {
    mkdir(str_path.c_str(), 0777);
}
```

abhiram-nervana: will move to a separate function

kept 0777 because i wanted user to not worry about any permission issues for their experiments. if change is needed, what permissions do we want for user/group/other? rwxrwx---? or rwx------?

abhiram-nervana: doing 0750. user can rwx, group can rw, other can't do anything

ngraph_bridge/ngraph_utils.cc
```diff
@@ -463,6 +475,198 @@ bool IsProcessedByNgraphPass(Graph* g) {
...
+    return errors::Internal("Error writing graph def to event file");
+  }
+
+  return Status::OK();
+}
+
+void AddSessionNameAttr(int file_idx, std::set<string> nodes, Graph* graph) {
+  for (auto node : graph->op_nodes()) {
+    if (node->type_string() == "NGraphEncapsulate") {
+      node->AddAttr(("_session_name" + to_string(file_idx)),
```
sayantan-nervana: Since you are adding an attribute here are some things to take care of:

1. we have a sort-of convention that we begin optional attr names that we add with `_ngraph`. so maybe name it `_ngraph_session_name + ...`
2. Whatever you name it, make sure to filter it out in `ParseNodeAttributes`: https://github.com/tensorflow/ngraph-bridge/blob/d7a0b4a033302a9305ae418e419901f083c343bb/ngraph_bridge/ngraph_encapsulate_impl.cc#L476

abhiram-nervana: no changes because "_ngraph" prefix to attribute will send attribute to backend. don't want to do that for "_session_name"

ngraph_bridge/ngraph_utils.cc
```diff
@@ -463,6 +475,198 @@ bool IsProcessedByNgraphPass(Graph* g) {
   return false;
 }
 
+Status DumpNGraph(int file_idx, tensorflow::GraphDef* graph_def,
```
sayantan-nervana: So this dumps an event right?
Lets rename this function to reflect that.

We already have an existing function called `DumpGraphs`, so we do not want to confuse with that

ngraph_bridge/ngraph_utils.cc
```diff
@@ -463,6 +475,198 @@ bool IsProcessedByNgraphPass(Graph* g) {
...
+  return Status::OK();
+}
+
+Status CreateSummaryFromGraph(tensorflow::Graph* graph,
+                              std::string filename_prefix) {
+  // convert Graph* to GraphDef*
+  tensorflow::GraphDef* gdef = nullptr;
+  graph->ToGraphDef(gdef);
+
+  // create event summary writer
```
sayantan-nervana: Since u convert to graph def, lets use `CreateSummaryFromGraphDef` inside `CreateSummaryFromGraph` to avoid code duplication

ngraph_bridge/ngraph_utils.cc
```diff
@@ -463,6 +475,198 @@ bool IsProcessedByNgraphPass(Graph* g) {
...
+  if (nodes.size() == 0) {
+    name = to_string(file_idx);
+  } else {
+    std::string entry;
+
+    for (auto it = nodes.begin(); it != nodes.end();
+         ++it)  // get last element in set
+    {
+      entry = *it;
+    }
```
sayantan-nervana: I *think* you are trying to find the topologically last node here, yes?

sets do not have last elements. If the last node happens to be the last node of the graph, that is coincidental and we should not rely upon that behaviour.

If u want the last node, you could use toposort: https://github.com/tensorflow/ngraph-bridge/blob/76055645afbbe1a6041bf468937467e998f787fb/ngraph_bridge/ngraph_builder.cc#L4898

Also consider that graph could have 2 "last" nodes. like X<--A-->Y has 2 "last" nodes X and Y

ngraph_bridge/ngraph_utils.cc
```diff
@@ -463,6 +475,198 @@ bool IsProcessedByNgraphPass(Graph* g) {
...
+      name = to_string(file_idx) + "_" + entry.substr(0, scope_idx);
+    } else {
+      name = to_string(file_idx) + "_" + entry;
+    }
+  }
+
+  return name;
+}
+
+void MakeTBDir(std::string str_path) {
```
sayantan-nervana: try this: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/file_util.hpp#L60

Then you don't have to write a new function

ngraph_bridge/ngraph_utils.cc
sayantan-nervana: Also is the `CreateSummaryFromGraph` function used anywhere? If not we can remove it completely

test/python/test_tensorboard_ngraph_stats.py
```diff
@@ -0,0 +1,212 @@
...
+            saver.save(sess, './mnist_trained/')
+
+        dirs = [o for o in os.listdir(graph_location)]
+
+        check = (len(dirs) == 8) and ("ngraph0_init" in dirs) and (
+            "ngraph1_accuracy" in dirs) and ("ngraph2_loss" in dirs) and (
+                "ngraph3_save" in dirs) and ("stats0_init" in dirs) and (
+                    "stats1_accuracy" in dirs) and ("stats2_loss" in dirs)
+
+        assert (check)
```
sayantan-nervana: i would assert them indivisually, so that the error message can reflect which one of the many tests actually failed.

I mean:
```
assert len(dirs) == 8
assert "ngraph0_init" in dirs
...
```

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -272,12 +273,21 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+      &graph);  // embed the names of those directories in graph as an attribute
+
   // Convert the graph back to Graphdef
   graph.ToGraphDef(output);
   // According to the doc, the message takes ownership of the allocated object
   // https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#proto3_string
   // Hence no need to free fdeflib_new
   output->set_allocated_library(fdeflib_new);
+
+  DumpNGraph(idx, output, fetch_nodes);  // dump NGraphs to tfevent files,
```
shresthamalik: Rename the function to AddGraphToEventFile

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -272,12 +273,21 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
                "Graph with Variables Rewritten for Tracking");
   }
 
+  AddSessionNameAttr(
```
shresthamalik: Add a utility function is_tensorboard_diagnostics_requested and call these functions only if that is true.

You can check for the environment flag in that utility

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -111,6 +111,10 @@ NGraphEncapsulateOp::NGraphEncapsulateOp(OpKernelConstruction* ctx)
   int graph_id{-1};
   OP_REQUIRES_OK(ctx, ctx->GetAttr("ngraph_graph_id", &graph_id));
   ng_encap_impl.SetGraphId(graph_id);
+
+  OP_REQUIRES_OK(ctx, ctx->GetAttr(("_session_name" + to_string(graph_id)),
```
shresthamalik: Same thing guard the function calls, on whether TB diagnostics is requested.

ngraph_bridge/ngraph_utils.cc
shresthamalik: change attr name to _ngraph_tb_session_name

test/python/test_tensorboard_ngraph_stats.py
```diff
@@ -0,0 +1,212 @@
...
+
+                _, summary, loss = sess.run([train_step, merged, cross_entropy],
+                                            feed_dict={
+                                                x: batch[0],
+                                                y_: batch[1],
+                                                keep_prob: 0.5
+                                            })
+                loss_values.append(loss)
+                train_writer.add_summary(summary, i)
+
```
shresthamalik: Please add comments mentioning the changes you have made and why

ngraph_bridge/ngraph_utils.cc
```diff
@@ -463,6 +475,198 @@ bool IsProcessedByNgraphPass(Graph* g) {
...
+void AddSessionNameAttr(int file_idx, std::set<string> nodes, Graph* graph) {
+  for (auto node : graph->op_nodes()) {
+    if (node->type_string() == "NGraphEncapsulate") {
+      node->AddAttr(("_session_name" + to_string(file_idx)),
+                    GetSessionName(file_idx, nodes));
+    }
+  }
+}
+
+std::string GetSessionName(int file_idx, std::set<std::string> nodes) {
```
shresthamalik: Add a comment on what is being done here

ngraph_bridge/ngraph_encapsulate_op.h
```diff
@@ -42,6 +42,7 @@ class NGraphEncapsulateOp : public OpKernel {
  private:
   NGraphEncapsulateImpl ng_encap_impl;
   std::mutex m_compute_lock;
+  std::map<int, std::string> session_names;
```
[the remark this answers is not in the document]
shresthamalik: the session name is linked to the graph id? Is it linked to the encap op? I am wondering if this map belongs here

ngraph_bridge/ngraph_utils.cc
abhiram-nervana: i tried with a few examples, it seemed that the last node always was the highest ordered node that was requested. i know sets dont have an ordering, but the order in which the elements were inserted into the set remain the same, correct? i was thinking this ordering is preserved (when the graph nodes are added to the set), and am simply selecting the last one. let me know if this logic makes sense

ngraph_bridge/grappler/ngraph_optimizer.cc
abhiram-nervana: already have an environment flag for tensorboard (NGRAPH_TF_TB_LOGDIR). i'll check if this exists, and if so, do AddSessionNameAttr

ngraph_bridge/ngraph_encapsulate_op.cc
abhiram-nervana: see comment above

ngraph_bridge/ngraph_utils.cc
abhiram-nervana: if I change the attr name to _ngraph_tb_session_name, the attr will get moved to the backend. was talking to sayantan about this today - do we need to move it to the backend? if i don't add the _ngraph_ portion, it won't get moved. let me know

ngraph_bridge/ngraph_utils.cc
sayantan-nervana: It seems that we have a simple solution to this problem:

If you look in this file, there is a part where we do:

```
  std::set<string> fetch_nodes;
  for (const string& f : item.fetch) {
    int pos = f.find(":");
    fetch_nodes.insert(f.substr(0, pos));
  }```

Now `item.fetch` happens to be a vector. Which means, `fetch_nodes` in the above code could be a vector. That solves the problem of ordering.

It seems `GetSessionName`'s `nodes` argument is always called with fetch nodes. So we can rename it `fetch_nodes` in `GetSessionName`, to be explicit that we are passing a vector of fetch nodes and not all nodes.

Now, fetch nodes are the "last" nodes. they are "incomparable" in a partial ordering. I think if u pass a vector of fetch nodes, ALL of them should be used to create the string. But maybe I am misunderstanding this piece of code. Lets f2f tomorrow.

Comment (abhiram-nervana) [reply ngraph_bridge/ngraph_utils.cc]: picking random node in set as per discussion just now

File: ngraph_bridge/grappler/ngraph_optimizer.cc (hunk 9 above)
Comment (shresthamalik) [reply ngraph_bridge/grappler/ngraph_optimizer.cc]: yes, the check for the env flag can be made into a utility function, as you will have to do it many times

File: ngraph_bridge/ngraph_utils.cc (hunk 2 above)
Comment (shresthamalik) [reply ngraph_bridge/ngraph_utils.cc]: ok. Lets rename it _tensorboard_session_name for now. so that we know it was added for tensorboard

File: ngraph_bridge/ngraph_encapsulate_op.h [hunk 14]
```diff
@@ -42,6 +42,7 @@ class NGraphEncapsulateOp : public OpKernel {
  private:
   NGraphEncapsulateImpl ng_encap_impl;
   std::mutex m_compute_lock;
+  std::map<int, std::string> session_names; // key: graph_id, value: session name (for session's graph, and Encapsulate usage statistics)
```
shresthamalik: This map can be made a static map in the utils. Doesnt need to be a private member in the encap. as it ties up graph id, with session name. It is independent of the encap op

shresthamalik: Add a function with the static map inside it, we dont need a global map.
Something like this
https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_builder.cc#L323

diagnostics/README.md
```diff
@@ -48,6 +48,15 @@ if ngraph is enabled by calling ```ngraph.is_enabled()```
 * _Caution_: The above functions are only effective at the beginning of the execution. Once the session is created and ```run``` is called, the above functions will not be able to disable ngraph. 
 * For example usage, take a look at the ```model_test/verify_model.py``` in the diagnostics folder
 
+### Visualizing encapsulates and encapsulate usage statistics in TB in real-time
+* Run your script with this flag: ```NGRAPH_TF_TB_LOGDIR="<directory>" python run_TF_network.py```. During runtime, you can start TensorBoard separately by running ```tensorboard --logdir <directory>```.
```
sayantan-nervana: Lets mention it works with a grappler only build

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -272,12 +273,23 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
                "Graph with Variables Rewritten for Tracking");
   }
 
+  // TODO: sayantan, see if these changes can be moved into rewrite_pass.cc
+  AddSessionNameAttr(
```
shresthamalik: These need to be guarded with TBDirExists()

ngraph_bridge/grappler/ngraph_optimizer.cc
```diff
@@ -272,12 +273,23 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+      idx, fetch_nodes,
+      &graph);  // embed the names of those directories in graph as an attribute
+
   // Convert the graph back to Graphdef
   graph.ToGraphDef(output);
   // According to the doc, the message takes ownership of the allocated object
   // https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#proto3_string
   // Hence no need to free fdeflib_new
   output->set_allocated_library(fdeflib_new);
+
```
shresthamalik: These need to be guarded with TBDirExists()

ngraph_bridge/ngraph_encapsulate_op.h
```diff
@@ -42,6 +42,9 @@ class NGraphEncapsulateOp : public OpKernel {
  private:
   NGraphEncapsulateImpl ng_encap_impl;
   std::mutex m_compute_lock;
+  std::map<int, std::string> session_names;  // key: graph_id, value: session
```
shresthamalik: Map is not specific to the Op

sindhu-nervana: +1

PR #220, opened 2019-08-28
Sarkars/fix dump file name

sayantan-nervana:
1. Modify `NgraphSerialize` to replace `/` with `--`. So file names like `tf_function_import/x.json` can be dumped correctly. Such file names are generated because TF node names can have `/` due to scoping
2. Added a pytest UT that fails on master, but passes with this PR, to check that the `/` replacement is in place
3. Make `NgraphSerialize` return status and use it appropriately.

TODO: check if `/` is the only special character that needs to be taken care of in TF node names.
The valid names seem to be: `[A-Za-z0-9.][A-Za-z0-9_.\\-/]*`
`.` is another non-alphanumeric char, but once `/` are replaced by `--`, `.` is fine in a file name.

JIRA: https://jira.devtools.intel.com/browse/NGTF-2227


test/python/test_serialize_name.py
```diff
@@ -36,15 +36,10 @@ def get_tensor(self, graph, tname, tag):
...
-    @pytest.mark.parametrize(('import_name_tag',), (
-        (None,),
-        ("",),
-        ("import/scope1/scope2",),
-        ("hello",),
-        ("hello/",),
-        ("hello//",),
-        ("hello//a",),
-    ))
+    @pytest.mark.parametrize(
```
avijit-nervana: Restore the formatting.

PR #222, opened 2019-08-30
Add verbose option to the log parser

dmsuehir:
This PR updates the log parser to grab more values out of the logs, which map to the Avijit's spreadsheet. I added a `verbose` parameter to the `parse_logs()` function, which defaults to `False`, so it won't affect the existing behavior of the model_level_tests, or any other scripts that use this log parser.

This is being used with a new Jenkins job that I have a PR for at [cje-algo](https://github.intel.com/AIPG/cje-algo/pull/738) where the models are run and the logs are parsed with the verbose option on. The build artifacts have .json files with the results.


dmsuehir:
@sayantan-nervana @avijit-nervana 
Could you please review this PR when you get a chance? Thanks!

adk9:
LGTM, thanks.

PR #223, opened 2019-09-03
Kanvi/Bazel for cpp examples

kanvi-nervana:


examples/multisession/BUILD
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    copts = [
+        "-pthread", 
+        "-std=c++11", 
+        # "-D_FORTIFY_SOURCE=2",
+        # "-Wformat",
+        # "-Wformat-security",
+        # "-Wformat",
+        # "-fstack-protector-all",
+        "-D NDEBUG",
```
sayantan-nervana: This has `NDEBUG`. We should remove it right?

kanvi-nervana: Yes `NDEBUG` should be removed.
On another note Sample app is a duplicate of hello_tf, so it can be removed altogether.

bazel/ngraph.BUILD
```diff
@@ -314,17 +271,21 @@ cc_binary(
...
         "-D NGRAPH_TBB_ENABLE",
         '-D PROJECT_ROOT_DIR=\\"\\"',
+        '-D NGRAPH_CPU_STATIC_LIB_ENABLE'
     ] + CXX_ABI,
     linkopts = [
         "-Wl,-z,noexecstack",
         "-Wl,-z,relro",
         "-Wl,-z,now",
+        "-Wl,-Bsymbolic-functions",
+        "-Wl,--exclude-libs=ALL",
```
sayantan-nervana: If the meaning/purpose of the flags are not obvious from the bazel documentation, then can you please add some code review comments what they do and what problem these flag solved. for posterity.

configure_bazel.sh
```diff
@@ -36,6 +36,7 @@ TF_CXX_ABI=( $(python -c 'import tensorflow as tf; print(" ".join(str(tf.__cxx11
 
 write_action_env_to_bazelrc "TF_HEADER_DIR" ${TF_CFLAGS:2}
 write_action_env_to_bazelrc "TF_SHARED_LIBRARY_DIR" ${TF_LFLAGS:2}
+write_to_bazelrc "build --define=grpc_no_ares=true"
```
sayantan-nervana: eg, what does this flag do

examples/multisession/sample.cpp
```diff
@@ -0,0 +1,84 @@
...
+  tensorflow::ClientSession session(root, options);
+
+  std::vector<tensorflow::Tensor> outputs;
+  session.Run({R}, &outputs);
+
+  // Print the output
+  std::cout << "Result: " << outputs[0].matrix<float>() << std::endl;
+
+  // Close the session.
+  // session.Close();
```
sayantan-nervana: We can uncomment this line?

examples/multisession/sample.cpp
```diff
@@ -0,0 +1,84 @@
...
+  tensorflow::SessionOptions options;
+  options.config.mutable_graph_options()
+      ->mutable_optimizer_options()
+      ->set_opt_level(tensorflow::OptimizerOptions_Level_L0);
+  options.config.mutable_graph_options()
+      ->mutable_rewrite_options()
+      ->set_constant_folding(tensorflow::RewriterConfig::OFF);
+
+#if 0
+  // The following is related to Grapler - which we are turning off
```
sayantan-nervana:
```suggestion
  // The following is related to Grappler - which we are turning off
```

examples/cpp/hello_tf.cpp
```diff
@@ -0,0 +1,130 @@
...
+            << (tensorflow::ngraph_bridge::ngraph_tf_are_variables_enabled()
+                    ? std::string("Yes")
+                    : std::string("No"))
+            << std::endl;
+
+  PrintAvailableBackends();
+}
+
+int main(int argc, char** argv) {
+  PrintVersion();
```
sayantan-nervana: If we mean this app to run only on grapple build, we should assert tensorflow::ngraph_bridge::ngraph_tf_is_grappler_enabled()

bazel/ngraph.BUILD
kanvi-nervana: These are not bazel specific flags. But a quick google search gave this:
`Bsymbolic-functions`. Binds references to all global function symbols in a program to the definitions within a user's shared library.
`--exclude-libs ALL` excludes symbols in all archive libraries from automatic export. https://linux.die.net/man/1/ld
May be we should not have this. @avijit-nervana any thoughts?

examples/cpp/hello_tf.cpp
kanvi-nervana: Actually, this is made sure while building the app. The app's build rule has `ngraph_bridge_lib` as a dependency which has `NGRAPH_TF_USE_GRAPPLER_OPTIMIZER` specified as the build flag. So, we don't need the assert.

configure_bazel.sh
avijit-nervana: This has to do with how we re pulling in TF code and building it. When we invoke `tf_workspace(...)` I needed to add this to get the tf workspace setup going.

examples/multisession/sample.cpp
avijit-nervana: This is removed.

PR #224, opened 2019-09-04
sarkars/Remove format check for centos

sayantan-nervana:


sayantan-nervana:
Changes already in here:
https://github.com/tensorflow/ngraph-bridge/commit/ebf00905b8091ddecbd0ff15469cc4dad74ca487

PR #227, opened 2019-09-05
sarkars/Add a comparator to GetReversePostOrder to make toposort deterministic

sayantan-nervana:
Making the translation process more deterministic.

For a graph, Y<-A->X, both A,X,Y and A,Y,Z are valid toposorts. This PR adds a comparator to make the sort more deterministic, so that only one of the above 2 possibilities are chosen consistently.

https://jira.devtools.intel.com/browse/NGTF-2191


ngraph_bridge/ngraph_builder.cc
```diff
@@ -4915,7 +4915,7 @@ Status Builder::TranslateGraph(
   // ought to be `const Node*`, but GetReversePostOrder doesn't use `const`
 
   vector<Node*> ordered;
-  GetReversePostOrder(*input_graph, &ordered);
+  GetReversePostOrder(*input_graph, &ordered, NodeComparatorName());
```
avijit-nervana: Add a unit test to validate this.

sayantan-nervana: `GetReversePostOrder` is a TF function. It stable comparator implementation and testing are in the TF repo. eg, here: https://github.com/tensorflow/tensorflow/blob/c42d974e7abb7380722a6f383084a84d8f3b75b0/tensorflow/core/graph/algorithm_test.cc#L155

Therefore I did not add any unit tests.

PR #228, opened 2019-09-05
Sarkars/use in-graph ngfunc string instead of deserializing

sayantan-nervana:
Main change:
1. In case of AOT, we deserialize and serialize the ng function. No need to do that, we can directly pass the ng function to file.

Other changes:
1. Added an overload of `NgraphSerialize` that accepts a string to dump to file, rather than an ng_function itself
3. Removed some unused functions from impl, since they were not being used by ngraph_encapsulate_op.cc (such as `SetNgFunctionMap`)
4. Unified all the exec map clearing functions into one function `ClearExecMaps`
5. Some functions in impl like `GetNumberOfCopies`, `SetNumberOfCopies` and `GetGraphId` are only used in var-opts path. so moved them under`#if defined(NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS)`
6. Corrected the behaviour or `SetCopyLog` and renamed it to `AppendCopyLog`
7. Added a new function `DumpNgFunction` that will dump the ng function in case of errors. It hides if its actually dumping the ngfunction or the serialized string (in case of aot) 

https://jira.devtools.intel.com/browse/NGTF-2255

### DIFF
--- a/ngraph_bridge/ngraph_encapsulate_impl.cc
+++ b/ngraph_bridge/ngraph_encapsulate_impl.cc
@@ -152,7 +152,7 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
     }
 
     int function_size;
-    if (!m_do_aot){
+    if (!m_do_aot) {
       function_size = ng_function->get_graph_size() / 1024;  // kb unit
     }
 
@@ -165,7 +165,7 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
       file_name += ("_" + to_string(rank_id));
 #endif
       file_name += ".json";
-      if (m_do_aot){
+      if (m_do_aot) {
         NgraphSerialize(file_name, serialized_ng_function);
       } else {
         NgraphSerialize(file_name, ng_function);
@@ -181,7 +181,7 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
       int input_tensors_bytes_free = 0;
       evicted_ng_exec = m_ng_exec_map[m_lru.back()];
       m_ng_exec_map.erase(m_lru.back());
-      if (m_do_aot){
+      if (m_do_aot) {
         m_serialized_ng_function_map.erase(evicted_ng_exec);
       } else {
         m_ng_function_map.erase(evicted_ng_exec);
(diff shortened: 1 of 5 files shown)


ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -222,21 +235,36 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
...
+      } else {
+        NgraphSerialize(file_name, ng_function);
+      }
       return errors::Internal("Caught exception while compiling op_backend: ",
                               exp.what(), "\n");
     } catch (...) {
       BackendManager::UnlockBackend(m_op_backend_name);
-      NgraphSerialize("tf_function_error_" + m_name + ".json", ng_function);
+      string file_name = "tf_function_error_" + m_name + ".json";
+      if (m_do_aot) {
```
sayantan-nervana: use DumpNgFunction

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -222,21 +235,36 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
       }
     } catch (const std::exception& exp) {
       BackendManager::UnlockBackend(m_op_backend_name);
-      NgraphSerialize("tf_function_error_" + m_name + ".json", ng_function);
+      string file_name = "tf_function_error_" + m_name + ".json";
```
sayantan-nervana: use DumpNgFunction

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -145,22 +148,28 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
...
 #if defined NGRAPH_DISTRIBUTED
       int rank_id;
       rank_id = ng::get_distributed_interface()->get_rank();
-      NgraphSerialize(
-          "tf_function_" + m_name + "_" + to_string(rank_id) + ".json",
-          ng_function);
+      file_name += ("_" + to_string(rank_id));
 #endif
+      file_name += ".json";
+      if (m_do_aot) {
```
sayantan-nervana: use DumpNgFunction

PR #229, opened 2019-09-06
shrestha/backend_error_message_fixes

shresthamalik:
- Added a new api CanCreateBackend.  
This api tries to create a backend with the full backend string (backend_type:device_id). Meant to be used to check for a valid backend before assigning it to the op. There is no user-facing api for this. It is meant for internal use alone
- IsSupportedBackend() is removed
- changed the parameters for fused_batchnorm_test
- created a env variable NGRAPH_TF_SEED to fix the seed for running tests, helps in debugging. Works only for C++ test. needs to extend to python tests.

### DIFF
--- a/ngraph_bridge/ngraph_backend_manager.cc
+++ b/ngraph_bridge/ngraph_backend_manager.cc
@@ -36,9 +36,9 @@ map<std::string, int> BackendManager::ref_count_each_backend_;
 
 Status BackendManager::SetBackendName(const string& backend_name) {
   std::lock_guard<std::mutex> lock(BackendManager::ng_backend_name_mutex_);
-  if (backend_name.empty() || !IsSupportedBackend(backend_name)) {
-    return errors::Internal("Backend ", backend_name,
-                            " is not supported on nGraph");
+  auto status = BackendManager::CanCreateBackend(backend_name);
+  if (status != Status::OK()) {
+    return errors::Internal("Failed to set backend: ", status.error_message());
   }
   BackendManager::ng_backend_name_ = backend_name;
   return Status::OK();
@@ -58,8 +58,8 @@ Status BackendManager::CreateBackend(const string& backend_name) {
     }
 
     if (bend_ptr == nullptr) {
-      return errors::Internal("Could not create backend of type ",
-                              backend_name);
+      return errors::Internal("Could not create backend of type ", backend_name,
+                              " got nullptr");
     }
     std::unique_ptr<Backend> bend = std::unique_ptr<Backend>(new Backend);
     bend->backend_ptr = std::move(bend_ptr);
@@ -134,14 +134,22 @@ size_t BackendManager::GetNumOfSupportedBackends() {
 }
 
 bool BackendManager::IsSupportedBackend(const string& backend_name) {
-  auto status = BackendManager::CreateBackend(backend_name);
-  if (status != Status::OK()) {
-    return false;
+  vector<string> registered_backends =
+      ng::runtime::BackendManager::get_registered_backends();
+  vector<string>::iterator itr = find(registered_backends.begin(),
+                                      registered_backends.end(), backend_name);
+
+  return itr != registered_backends.end();
+};
+
+Status BackendManager::CanCreateBackend(const string& backend_string) {
+  auto status = BackendManager::CreateBackend(backend_string);
+  if (status == Status::OK()) {
+    // The call to create backend increases the ref count
+    // so releasing the backend here
+    BackendManager::ReleaseBackend(backend_string);
   }
-  // The call to create backend increases the ref count
-  // so releasing the backend here
-  BackendManager::ReleaseBackend(backend_name);
-  return true;
+  return status;
 };
 
 Status BackendManager::GetCurrentlySetBackendName(string* backend_name) {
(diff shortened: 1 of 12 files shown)


shresthamalik:
This PR is created on r0.18 and will be cherry-picked to master and other release branches (if required). The variable-pipeline will not run on this PR as its not available on this branch.

We could cherry-pick it, add it to this PR. Let me know.

test/test_utilities.cpp
```diff
@@ -93,6 +93,27 @@ void SetNGraphTFBackend(const string& backend_name) {
...
+    NGRAPH_VLOG(5) << "Got seed " << seed;
+    return seed;
+  }
+
+  string seedstr = GetEnvVariable(env_name);
+  try {
+    seed = stoi(seedstr);
+  } catch (const std::exception& exp) {
+    throw std::runtime_error{"Cannot set " + env_name + " with value " +
+                             seedstr + " ,got exception " + exp.what()};
```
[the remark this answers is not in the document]
sayantan-nervana: So this allows you to set seeds externally through `NGRAPH_TF_SEED`. But in CI, will we be setting `NGRAPH_TF_SEED`? In which case CI will be using time as seed. If we expect the test to fail for certain seeds, we should dump the seed value on failure so that they can be repro-ed later

ngraph_bridge/ngraph_backend_manager.cc
```diff
@@ -134,14 +134,22 @@ size_t BackendManager::GetNumOfSupportedBackends() {
...
+  vector<string> registered_backends =
+      ng::runtime::BackendManager::get_registered_backends();
+  vector<string>::iterator itr = find(registered_backends.begin(),
+                                      registered_backends.end(), backend_name);
+
+  return itr != registered_backends.end();
+};
+
+Status BackendManager::CanCreateBackend(const string& backend_string) {
+  auto status = BackendManager::CreateBackend(backend_string);
```
kanvi-nervana: Same as above

test/test_utilities.h
```diff
@@ -92,7 +95,7 @@ template <typename T>
 void AssignInputValuesRandom(Tensor& A, T min, T max) {
   auto A_flat = A.flat<T>();
   auto A_flat_data = A_flat.data();
-  srand(static_cast<unsigned>(time(0)));
+  srand(GetSeedForRandomFunctions());
```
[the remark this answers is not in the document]
kanvi-nervana: Same

test/graph_rewrites/backend_manager_test.cc
```diff
@@ -109,6 +109,13 @@ TEST(BackendManager, GetCurrentlySetBackendName) {
   RestoreEnv(env_map);
 }
 
+// Test CanCreateBackend
+TEST(BackendManager, CanCreateBackend) {
+  ASSERT_OK(BackendManager::CanCreateBackend("CPU"));
```
sayantan-nervana: ASSERT_OK(BackendManager::CanCreateBackend("CPU:0"));

test/test_utilities.h
```diff
@@ -59,6 +59,9 @@ void PrintTensorAllValues(
     const Tensor& T1,
     int64 max_entries);  // print max_entries of elements in the Tensor
 
+// Generating Random Seed
```
sayantan-nervana: // Generating Seed for PRNG

ngraph_bridge/ngraph_backend_manager.h
```diff
@@ -57,9 +57,11 @@ class BackendManager {
   // Returns number of nGraph supported backends
   static size_t GetNumOfSupportedBackends();
 
-  // Returns True if the backend is supported or not
-  // Tries to create a backend with backend_name string to determine whether
-  // supported or not
+  // Returns True if the backend is supported
+  // Searches the registered backends with nGraph to determine this
```
kanvi-nervana:
```suggestion
  // Searches the list of backends registered with nGraph to determine this
```

test/test_utilities.cpp
```diff
@@ -93,6 +93,31 @@ void SetNGraphTFBackend(const string& backend_name) {
...
+  if (!IsEnvVariableSet(env_name)) {
+    NGRAPH_VLOG(5) << "Got seed " << seed;
+    return seed;
+  }
+
+  string seedstr = GetEnvVariable(env_name);
+  try {
+    int temp_seed = stoi(seedstr);
+    if (temp_seed < 0) {
+      throw std::runtime_error{"Cannot set negative seed"};
```
avijit-nervana: This should be invalid_arg exception.

ngraph_bridge/ngraph_backend_manager.h
```diff
@@ -57,9 +57,11 @@ class BackendManager {
...
 
-  // Returns True if the backend is supported or not
-  // Tries to create a backend with backend_name string to determine whether
-  // supported or not
+  // Returns True if the backend is supported
+  // Searches the list of backends registered with nGraph to determine this
+  // Meant to be used without device_id
+  // For eg. IsSupportedBackend("GPU") --> returns True
+  //         IsSupportedBackend("GPU:0") --> returns False
   static bool IsSupportedBackend(const string& backend_name);
```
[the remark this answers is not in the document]
shresthamalik: We already have an api for that , this returns the list of registered backends
https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_api.h#L39
which calls
https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_backend_manager.h#L55

We can retire the api `bool IsSupportedBackend(const string& backend_name)` if we don't see any use for this. Let me know.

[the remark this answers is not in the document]
avijit-nervana: Yes - that makes sense.

PR #238, opened 2019-09-13
sarkars/Set pipeline depth

sayantan-nervana:
1. Normal run: attach `pipeline_depth` to rewriter config
2. tf2ngraph run: pass `--pipeline_depth 2` to CLI
3. Add try-except around `create_input_tensor` and `create_output_tensor`

https://jira.devtools.intel.com/browse/NGTF-2270


adk9:
Thanks for the PR, but we're closing this because it's no longer needed; training support is being removed from the bridge.

PR #239, opened 2019-09-13
Sarkars/use in graph ngfunc instead of deserializing

sayantan-nervana:
1. Avoid `deserialize` then `serialize`. Directly use the ng function serialized string in case we have to dump it during errors

2. Also, we do not need a cache of exec->ng_function, as long as we have a cache of ng_exec->serialized_ng_func_string. Because the only reason we had `m_ng_function_map` was to serialize and dump in case of errors. Now that we have `m_serialized_ng_function_map`, `m_ng_function_map` is not needed for both aot and non-aot paths

3. Move TranslateGraph AFTER backend creation in AOT encapsulates path (since some TF nodes like CNMS/gather need backend specific ops)

4. Pulling in the minor fix of PR247

https://jira.devtools.intel.com/browse/NGTF-2276


ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -134,32 +134,37 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
...
       }
-      ng_function = ng::deserialize(itr->second);
+      serialized_ng_func = itr->second;
     }
 
-    auto function_size = ng_function->get_graph_size() / 1024;  // kb unit
+    int function_size;
+    if (!m_do_aot) {
+      function_size = ng_function->get_graph_size() / 1024;  // kb unit
+    }
```
avijit-nervana: If not not - then function size not calculated - is that ok?

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -134,32 +134,37 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
...
 
     // Serialize to nGraph if needed
     if (std::getenv("NGRAPH_ENABLE_SERIALIZE") != nullptr) {
       std::string file_name = "tf_function_" + m_name + ".json";
-      NgraphSerialize("tf_function_" + m_name + ".json", ng_function);
+      DumpStringToFile("tf_function_" + m_name + ".json", serialized_ng_func);
```
avijit-nervana: Rename this function from `DumpStringToFile` to `StringToFile`.

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -134,32 +134,37 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
     MemoryProfile(vm0, rss0);
 
     NGRAPH_VLOG(1) << "Compilation cache miss: " << m_name;
+    string serialized_ng_func;
     if (!m_do_aot) {
       TF_RETURN_IF_ERROR(Builder::TranslateGraph(input_shapes, static_input_map,
                                                  &m_graph, ng_function));
       ng_function->set_friendly_name(m_name);
+      serialized_ng_func = ngraph::serialize(ng_function, 4);
```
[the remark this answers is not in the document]
sayantan-nervana: 4 is the indentation size. By default it is 0. 4 is a reasonable indentation size that makes the dumped json look well-formatted and is used in other places in the bridge like `NgraphSerialize` (https://github.com/tensorflow/ngraph-bridge/blob/6286fdddd19033a840ea0795742d2edc4d8f7213/ngraph_bridge/ngraph_utils.cc#L315)

ngraph_bridge/ngraph_encapsulate_impl.cc
sayantan-nervana: Here is the description: https://github.com/NervanaSystems/ngraph/blob/0a6f5bca7de2013518fe5124534b6a6ca97970f7/src/ngraph/function.hpp#L109

Given that the function is not a reliable measure, will remove it as discussed f2f.

Currently we specify cache depth rather than memory consumption through the environment flag.

ngraph_bridge/ngraph_utils.cc
```diff
@@ -312,12 +312,15 @@ Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
 void NgraphSerialize(const std::string& file_name,
                      const std::shared_ptr<ngraph::Function>& ng_function) {
   NGRAPH_VLOG(0) << "Serializing graph to: " << file_name << std::endl;
-  std::string js = ngraph::serialize(ng_function, 4);
+  StringToFile(file_name, ngraph::serialize(ng_function, 4));
+}
+
+void StringToFile(const std::string& file_name, const std::string& contents) {
```
shresthamalik: Have we check if the dumped .json is able to load on nGraph?

sayantan-nervana: Checked the following:
```
NGRAPH_ENABLE_SERIALIZE=1 python mnist_deep_simplified.py
../../../nbench -f tf_function_ngraph_cluster_23.json -b CPU
```
which works

ngraph_bridge/ngraph_encapsulate_impl.cc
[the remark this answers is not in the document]
avijit-nervana: Add a comment stating this. Ideally you should define a constant with the same name

ngraph_bridge/ngraph_utils.cc
avijit-nervana: Add a unit test to do the same if one doesn't exist.

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -134,32 +134,32 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
...
             signature);
       }
-      ng_function = ng::deserialize(itr->second);
+      serialized_ng_func = itr->second;
     }
 
-    auto function_size = ng_function->get_graph_size() / 1024;  // kb unit
-
     // Serialize to nGraph if needed
     if (std::getenv("NGRAPH_ENABLE_SERIALIZE") != nullptr) {
```
shresthamalik: While we are at it should we change it to NGRAPH_TF_ENABLE_SERIALIZE?

ngraph_bridge/ngraph_encapsulate_clusters.cc
```diff
@@ -793,6 +789,14 @@ Status EncapsulateClusters(
           }
           TF_RETURN_IF_ERROR(BackendManager::CreateBackend(
               op_backend_name));  // Created a backend here. must free it
+          // TranslateGraph must be called AFTER CreateBackend because some TF
+          // ops like CNMS and gather use backend specific nodes
+          TF_RETURN_IF_ERROR(Builder::TranslateGraph(
+              input_shapes, static_input_map, &graph_for_current_encapsulate,
+              ng_function));
+          int json_indentation = 4;
```
shresthamalik: Curious, why are we translating here? 
Is is to attach the ng_function in tf2ngraph? The ngfunction is different for different shapes in case the graph has static inputs, right?

ngraph_bridge/ngraph_encapsulate_impl.cc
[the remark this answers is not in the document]
sayantan-nervana: Added a pytest

ngraph_bridge/ngraph_encapsulate_clusters.cc
sayantan-nervana: AOT is only supported for certain kinds of graphs. static inp is not suported. https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_encapsulate_clusters.cc#L712

PR #247, opened 2019-09-17
sarkars/Fix setconfig

sayantan-nervana:
Right now only `_ngraph_aot_requested` is being passed through, other configs are not :(

https://jira.devtools.intel.com/browse/NGTF-2276

### DIFF
--- a/ngraph_bridge/ngraph_encapsulate_clusters.cc
+++ b/ngraph_bridge/ngraph_encapsulate_clusters.cc
@@ -804,7 +804,7 @@ Status EncapsulateClusters(
             // For e.g. _ngraph_ice_cores --> ice_cores
             if (itr.first.find("_ngraph_") != std::string::npos) {
               // leave out _ngraph_aot_requested
-              if (itr.first.find("_ngraph_aot_requested") !=
+              if (itr.first.find("_ngraph_aot_requested") ==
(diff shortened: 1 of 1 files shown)


ngraph_bridge/ngraph_encapsulate_clusters.cc
```diff
@@ -804,7 +804,7 @@ Status EncapsulateClusters(
             // For e.g. _ngraph_ice_cores --> ice_cores
             if (itr.first.find("_ngraph_") != std::string::npos) {
               // leave out _ngraph_aot_requested
-              if (itr.first.find("_ngraph_aot_requested") !=
+              if (itr.first.find("_ngraph_aot_requested") ==
                   std::string::npos) {
```
sayantan-nervana: regd unit tests.

this is in the aot path, where we create a temporary backend and use it to compile the ng graph. We have unit tests for the `EncapsulateClusters` function which checks if aot is requested then ngfunction and ngexecutable are attached to the graph.

However with this bug the temporary backend we were creating for doing the AOT compilation was not getting the right configurations. In other words, the effects of NOT setting config in the temporary backend are not visible/differentiable in the outputs of `EncapsulateClusters` function. Hence this does not look testable

PR #253, opened 2019-09-19
Sarkars/fix dump file name  and NGraphSerialize/StringToFile now return statuses

sayantan-nervana:
1. replace / with -- for filenames
2. NgraphSerialize and StringToFile now return Status so we can bubble them up

duplicate of https://github.com/tensorflow/ngraph-bridge/pull/220, but upgraded to master now

JIRA: https://jira.devtools.intel.com/browse/NGTF-2227


ngraph_bridge/ngraph_encapsulate_clusters.cc
```diff
@@ -830,9 +830,12 @@ Status EncapsulateClusters(
             ng_exec = op_backend->compile(ng_function);
           } catch (...) {
             BackendManager::UnlockBackend(op_backend_name);
-            NgraphSerialize("tf_function_error_aot.json", ng_function);
+            Status st =
+                NgraphSerialize("tf_function_error_aot.json", ng_function);
```
shresthamalik: do we need to check the st here?

ngraph_bridge/ngraph_utils.cc
```diff
@@ -309,25 +309,55 @@ Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
...
+  // Sanitizing file name to take care of '/' that might be present in TF node
+  // names
+  string new_file_name;
+  // The valid TF node names seem to be: [A-Za-z0-9.][A-Za-z0-9_.\\-/]*
+  // . is another non-alphanumeric char, but once / are replaced by --, . is
+  // fine in a file name.
+  for (const auto& itr : file_name) {
+    new_file_name += ((itr == '/') ? string("--") : string({itr}));
+  }
+  return new_file_name;
```
shresthamalik: what is the relation between the file name and TF node name. Failing to make the connection, how does one impact other.

shresthamalik: Where are we dumping using the TF node name?

ngraph_bridge/ngraph_utils.cc
```diff
@@ -309,25 +309,55 @@ Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
...
+        "Passed a null pointer as ng function to serialize");
+  }
+  try {
+    ngraph::serialize(ng_function, json_indentation);
+  } catch (...) {
+    return errors::Internal("Failed to serialize ngraph function");
+  }
+  string new_file_name = SanitizeFileName(file_name);
+  NGRAPH_VLOG(0) << "Serializing graph to: " << new_file_name << std::endl;
+  return StringToFile(new_file_name, serialized, false);
```
shresthamalik: Let StringToFile handle the sanitizing? Pass
StringToFile(new_file_name, serialized, true);

ngraph_bridge/ngraph_utils.cc
sayantan-nervana: Consider this line here: https://github.com/tensorflow/ngraph-bridge/blob/e247c6e81cf0ef3fc5bd774a916364fdafa9405c/ngraph_bridge/ngraph_encapsulate_impl.cc#L157

`StringToFile("tf_function_" + m_name + ".json", serialized_ng_func);`

It is set here: https://github.com/tensorflow/ngraph-bridge/blob/e247c6e81cf0ef3fc5bd774a916364fdafa9405c/ngraph_bridge/ngraph_encapsulate_op.cc#L68


when importing from pb, the nodes could get an extra `import/` to their names. So we could be attempting to dump: `tf_function_import/ngraph_cluster_0.json`, which will fil now on master, because it is going to be interpreted as folder = `tf_function_import` and file = `ngraph_cluster_0.json` (and teh folder does not exist)

shresthamalik: I see. Thanks! 👍

PR #254, opened 2019-09-19
Sarkars/modify update config to check if ngoptimizer is enabled

sayantan-nervana:
1. Fix learning rate for SGD so that the mnist_deep_simplified.py produces reasonable results (~95%) as opposed to ~20% it is doing now

2. Fix i'th pytest running ngoptimizer i times:
1.a update_config checks if ngraph optimizer is enabled, if so then simply return (instead of appending another request for ngraph optimizer)
1.b fixed common.py by removing mutable default arguments to functions

3. Add more pytests for update_config.


test/python/common.py
```diff
@@ -47,7 +47,10 @@ def import_pbtxt(self, pb_filename):
             tf.import_graph_def(graph_def)
         return graph
 
-    def with_ngraph(self, l, config=tf.ConfigProto()):
+    def with_ngraph(self, l, config=None):
+        # Passing config as None and then initializing it inside because mutable objects should not be used as defaults in python
```
avijit-nervana: Comment too long. Plit this at 80 column..

test/python/test_updateconfig.py
```diff
@@ -46,3 +46,23 @@ def test_update_config(self):
...
+        # Helper function to count the number of occurances in a config
+        def count_ng_optimizers(config):
+            custom_opts = config.graph_options.rewrite_options.custom_optimizers
+            count = 0
+            for i in range(len(custom_opts)):
+                if custom_opts[i].name == 'ngraph-optimizer':
+                    count += 1
+            return count
+
+        config = tf.ConfigProto(allow_soft_placement=True)
```
avijit-nervana: We should remove the `allow_soft_placement` as it's not needed.

test/python/common.py
sayantan-nervana: Split into 2 lines

python/ngraph_bridge/__init__.in.py
```diff
@@ -201,12 +201,20 @@ def is_grappler_enabled():
 
 def update_config(config, backend_name = "CPU", device_id = ""):
     #updating session config if grappler is enabled
+    opt_name = 'ngraph-optimizer'
```
shresthamalik: should the opt_name be defined inside the if

test/python/test_updateconfig.py
sayantan-nervana: This reflects a real world config, where the user's config might have extra information in it. We have pytests for update_config that check that "allow_soft_placement" remains in place after passing through update_config.

In this particular case we are not rechecking if `allow_soft_placement` is present in the output. However `allow_soft_placement` is there as a representation of a real world non empty config

python/ngraph_bridge/__init__.in.py
sayantan-nervana: moved inside

test/python/test_updateconfig.py
avijit-nervana: Ok - that makes sense. Then please add a comment in the code to reflect that. Remember - git hub review comments go away - but the comments in the code aren't.

sayantan-nervana: Added comments

PR #255, opened 2019-09-20
Sarkars/add provenance tags to all nodes

sayantan-nervana:
- add provenance tags to some missing nodes
- add a check that all nodes (except for result and parameter nodes) in the ng_function created in TranslateGraph have provenance tags
-  upgraded nGraph to v0.25.1-rc.7


ngraph_bridge/ngraph_builder.cc
```diff
@@ -5056,6 +5136,27 @@ Status Builder::TranslateGraph(
     result->set_needs_default_layout(true);
   }
 
+  // TODO: Use is_type<ng::op::Result>(n) when we upgrade to new ngraph
+  auto check_if_result_type = [&ng_function](shared_ptr<ng::Node> n) {
```
shresthamalik: Are we saying ng::op::Result dont need a provenance tag?

ngraph_bridge/ngraph_builder.cc
```diff
@@ -5056,6 +5136,27 @@ Status Builder::TranslateGraph(
     result->set_needs_default_layout(true);
   }
 
+  // TODO: Use is_type<ng::op::Result>(n) when we upgrade to new ngraph
```
shresthamalik: Reviewed this part. We are safeguarded by the check, if some node is not tagged we will know.

shresthamalik: Is nGraph is only attaching the provenance tags when certain env flags are enabled then we might run

ngraph_bridge/ngraph_builder.cc
sayantan-nervana: Yes.. Result and Parameter don't have provenance tags.

ngraph_bridge/ngraph_builder.cc
sayantan-nervana: Its always going to attach tags in builders. If flags are present it will start to do something with the tags.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -5056,6 +5146,20 @@ Status Builder::TranslateGraph(
     result->set_needs_default_layout(true);
   }
 
+  auto check_if_terminal_type = [](shared_ptr<ng::Node> n) {
```
shresthamalik: Please add comment on what this check is doing. Rename the function if required

sayantan-nervana:
Regd tests:

`TranslateGraph` now ends with this a check that all nodes have a tag

PR #256, opened 2019-09-21
Shrestha/detect multiple variable updates from the same output of encap

shresthamalik:
This PR is mainly to detect and throw an error in case a scenario like below is encountered.
https://github.com/tensorflow/ngraph-bridge/pull/256/files#diff-02acaea7e11deb779e44f53cdfd1ca84R467

Two Variables are being updated with the same value --> by same output (output index) of Enacap Op. 

Changes
- Catalog throws an error if an attempt is made to enter the same key again
- Catalog api s are more robust (const added wherever required, replaced the use of [] operator by insert, at)


ngraph_bridge/enable_variable_ops/ngraph_catalog.cc
```diff
@@ -68,89 +92,114 @@ bool NGraphCatalog::EncapOutputIndexNeedsCopy(int graphid, string node_name,
...
+const bool NGraphCatalog::ExistsInEncapOutputInfoMap(const string& key) {
   auto itr = NGraphCatalog::encap_output_info_map_.find(key);
   return itr != NGraphCatalog::encap_output_info_map_.end();
 }
 
-bool NGraphCatalog::ExistsInEncapOutputInfoMap(int graphid, string node_name,
-                                               int input_index) {
-  std::string key =
-      NGraphCatalog::CreateNodeKey(graphid, node_name, input_index);
+const bool NGraphCatalog::ExistsInEncapOutputInfoMap(const int& graphid,
```
sayantan-nervana: `const bool`.. perhaps you meant `const bool&`?
Also for bools and ints, just returning `bool` vs returning `const bool&` wont make much of a difference.

Similarly the return type of `CreateNodeKey` is `const string`. maybe you meant `const string&`?

shresthamalik: It is const string/bool and not & because the value being returned is created inside the function. returning const coz dont want the user of the api to change its value.

shresthamalik: Will change it to return string/bool as the user might still read them into non-const types and modify it

ngraph_bridge/enable_variable_ops/ngraph_catalog.cc
```diff
@@ -35,29 +35,53 @@ unordered_map<string, tuple<string, bool, bool>>
     NGraphCatalog::encap_output_info_map_;
 
 // Function to create the Node Key
-string NGraphCatalog::CreateNodeKey(int graph_id, string node_name, int index) {
+const string NGraphCatalog::CreateNodeKey(const int& graph_id,
```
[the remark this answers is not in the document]
shresthamalik: For static member functions, const does not apply
https://stackoverflow.com/questions/7035356/c-why-static-member-function-cant-be-created-with-const-qualifier

ngraph_bridge/enable_variable_ops/ngraph_catalog.cc
```diff
@@ -68,89 +92,114 @@ bool NGraphCatalog::EncapOutputIndexNeedsCopy(int graphid, string node_name,
...
 
 // Functions relating Input Variable Shared Name Map
-string NGraphCatalog::GetInputVariableSharedName(int graphid, string node_name,
-                                                 int input_index) {
-  std::string node_key =
-      NGraphCatalog::CreateNodeKey(graphid, node_name, input_index);
-  return NGraphCatalog::input_variable_sharedname_map_[node_key];
+void NGraphCatalog::AddToInputVariableSharedNameMap(const string& key,
+                                                    const string& val) {
+  if (NGraphCatalog::ExistsInInputVariableSharedNameMap(key)) {
```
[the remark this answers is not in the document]
shresthamalik: I think the current implementation is simpler than trying to insert, get a pair out, and check whether the 2nd element of that pair is true or false and then throw an error.
I prefer reusing the apis (ExistsInInputVariableSharedNameMap) that we already have for the maps. 

Let me know if there is any other advantage?

ngraph_bridge/enable_variable_ops/ngraph_catalog.cc
```diff
@@ -68,89 +92,114 @@ bool NGraphCatalog::EncapOutputIndexNeedsCopy(int graphid, string node_name,
...
+  }
+
   // create a tuple
   tuple<string, bool, bool> val =
       make_tuple(shared_name, copy_to_tf, is_tf_just_looking);
-  NGraphCatalog::encap_output_info_map_[key] = val;
+  NGraphCatalog::encap_output_info_map_.insert({key, val});
 }
 
-bool NGraphCatalog::ExistsInEncapOutputInfoMap(string key) {
```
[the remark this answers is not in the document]
shresthamalik: It is used at other places

ngraph_bridge/enable_variable_ops/ngraph_catalog.cc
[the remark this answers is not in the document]
vishakha-nervana: Okay, only thing is we can avoid extra function call, but again, I don't think it would make much difference, so its all good as it is.

test/graph_rewrites/remove_ngraphassigns.cc
```diff
@@ -463,6 +463,73 @@ TEST(RemoveNGraphAssigns, Graph5) {
   ASSERT_NOT_OK(RemoveNGraphAssigns(&graph));
 }
 
+// Graph with 2 Variables
+// This graph will throw an error
+//
+// Two NGraphVariables are being assigned the same value.
```
shresthamalik: The const could be a subgraph, mention in the comment

test/graph_rewrites/remove_ngraphassigns.cc
```diff
@@ -463,6 +463,73 @@ TEST(RemoveNGraphAssigns, Graph5) {
...
+// The ConstOp gets encapsulated and both the variables are being
+// assigned from the same output index of EncapsulateOp
+// As the update is happening in-place
+// inside the encapsulate op, only one of the variable gets updated, leading
+// to functional incorrectness.
```
shresthamalik: Mention : more details on what goes on inside the encap op

PR #233, opened 2019-09-11
Sarkars/some cleanup

sayantan-nervana:
1. Encapsulate's member variables end with `_` to follow TF convention
2. `CachePipelinedTensorIfNeeded` renamed to `UpdatePipelinedTensorCache`
3. Single `ClearExecMaps` function instead of 5 different clear functions to clear the ngexec maps
4. Certain functions of impl are only used in varopts path, so moved them under an `#ifdef NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS`
5. `SetCopyLog` incorrectly resets the stringstream. Correcting it to append, and renaming it to `AppendCopyLog`
6. Added `GetPipelineIdxAndTensors` that subsumes 3 function calls. Moved subsumed functions to `private`, since they are no longer being used by encapsulate op directly

TODO: check speeds of resnet etc on CPU/GPU on normal/varopts builds

https://jira.devtools.intel.com/browse/NGTF-2263

Abandoning in favor of: https://github.com/tensorflow/ngraph-bridge/pull/257

PR #257, opened 2019-09-24
Sarkars/some cleanup  1

sayantan-nervana:
1. Encapsulate's member variables end with _ to follow TF convention
2. `CachePipelinedTensorIfNeeded` renamed to `UpdatePipelinedTensorCache`
3. Single `ClearExecMaps` function instead of 5 different clear functions to clear the ngexec maps
4. Certain functions of impl are only used in varopts path, so moved them under an `#ifdef NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS`
5. `SetCopyLog` incorrectly resets the stringstream. Correcting it to append, and renaming it to `AppendCopyLog`
6. Added `GetPipelineIdxAndTensors` that subsumes 3 function calls. Moved subsumed functions to private, since they are no longer being used by encapsulate op directly
7. removed the if-def around the Get and Set graph id, since it is useful debugging info fro all types of builds (not just relevant to varopts build)

Merge conflict resolved version of https://github.com/tensorflow/ngraph-bridge/pull/233


ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -535,11 +535,11 @@ Status NGraphEncapsulateImpl::ParseNodeAttributes(
   return Status::OK();
 }
 
-Status NGraphEncapsulateImpl::CachePipelinedTensorIfNeeded(
+Status NGraphEncapsulateImpl::UpdatePipelinedTensorCache(
```
shresthamalik: Rename to InitializePipelinedTensorCache ?

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -588,6 +588,55 @@ void NGraphEncapsulateImpl::DumpNgFunction(
...
+  m_ng_exec_output_cache_map.clear();
+  m_ng_exec_map.clear();
+  m_serialized_ng_function_map.clear();
+  m_executable_pipelined_tensors_map.clear();
+}
+
+Status NGraphEncapsulateImpl::GetPipelineIdxAndTensors(
+    const std::shared_ptr<ngraph::runtime::Executable>& ng_exec,
+    std::tuple<int, PipelinedTensorVector, PipelinedTensorVector>& tpl) {
+  if (GetExecCanCreateTensor()) {
```
shresthamalik: Do we need this if?

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -588,6 +588,55 @@ void NGraphEncapsulateImpl::DumpNgFunction(
...
+          "Expected GetTensorsFromPipeline to return an index >= 0, but got ",
+          pipeline_idx);
+    }
+  }
+  return Status::OK();
+}
+
+Status NGraphEncapsulateImpl::ReturnPipelinedTensors(
+    std::shared_ptr<ngraph::runtime::Executable> ng_exec, size_t idx) {
+  if (GetExecCanCreateTensor()) {
```
shresthamalik: Same comment

ngraph_bridge/ngraph_encapsulate_impl.h
```diff
@@ -85,25 +85,43 @@ class NGraphEncapsulateImpl {
...
 
-  const int& GetNgraphCluster() { return m_ngraph_cluster; }
-
-  void SetNgraphCluster(const int& cluster) { m_ngraph_cluster = cluster; }
-
   int GetGraphId() { return m_graph_id; }
 
   void SetGraphId(const int& graph_id) { m_graph_id = graph_id; }
 
+  void AppendCopyLog(const string str) { copy_log_str << str; }
```
shresthamalik: The graph ids are attached irrespective of the build. Need to be out of the ifdef

ngraph_bridge/ngraph_encapsulate_impl.cc
sayantan-nervana: Yes.

If we call `UpdatePipelinedTensorCache` or `GetTensorsFromPipeline` without the `if`, they will throw errors.

ngraph_bridge/ngraph_encapsulate_impl.h
sayantan-nervana: True, but `GetGraphId` is used only in `NGRAPH_TF_ENABLE_VARIABLES_AND_OPTIMIZERS` build

ngraph_bridge/ngraph_encapsulate_impl.cc
sayantan-nervana: `UpdatePipelinedTensorCache` was the name decided in one of the meetings, so used that.

ngraph_bridge/ngraph_encapsulate_impl.h
shresthamalik: Please remove the if-def from the encap_op.cc too. The graph-ids help in debugging/logging and can be used in other builds.  It is also more consistent: we attach it, we read it and keep it handy incase we need it.
We did not have an in-def originally, must have been added somewhere in the refactoring
https://github.com/tensorflow/ngraph-bridge/blob/r0.18/ngraph_bridge/ngraph_encapsulate_op.cc#L110

ngraph_bridge/ngraph_encapsulate_impl.cc
shresthamalik: The calling function have the same check
https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_encapsulate_op.cc#L328.

Maybe its better that way so the user knows what it is calling

ngraph_bridge/ngraph_encapsulate_impl.h
sayantan-nervana: moved graphis functions outside ifdef, and removed ifdef from calling code

PR #261, opened 2019-09-25
Added IsFiniteOp

maxbubula:


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -473,6 +474,7 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
       type_constraint_map["HorovodBroadcast"]["T"] = NGraphNumericDTypes();
 #endif
       type_constraint_map["Identity"]["T"] = NGraphDTypes();
+      type_constraint_map["IsFinite"]["T"] = NGraphDTypes();
```
sayantan-nervana: https://github.com/tensorflow/tensorflow/blob/5dfacce7a7c73d1745d5c7af546d7846bdde0232/tensorflow/core/ops/math_ops.cc#L337

According to the line above, `IsFinite` acts on real types only, hence please use `NGraphRealDTypes`

https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_utils.cc#L291

maxbubula: You are right, fixed.

PR #268, opened 2019-09-27
Sarkars/add provenance tags to all nodes (#255)

sayantan-nervana:
cherry pick from r19 to master

will fail till #267 merges and upgrades ngcore

run the run all models ...  looks good


ngraph_bridge/ngraph_builder.cc
```diff
@@ -5090,6 +5165,54 @@ Status Builder::TranslateGraph(
...
+    // Pointer will cast to nullptr if this node is not a Result
+    auto ng_node = dynamic_pointer_cast<ng::op::Result>(n);
+    bool is_result = (ng_node != nullptr);
+    return n->is_parameter() || is_result;
+  };
+
+  size_t num_tags = 0;
+  for (auto n : ng_function->get_ordered_ops()) {
+    // Results and Parameters are not expected to have provenance tags
+    if (!check_if_result_or_parameter(n)) {
```
kanvi-nervana: This check will need to be modified when the PR for adding provenance tags to parameters is merged.

PR #269, opened 2019-09-27
shrestha/Added test cases for var as static input

shresthamalik:
- Adds a test case that simulates the scenario where the Variable is providing static input to Encapsulate Op. If this Variable is updated by nGraph the TF tensor is outdated.
However while translating graph with static inputs we use the input TF tensor's value leading to functional incorrectness.
test_variableops_static_input.py

Also cherrypicks parts of this commit from master: 33abaab7515832e82557f71162c82c43f5735b38


test/python/test_variableops_static_input.py
```diff
@@ -0,0 +1,153 @@
...
+        # property we can use this env flag NGRAPH_TF_NGVARIABLE_BUFFER_SHARING
+
+        # set env variable to disable NGraphVariable's buffer sharing
+        buffer_sharing_env = "NGRAPH_TF_NGVARIABLE_BUFFER_SHARING"
+        is_buffer_sharing_env_set = self.is_env_variable_set(buffer_sharing_env)
+        if is_buffer_sharing_env_set:
+            buffer_sharing_env_val = self.get_env_variable(buffer_sharing_env)
+            self.unset_env_variable(buffer_sharing_env)
+
+        self.set_env_variable(buffer_sharing_env, "0")
```
sayantan-nervana: We can also repeat the test for `NGRAPH_TF_NGVARIABLE_BUFFER_SHARING` set to 1 and not set. Or maybe another test where `NGRAPH_TF_NGVARIABLE_BUFFER_SHARING` is tested somehow?

shresthamalik: NGRAPH_TF_NGVARIABLE_BUFFER_SHARING setting to 1 or not setting at all, will show the same behavior on CPU. 
I can unset it and compare the values

shresthamalik: Added a test with NGRAPH_TF_NGVARIABLE_BUFFER_SHARING=1 set

test/python/test_variableops_static_input.py
```diff
@@ -0,0 +1,182 @@
...
+import ngraph_bridge
+import os
+
+# If the below graph is run for many iterations
+# The NGraphVar's NGTensor is updated every iteration
+# NGraphVar's TFTensor is not updated as no TF Node needs it
+# However StaticInputs are derived from the input TF Tensor (which is stale)
+# giving functionally incorrect results
+# TF MeanOp expects a static input
+#
```
shresthamalik: Add * for static input

PR #272, opened 2019-10-01
Sarkars/debug 2 provenance tags for single node

sayantan-nervana:
1.
Add check so that each ng node (other than parameter an dresult) have exactly 1 provenance tag

2.
Make `ngraph_conversion.cpp` functions (like `BatchToNGraph`) accept a prov tag and only tag if a new node is created

3.
Adds `PerformNgBroadcast` to call `ng::builder::numpy_broadcast` and add provenance tags appropriately. All other builders used will return a new ng node, so this special care does not need to be taken

4.
Without these changes, certain graphs (as shown in the pytest) will have nodes that have 2 provenance tags
Tests: 1 pytest that has a broadcast that does not create new ng nodes, and another with a broadcast that does create new ng nodes

5.
Also updates r0.19's version and ngcore version

6.
Also add full path to header in ngraph_conversions.cc
#include "ngraph_conversions.h" -->
#include "ngraph_bridge/ngraph_conversions.h"

7.
Add `SetTracingInfo` to do provenance tagging, friendly name setting and printing

### DIFF
--- a/ngraph_bridge/ngraph_builder.cc
+++ b/ngraph_bridge/ngraph_builder.cc
@@ -346,6 +346,17 @@ Builder::TF_NGRAPH_CONST_MAP() {
   return the_map;
 }
 
+static std::pair<std::shared_ptr<ng::Node>, std::shared_ptr<ng::Node>>
+PerformNgBroadcast(const string& prov_tag, std::shared_ptr<ng::Node> ng_lhs,
+                   std::shared_ptr<ng::Node> ng_rhs) {
+  std::shared_ptr<ng::Node> ng_lhs_new, ng_rhs_new;
+  std::tie(ng_lhs_new, ng_rhs_new) =
+      ng::builder::numpy_broadcast(std::make_pair(ng_lhs, ng_rhs));
+  if (ng_lhs_new != ng_lhs) ng_lhs->add_provenance_tag(prov_tag);
+  if (ng_rhs_new != ng_rhs) ng_rhs->add_provenance_tag(prov_tag);
+  return make_pair(ng_lhs_new, ng_rhs_new);
+}
+
 // Helper function to translate a unary op.
 //
 // Parameters:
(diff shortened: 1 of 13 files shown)


ngraph_bridge/ngraph_builder.cc
```diff
@@ -92,15 +92,28 @@ static void SaveNgOp(Builder::OpMap& ng_op_map, const std::string& op_name,
   ng_op_map[op_name].push_back(output_node);
 }
 
-template <class TOpType, class... TArg>
-std::shared_ptr<TOpType> ConstructNgNode(const std::string& op_name,
-                                         TArg&&... Args) {
-  auto ng_node = std::make_shared<TOpType>(std::forward<TArg>(Args)...);
+void SetTracingInfo(const std::string& op_name,
+                    const shared_ptr<ng::Node> ng_node) {
+  // This function is used to trace whng node came from which tf node
```
kanvi-nervana: Typo --- `whng` -->did you mean `which` ?

ngraph_bridge/ngraph_builder.cc
```diff
@@ -92,15 +92,28 @@ static void SaveNgOp(Builder::OpMap& ng_op_map, const std::string& op_name,
   ng_op_map[op_name].push_back(output_node);
 }
 
-template <class TOpType, class... TArg>
-std::shared_ptr<TOpType> ConstructNgNode(const std::string& op_name,
-                                         TArg&&... Args) {
-  auto ng_node = std::make_shared<TOpType>(std::forward<TArg>(Args)...);
+void SetTracingInfo(const std::string& op_name,
```
kanvi-nervana: You may put this in utils so both builder and conversions can use it.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -5153,12 +5141,39 @@ Status Builder::TranslateGraph(
...
     // Results and Parameters are not expected to have provenance tags
     if (!check_if_result_or_parameter(n)) {
-      if (n->get_provenance_tags().size() == 0) {
-        return errors::Internal("Found ngraph node ", n->get_name(),
-                                " which does not have provenance tag set");
+      num_tags = n->get_provenance_tags().size();
+      if (num_tags != 1) {
+        char* original_provenance_flag_value =
+            getenv("NGRAPH_PROVENANCE_ENABLE");
+        // No need to free original_provenance_flag_value sccording to the
```
kanvi-nervana: Typo --- `sccording` --> `according`

ngraph_bridge/ngraph_builder.cc
```diff
@@ -5153,12 +5141,39 @@ Status Builder::TranslateGraph(
...
   };
 
+  size_t num_tags = 0;
   for (auto n : ng_function->get_ordered_ops()) {
     // Results and Parameters are not expected to have provenance tags
     if (!check_if_result_or_parameter(n)) {
-      if (n->get_provenance_tags().size() == 0) {
-        return errors::Internal("Found ngraph node ", n->get_name(),
-                                " which does not have provenance tag set");
+      num_tags = n->get_provenance_tags().size();
```
kanvi-nervana: We'll have change this check once prov tags are added for Parameters as well.

ngraph_bridge/ngraph_builder.cc
sayantan-nervana: Put it in `ngraph_builder.h`

ngraph_bridge/ngraph_builder.cc
sayantan-nervana: yes

PR #273, opened 2019-10-02
Kanvi/Add provenance tag for input/parameter nodes

kanvi-nervana:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -5086,7 +5086,10 @@ Status Builder::TranslateGraph(
     ng::Shape ng_shape;
     TF_RETURN_IF_ERROR(TFTensorShapeToNGraphShape(inputs[index], &ng_shape));
 
-    auto ng_param = make_shared<ng::op::Parameter>(ng_et, ng_shape);
+    string prov_tag;
+    GetNodeAttr(parm->attrs(), "_prov_tag", &prov_tag);
```
sayantan-nervana: Unrelated to this PR, but in general we need to compile a list of compulsory, optional (bridge or backend specific) flags that we attach to the encapsulate op.

PR #275, opened 2019-10-03
gopoka/[BERT] Implementation of GatherV2 for BERT model.

gopoka:
This implementation allows usage of GatherV2 on other backends than NNPI, with the constraint that axis == 0.


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -661,9 +661,9 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
                                                              bool* result) {
     // TODO: replace current_backend ->
     // BackendManager::GetCurrentlySetBackendName()
-    auto config_map =
-        BackendManager::GetBackendAttributeValues(current_backend);
-    *result = (config_map.at("ngraph_backend") == "NNPI");
+    // FIXME(gopoka): GatherV2 is supported only when axis == 0.
+    //                If possible, check should be added here.
+    *result = true;
```
sayantan-nervana: please add `GatherV2` in `type_constraint_map`. 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt#L16493

Tparams could be NGraphDTypes
Tindices and Taxis could be NGraphIndexDTypes

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -661,9 +661,9 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
                                                              bool* result) {
     // TODO: replace current_backend ->
     // BackendManager::GetCurrentlySetBackendName()
-    auto config_map =
-        BackendManager::GetBackendAttributeValues(current_backend);
-    *result = (config_map.at("ngraph_backend") == "NNPI");
+    // FIXME(gopoka): GatherV2 is supported only when axis == 0.
+    //                If possible, check should be added here.
```
sayantan-nervana: No, a check cant be added here. Here we can only check for "attributes" of the node. But the tf_axis check is an input to the node, which will be known only at runtime.

ngraph_bridge/ngraph_mark_for_clustering.cc
gopoka: `GatherV2` is already in `type_constraint_map`. This PR adds one special case when `axis == 0` (because this case can be handled by `GatherND` op in ngraph).

ngraph_bridge/ngraph_mark_for_clustering.cc
gopoka: Ok, I assumed that because `axis` is a static input it could be possibly checked here.

Is there any other way to remove node from being clustered when this kind of constraint is not satisfied?

ngraph_bridge/ngraph_mark_for_clustering.cc
sayantan-nervana: Ah, right!

ngraph_bridge/ngraph_mark_for_clustering.cc
sayantan-nervana: We can only check `attributes` here. The graph is not running yet, so we do not have access to values feeding into that node yet. So its fine to accept the node here, and later fail in `TranslateGraph`

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2237,35 +2237,76 @@ static Status TranslateGatherV2Op(
   // the full backend creation string
   auto config_map = BackendManager::GetBackendAttributeValues(backend_name);
   if (config_map.at("ngraph_backend") != "NNPI") {
-    return errors::Internal("In translating GatherV2 op ", op->name(),
-                            " found requested backend ", backend_name,
-                            " which is unsupported");
-  }
+    // This workaround works only if axis == 0
```
shresthamalik: Can we not support Gather for all valid values of axis using nGraph ops Gather or GatherND?

gopoka: I wasn't aware of `Gather` op in nGraph core (I couldn't find it previously), but I rewrote code to use it and all seems to be working fine.

gopoka: I have a small problem in python tests from tensorflow (I think). There is a case (file `build_cmake/artifacts/tensorflow/python/kernel_tests/gather_op_test.py`, test `testBadIndicesCPU`) that expects to catch an error from an op in a form of a string that contains `"indices\[0,0\] = 7 is not in \[0, 2\)"` message. However the nGraph implementation throws a different error and the whole test case fails. How should I handle such test cases?

ngraph_bridge/ngraph_mark_for_clustering.cc
shresthamalik: @gopoka  this comment is obsolete, is it?. please remove
```
// FIXME(gopoka):  GatherV2 is supported only when axis == 0.
    //                If possible, check should be added here.
```

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2282,14 +2282,6 @@ static Status TranslateGatherV2Op(
 
   // split and check the first part only, since the node attribute contains
```
kanvi-nervana: this comment can be moved

ngraph_bridge/ngraph_mark_for_clustering.cc
gopoka: Removed.

PR #276, opened 2019-10-03
mbubula/Add BatchMatMulV2 for BERT

maxbubula:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -913,9 +913,37 @@ static Status TranslateBatchMatMulOp(
...
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_lhs, ng_rhs;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_lhs, &ng_rhs));
+
+  auto ng_lhs_shape = ng_lhs->get_shape();
+  auto ng_rhs_shape = ng_rhs->get_shape();
+  size_t n_dims = ng_lhs_shape.size();
+
+  for (size_t i = 0; i < n_dims - 2; ++i) {
```
sayantan-nervana: Is this check also applicable for `TranslateBatchMatMulOp` In which case we can just add the check in `TranslateBatchMatMulOp` and use the same function for `BatchMatMul` and `BatchMatMulV2`

ngraph_bridge/ngraph_builder.cc
```diff
@@ -913,9 +913,37 @@ static Status TranslateBatchMatMulOp(
...
+    if (!((ng_lhs_shape[i] == ng_rhs_shape[i]) || (ng_lhs_shape[i] == 1) ||
+          (ng_rhs_shape[i] == 1))) {
+      return errors::InvalidArgument(
+          "ng_lhs_shape and ng_rhs_shape must be broadcastable for "
+          "BatchMatMulV2 "
+          "for each dimension",
+          i);
```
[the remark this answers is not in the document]
maxbubula: Done.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -913,9 +913,37 @@ static Status TranslateBatchMatMulOp(
...
+    } else if (ng_lhs_shape[i] != ng_rhs_shape[i]) {
+      return errors::Unimplemented(
+          "ng_lhs_shape and ng_rhs_shape must be the same for BatchMatMulV2 "
+          "for each dimension",
+          i);
```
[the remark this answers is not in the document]
maxbubula: I think the two `if` statements are necessary because, in the first one, we are checking if the shapes are not the same and not broadcastable than `InvalidArgument` is returned. But in the second check, the only availability of broadcasting is checked, and if it is broadcastable `Unimplemented` is returned. I think this is the essential difference. But I also changed the error message to be more informative and more diverse between these cases.

ngraph_bridge/ngraph_builder.cc
maxbubula: I was thinking about maintainability when more features will be needed in `BatchMatMulV2`, it will be easier to add them only changing according to translate function, without taking into account old `BatchMatMul` version. What's more when in code someone makes a mistake by using old `BatchMatMul` instead of `BatchMatMulV2` it could be working having only one translation function. I think such behavior can lead to confusion and complicate the debugging of the model.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -951,9 +951,45 @@ static Status TranslateBatchMatMulOp(
...
+    if (!((ng_lhs_shape[i] == ng_rhs_shape[i]) || (ng_lhs_shape[i] == 1) ||
+          (ng_rhs_shape[i] == 1))) {
+      return errors::InvalidArgument(
+          "ng_lhs_shape and ng_rhs_shape must be broadcastable for "
+          "BatchMatMulV2 "
+          "for each dimension. Failed to match dimension ",
+          i, " where lhs was ", ng_lhs_shape[i], " and rhs was ",
+          ng_rhs_shape[i]);
+    } else if ((ng_lhs_shape[i] == 1) || (ng_rhs_shape[i] == 1)) {
+      return errors::Unimplemented(
```
shresthamalik: Dont see the error thrown and the shapes matching

ngraph_bridge/ngraph_builder.cc
shresthamalik: @maxbubula Are we supporting the entire functionality of BatchMatMulV2 TF op or what is required for BERT alone?

ngraph_bridge/ngraph_builder.cc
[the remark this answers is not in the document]
sayantan-nervana: Ah, I see your point.
check 1 is for broadcastability (which TF allows)
but since we don't support it, we return Unimplemented.

The 2 ifs are fine then. Maybe we can change the error message a bit in the 2 cases to make that explicit

ngraph_bridge/ngraph_builder.cc
shresthamalik: can you add some comments. for e.g. shapes [2,3] and [1,3] are broadcast-able. Gives an idea why comparisons with 1 are being made. Also good to add the differences between BatchMatMul and BatchMatMulV2.
One clearly mentioned difference in the documentation, is that BatchMatMulV2 supports broadcasting in batch dimension

shresthamalik: Does nGraph not support the broadcast implementation?

test/test_math_ops.cpp
```diff
@@ -814,6 +814,56 @@ TEST(MathOps, BatchMatMul4DAdjXY) {
...
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "BatchMatMulV2", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}  // end of test op BatchMatMulV2
+
```
sindhu-nervana: Can we add more tests for BatchMatMulV2 like Adjx, Adjy and 4D  like we have for BatchMatmul https://github.com/tensorflow/ngraph-bridge/blob/master/test/test_math_ops.cpp#L517

ngraph_bridge/ngraph_builder.cc
maxbubula: We are supporting only the part of functionality required for BERT alone. I think that BatchMatMul (V1) also do not implement all the functionalities.

test/test_math_ops.cpp
maxbubula: Done ;)

ngraph_bridge/ngraph_builder.cc
shresthamalik: @avijit-nervana  Is this acceptable? BatchMatMulV2 will support only the functionality for BERT.

jbobba:
` "if` (n_dims >= 3 && backend_name == "CPU") `  is this still required in `TranslateBatchMatMulOp`? Can we remove the CPU check?

ngraph_bridge/ngraph_builder.cc
shresthamalik: Checked with @avijit-nervana  we can focus on just BERT functionality for now

PR #286, opened 2019-10-15
Sarkars/tf2ngraph infer out nodes

sayantan-nervana:
Make `--output_nodes` optional in `tf2ngraph` and in case it is not passed, print suggested output node names before failing

How to find output nodes:
1. For saved models that have output node info saved, parse the signaturedef to infer the output names
2. For others scan the graph to find nodes which are not input to other nodes and provide a list of suggested output nodes

Sample output for `python tf2ngraph.py --input_pbtxt ../test/test_axpy.pbtxt --output_pbtxt axpy_ngraph.pbtxt` (no --output_nodes passed)
```
Analysed graph for possible list of output nodes. Please supply one or more output node in --output_nodes
add of type Add
Traceback (most recent call last):
  File "tf2ngraph.py", line 443, in <module>
    main()
  File "tf2ngraph.py", line 431, in main
    raise ValueError("No output node name provided in --output_nodes")
ValueError: No output node name provided in --output_nodes

```

Also replace `assert`s and `raise`s by `exit_on_error`. `exit_on_error` (by default) prints the last error message and not the whole stack trace which is easier to read/understand UX-wise. Also expanded some error messages.

### DIFF
--- a/test/python/test_tf2ngraph_script.py
+++ b/test/python/test_tf2ngraph_script.py
@@ -47,13 +47,18 @@ class Testtf2ngraph(NgraphTest):
 
     @pytest.mark.parametrize(('commandline'),
                              (True,))  #TODO: add False for functional api test
-    @pytest.mark.parametrize(('inp_format', 'inp_loc', 'out_node_name'), (
-        ('pbtxt', 'sample_graph.pbtxt', 'out_node'),
-        ('savedmodel', 'sample_graph', 'out_node'),
-        ('pb', 'sample_graph.pb', 'out_node'),
-        ('pbtxt', 'sample_graph_nodevice.pbtxt', 'out_node'),
-        ('pbtxt', 'sample_graph_nodevice.pbtxt', None),
-    ))
+    @pytest.mark.parametrize(
+        ('inp_format', 'inp_loc', 'out_node_name'),
+        (
+            ('pbtxt', 'sample_graph.pbtxt', 'out_node'),
+            ('savedmodel', 'sample_graph', 'out_node'),
+            ('pb', 'sample_graph.pb', 'out_node'),
+            ('pbtxt', 'sample_graph_nodevice.pbtxt', 'out_node'),
+            ('pbtxt', 'sample_graph_nodevice.pbtxt', None),
+            # this test does not pass an output node
+            # and tf2ngraph is supposed to infer the output nodes
+            # and ask user to select one
+        ))
     @pytest.mark.parametrize(('out_format',), (
         ('pbtxt',),
         ('pb',),
@@ -144,10 +149,11 @@ class Testtf2ngraph(NgraphTest):
             assert np.isclose(res1, exp).all()
 
     def test_output_node_inference_for_saved_model(self):
-        # This saved model has input and output specified, and tf2ngraph should be able to infer it without user input
+        # The saved model we create in this pytest
+        # has input and output specified,
+        # and hence tf2ngraph should be able to infer it without user input
         export_dir = 'temp_pytest_savedmodel'
         out_loc = 'temp_pytest_pbtxt.pbtxt'
-        # TODO take care of file deletion
         try:
             shutil.rmtree(export_dir)
             os.remove(out_loc)
--- a/test/python/test_tf2ngraph_script.py
+++ b/test/python/test_tf2ngraph_script.py
@@ -197,7 +197,7 @@ class Testtf2ngraph(NgraphTest):
         rc = p.returncode
 
         # Since no output nodes were provided, we expect tf2ngraph to print out possible output nodes
-        assert "Analysed graph for possible list of output nodes. " in output.decode(
+        assert "Analyzed graph for possible list of output nodes. " in output.decode(
         )
         assert "Please supply one or more output node in --output_nodes\n" in output.decode(
         )
@@ -226,7 +226,7 @@ class Testtf2ngraph(NgraphTest):
         output, err = p.communicate()
         rc = p.returncode
 
-        diagnostic_log = "Analysed graph for possible list of output nodes. " + \
+        diagnostic_log = "Analyzed graph for possible list of output nodes. " + \
         "Please supply one or more output node in --output_nodes\n" + \
         "output of type Softmax\n"
         assert diagnostic_log in output.decode()
--- a/test/python/test_tf2ngraph_script.py
+++ b/test/python/test_tf2ngraph_script.py
@@ -201,8 +201,8 @@ class Testtf2ngraph(NgraphTest):
         )
         assert "Please supply one or more output node in --output_nodes\n" in output.decode(
         )
-        assert "output of type Softmax" in output.decode()
-        assert "add of type Add" in output.decode()
+        assert "Name: `output` Type: `Softmax`" in output.decode()
+        assert "Name: `add` Type: `Add`" in output.decode()
 
         # Since no output nodes were provided, we expect the test to fail with this error
         assert 'No output node name provided in --output_nodes' in err.decode()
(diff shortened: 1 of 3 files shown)


tools/tf2ngraph.py
```diff
@@ -334,6 +335,83 @@ def convert(inp_format, inp_loc, out_format, out_loc, output_nodes, ng_backend,
...
+                return guess_output_nodes(sess.graph_def)
+    elif inp_format == 'pbtxt' or inp_format == 'pb':
+        return guess_output_nodes(get_gdef_from_protobuf(inp_loc))
+    else:
+        raise ValueError(
+            "inp_format expected to be pb, pbtxt or savedmodel, but found ",
+            inp_format)
+
+
 def main():
```
avijit-nervana: Unfriendly error message: Please fix this:
```
Traceback (most recent call last):
  File "../ngraph-bridge/tools/tf2ngraph.py", line 443, in <module>
    main()
  File "../ngraph-bridge/tools/tf2ngraph.py", line 423, in main
    out_format, out_loc = filter_dict("output", args.__dict__)
  File "../ngraph-bridge/tools/tf2ngraph.py", line 262, in filter_dict
    len(current_format)) + " " + prefix + " formats, expected only 1"
AssertionError: Got 0 output formats, expected only 1
```
Change this to indicate that there is missing output file format. Then display possible formats.

tools/tf2ngraph.py
```diff
@@ -343,7 +421,16 @@ def main():
     args = prepare_argparser(allowed_formats)
     inp_format, inp_loc = filter_dict("input", args.__dict__)
     out_format, out_loc = filter_dict("output", args.__dict__)
-    output_nodes = args.output_nodes.split(',')
+    if args.output_nodes is None:
+        possible_out_nodes = infer_output_nodes(inp_format, inp_loc)
+        print(
+            "Analysed graph for possible list of output nodes. Please supply one or more output node in --output_nodes"
```
avijit-nervana: `Analyzed`. Also break the sentense into two so that it doesn't exceed. Apply to all other longer than 8- character sentences in this file.

Also mention if more than one output node is specified, what is the delimiter? Give an example.

tools/tf2ngraph.py
```diff
@@ -343,7 +421,16 @@ def main():
...
+        )
+        for out_node in possible_out_nodes:
+            print(out_node, "of type", possible_out_nodes[out_node])
```
avijit-nervana: Print this with quotes. Currently the output looks like:
```
Analysed graph for possible list of output nodes. Please supply one or more output node in --output_nodes
num_detections of type Identity
detection_boxes of type Identity
detection_scores of type Identity
detection_classes of type Identity
Traceback (most recent call last):
...
```
Instead, consider changing this to:
```
Name: `num_detections` Type: `Identity`
Name: `detection_boxes` Type: `Identity`
...
```
Give a space between the final line and the beginning of the stack trace.

tools/tf2ngraph.py
```diff
@@ -334,6 +335,83 @@ def convert(inp_format, inp_loc, out_format, out_loc, output_nodes, ng_backend,
...
+    # get rid of output slot (name:0 -> name)
+    split_colon = node_name.split(':')
+    if len(split_colon) == 1 or len(split_colon) == 2:
+        return split_colon[0]
+    else:
+        raise NotImplemented(
+            "Expected name to have <= 1 colons. Handle case with > 1 colons")
+
+
+def guess_output_nodes(graph_def):
```
avijit-nervana: Why is it called `guess`? Change it to `get_possible_output_node_names()`

test/python/test_tf2ngraph_script.py
```diff
@@ -89,12 +96,19 @@ def test_command_line_api(self, inp_format, inp_loc, out_format,
...
-                    out_format + ' ' + out_loc + ' --ng_backend ' + ng_device +
-                    ' --config_file ' + config_file_name +
-                    ("", " --precompile ")[precompile])
+                # out_node_str is empty if out_node_name is None.
+                # Automatic output node inference display diagnostic logs
+                # But the tf2ngraph call will still fail
+                out_node_str = ' ' if out_node_name is None else ' --output_nodes ' + out_node_name + ' '
+                try:
+                    command_executor(
+                        'python ../../tools/tf2ngraph.py --input_' +
```
avijit-nervana: This is a bad idea to hardcode relative paths. It makes directories non-portable. Rather use cmake install step to copy the tf2ngraph.py to this directory and run the test.

tools/tf2ngraph.py
```diff
@@ -343,7 +421,16 @@ def main():
...
     out_format, out_loc = filter_dict("output", args.__dict__)
-    output_nodes = args.output_nodes.split(',')
+    if args.output_nodes is None:
+        possible_out_nodes = infer_output_nodes(inp_format, inp_loc)
+        print(
+            "Analysed graph for possible list of output nodes. Please supply one or more output node in --output_nodes"
+        )
+        for out_node in possible_out_nodes:
+            print(out_node, "of type", possible_out_nodes[out_node])
+        raise ValueError("No output node name provided in --output_nodes")
```
avijit-nervana: Don't throw exception as it leads to printing the stack trace - which makes it hard to understand the error message.  Follow the error reporting style from this script: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/tools/import_pb_to_tensorboard.py

test/python/test_tf2ngraph_script.py
sayantan-nervana: cmake creates the following directory structure (mimicking what we have in the original source tree)

artifacts:
  -test
    -python
      -test_tf2ngraph_script.py
  -tools
    -tf2ngraph.py


the `../../tools/tf2ngraph.py` is needed because pytests run from `artifacts/test/python` and `../../tools/tf2ngraph.py` accesses `artifacts/tools`

The relative path works inside artifacts's copy of tf2ngraph.py (and does not access the original source tf2ngraph.py)

tools/tf2ngraph.py
sayantan-nervana: Sorry, I don't see the error reporting you mention.

the most related thing in that script is:
```
try:
  from tensorflow.contrib.tensorrt.ops.gen_trt_engine_op import *
except ImportError:
  pass
```


I can rewrite the function to not throw an error, but just print something, but then if tf2ngraph is in an automated script, it is much easier to detect the thrown error, than to parse the stdout for prints saying that it failed.

tools/tf2ngraph.py
sayantan-nervana: Changed `Got 0 output formats, expected only 1` to `Got 0 output formats, expected only 1. Please add --output_pb or --output_pbtxt or --output_savedmodel and pass the output model location`

tools/tf2ngraph.py
sayantan-nervana: Replaced `assert`s and `raise`s with `exit_on_error`

tools/tf2ngraph.py
```diff
@@ -94,6 +97,17 @@ def dump_json(json_name, optional_params=None, shape_hints=None):
             json.dump(dict_to_dump, fp)
 
 
+# This function controls how errors are handled.
+# For developers/debugging set raise_exception to True
+def exit_on_error(success, error_message, raise_exception=False):
```
avijit-nervana: the term raise_exception is a bit misleading as we are not really raising exception. Rename it to use_assert or something like assert_on_failure

avijit-nervana: Thinking more on this, why do we event need this flag? If we really want to assert (which shouldn't be done in a  user focused script) - then we need to modify all the call sites with that change. So we should just remove this IMO.

PR #289, opened 2019-10-18
Sindhu/bazel perf regression

sindhu-nervana:


bazel/ngraph.BUILD
```diff
@@ -285,3 +303,46 @@ cc_library(
     alwayslink = 1,
 )
 
+cc_library(
```
avijit-nervana: The interpreter backend wasn't needed but since you've done it let's keep it. However, the target doesn't have the same flags - so please add them. Also - please ensure that it's tested.

bazel/BUILD
```diff
@@ -111,11 +111,19 @@ cc_library(
         "-Wformat",
         "-Wformat-security",
         "-fstack-protector-all",
+        "-march=native",
```
avijit-nervana: Since we are copying the same flags everywhere, please use a Bazel function to define these and then call them wherever they are needed.

bazel/README.md
```diff
@@ -8,21 +8,24 @@ This directory contains scripts necessary to build the nGraph TensorFlow bridge
...
         ./configure_bazel.sh
-        bazel build libngraph_bridge.so
-        bazel build @ngraph//:libinterpreter_backend.so
-        bazel build @ngraph//:libcpu_backend.so
+        bazel build ngraph_bridge_lib
+        bazel build @ngraph//:cpu_backend
+
+To build C++ application `infer_multiple_networks` 
+
+        bazel build --cxxopt='-march=native' --cxxopt='-mtune=native' --cxxopt='-O3' infer_multiple_networks
```
avijit-nervana: Review this article: https://github.com/tensorflow/benchmarks/issues/272
They are also other articles that talk about specific compiler flags (such as this: https://medium.com/@sometimescasey/building-tensorflow-from-source-for-sse-avx-fma-instructions-worth-the-effort-fbda4e30eec3)

PR #291, opened 2019-10-18
Vishakha/data cache

vishakha-nervana:
Created generic data cache class.


ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,139 @@
...
+  // it will create item, put it in the cache and returns item and stattus
+  std::pair<Status, T2> LookUpOrCreate(
+      T1 key, std::function<std::pair<Status, T2>()> create_item,
+      std::function<void(T2)> callback_destroy_item);
+  std::pair<Status, T2> LookUpOrCreate(
+      T1 key, std::function<std::pair<Status, T2>()> create_item);
+
+ private:
+  std::unordered_map<T1, T2> m_ng_items_map;
+  std::list<T1> m_lru;
```
avijit-nervana: Use a std::deque instead.

test/test_ngraph_data_cache.cpp
```diff
@@ -0,0 +1,79 @@
...
+    create_count++;
+    if (barrier_->Block()) {
+      delete barrier_;
+    }
+    return std::make_pair(Status::OK(), 3);
+    ;
+  }
+};
+
+TEST_F(NGraphDataCacheTest, SameKeyMultiThread) {
```
avijit-nervana: Need to add a few more test cases such as the following:
1. Test the callback for eviction i.e., create a cache and then add more entries than the depth. Now check to make sure whether the eviction callback was called.
2. Test the second form of the `LookupOrCreate()` function which doesn't take the callback for the cache eviction.

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,139 @@
...
+        m_lru.pop_back();
+      }
+      // Add item to cache
+      auto it = m_ng_items_map.emplace(key, item);
+      if (it.second == true) {
+        m_lru.push_front(key);
+      }
+    }  // lock ends here.
+    if (need_to_evict) {
+      callback_destroy_item(item_to_evict);
```
avijit-nervana: This will most likely crash when the callback provided is nullptr (for the [second](https://github.com/tensorflow/ngraph-bridge/pull/291/files#diff-64b3714a6dbdccd436adf42cbd580d53R58) overload)

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,139 @@
...
+  ~NgraphDataCache();
+
+  // This method performs lookup in the cache for requested key, if not found
+  // it will create item, put it in the cache and returns item and stattus
+  std::pair<Status, T2> LookUpOrCreate(
+      T1 key, std::function<std::pair<Status, T2>()> create_item,
+      std::function<void(T2)> callback_destroy_item);
+  std::pair<Status, T2> LookUpOrCreate(
+      T1 key, std::function<std::pair<Status, T2>()> create_item);
+
```
avijit-nervana: Need to provide the following functions:
1. An iterator so that the caller can browse the cache.
2. Add a `RemoveItem(key)` method that removes an item from the cache 
2.1. A variation that doesn't have any callback
2.2. A variation that takes a destroy callback 
3. Add a `RemoveAll(destroy_callback)`

These are necessary to enable deterministic resource release (e.g., backend tensors/executables can be predictably removed to reclaim the resources).

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,139 @@
+// Executor Cache.h
```
avijit-nervana: Remove

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,139 @@
...
+    return std::make_pair(Status::OK(), item);
+  }
+
+  return status_item_pair;
+}
+
+template <typename T1, typename T2>
+std::pair<Status, T2> NgraphDataCache<T1, T2>::LookUpOrCreate(
+    T1 key, std::function<std::pair<Status, T2>()> callback_create_item) {
+  return LookUpOrCreate(key, callback_create_item, nullptr);
```
sayantan-nervana:
```suggestion
  return LookUpOrCreate(key, callback_create_item, [](T2){});
```

The argument expects a std function, so we should provide it an empty std function instead of nullptr

vishakha-nervana: I used that earlier, for some reason thought nullptr would be better

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,139 @@
...
+}
+
+template <typename T1, typename T2>
+class NgraphDataCache {
+ public:
+  NgraphDataCache(int depth);
+  ~NgraphDataCache();
+
+  // This method performs lookup in the cache for requested key, if not found
+  // it will create item, put it in the cache and returns item and stattus
```
kanvi-nervana: typo stattus

test/test_ngraph_data_cache.cpp
```diff
@@ -0,0 +1,79 @@
...
+using namespace std;
+namespace ng = ngraph;
+namespace tf = tensorflow;
+
+namespace tensorflow {
+namespace ngraph_bridge {
+namespace testing {
+
+#define ASSERT_OK(x) ASSERT_EQ((x), ::tensorflow::Status::OK());
+#define ASSERT_NOT_OK(x) ASSERT_NE((x), ::tensorflow::Status::OK());
```
kanvi-nervana: These can be removed since they are defined in the test_utilities.h
Also, move `#include test/test_utilities.h` from line 20 to line 30 as it is ngraph_bridge file

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,139 @@
...
+    if (found_in_cache) {
+      return std::make_pair(Status::OK(), m_ng_items_map.at(key));
+    }
+  }
+  // Item not found in cache, create item
+  T2 item;
+  auto status_item_pair = callback_create_item();
+
+  // If item is successfully created we will place in the cache.
+  if (status_item_pair.first == Status::OK()) {
```
kanvi-nervana: What happens if the item is not created successfully ?

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,178 @@
...
+    KeyType key, std::function<void(ValueType)> callback_destroy_item) {
+  if (m_ng_items_map.find(key) != m_ng_items_map.end()) {
+    callback_destroy_item(m_ng_items_map.at(key));
+    RemoveItem(key);
+  }
+}
+
+template <typename KeyType, typename ValueType>
+void NgraphDataCache<KeyType, ValueType>::RemoveItem(KeyType key) {
+  if (m_ng_items_map.find(key) != m_ng_items_map.end()) {
```
shresthamalik: Do we need to lock the look-up?

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,178 @@
...
+
+template <typename KeyType, typename ValueType>
+NgraphDataCache<KeyType, ValueType>::~NgraphDataCache() {
+  m_ng_items_map.clear();
+}
+
+template <typename KeyType, typename ValueType>
+void NgraphDataCache<KeyType, ValueType>::RemoveItem(
+    KeyType key, std::function<void(ValueType)> callback_destroy_item) {
+  if (m_ng_items_map.find(key) != m_ng_items_map.end()) {
```
shresthamalik: Do we need to lock the look-up?

test/test_ngraph_data_cache.cpp
```diff
@@ -0,0 +1,128 @@
...
+    return std::make_pair(Status::OK(), 3);
+  }
+
+  void DestroyItem(int i) {
+    item_evicted = true;
+    destroy_count++;
+  }
+};
+
+TEST_F(NGraphDataCacheTest, SameKeyMultiThread) {
```
shresthamalik: Please add some descriptions about the test

ngraph_bridge/ngraph_data_cache.h
sayantan-nervana: I think yes.

sayantan-nervana: The overload should not need to do the lookup etc.
1. RemoveItem(KeyType key, std::function<void(ValueType)> callback_destroy_item) could do all the work
2. RemoveItem(KeyType key) should just call 1. with a dummy function passed for callback_destroy_item

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,178 @@
...
+        auto key_to_evict = m_lru.back();
+        item_to_evict = m_ng_items_map.at(key_to_evict);
+        m_ng_items_map.erase(key_to_evict);
+        m_lru.pop_back();
+        need_to_evict = true;
+      }
+      // Add item to cache
+      auto it = m_ng_items_map.emplace(key, item);
+      if (it.second == true) {
+        m_lru.push_front(key);
```
shresthamalik: if item was added we move the key to front.
if item was not added, ie. the key was already there, still the key needs to be moved to front, since this has been recently used?

shresthamalik: The lru needs to be updated. dont see that happening

ngraph_bridge/ngraph_data_cache.h
vishakha-nervana: Agree, but There is a possibility that this function(The one which takes only key) can be called independent of the other one,
and there is no harm in checking it again

test/test_ngraph_data_cache.cpp
```diff
@@ -0,0 +1,128 @@
...
+      std::bind(&NGraphDataCacheTest_TestItemEviction_Test::DestroyItem, this,
+                std::placeholders::_1);
+
+  m_ng_data_cache.LookUpOrCreate("abc", create_item);
+  m_ng_data_cache.LookUpOrCreate("def", create_item, destroy_item);
+  m_ng_data_cache.LookUpOrCreate("efg", create_item, destroy_item);
+  ASSERT_EQ(item_evicted, false);
+  m_ng_data_cache.LookUpOrCreate("hij", create_item, destroy_item);
+  ASSERT_EQ(item_evicted, true);
+  m_ng_data_cache.LookUpOrCreate("klm", create_item);
```
sayantan-nervana: Assert after klm insertion, an item is evicted.

vishakha-nervana: ???

vishakha-nervana: ???

vishakha-nervana: ???

test/test_ngraph_data_cache.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+TEST_F(NGraphDataCacheTest, TestItemEviction) {
+  auto create_item = std::bind(
+      &NGraphDataCacheTest_TestItemEviction_Test::CreateItemNoBarrier, this);
+  auto destroy_item =
+      std::bind(&NGraphDataCacheTest_TestItemEviction_Test::DestroyItem, this,
+                std::placeholders::_1);
+
+  m_ng_data_cache.LookUpOrCreate("abc", create_item);
```
shresthamalik: Assert ok on LookUpOrCreate

test/test_ngraph_data_cache.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+TEST_F(NGraphDataCacheTest, RemoveItemTest) {
+  auto create_item = std::bind(
+      &NGraphDataCacheTest_RemoveItemTest_Test::CreateItemNoBarrier, this);
+  auto destroy_item =
+      std::bind(&NGraphDataCacheTest_RemoveItemTest_Test::DestroyItem, this,
+                std::placeholders::_1);
+
+  m_ng_data_cache.LookUpOrCreate("abc", create_item);
```
shresthamalik: All LookUpOrCreate can be asserted to return true. Might be worthwhile to create a test where lookuporcreate does not return status ok and then check that the item is not added

ngraph_bridge/ngraph_data_cache.h
vishakha-nervana: Agree, Done.

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,180 @@
...
+      KeyType key,
+      std::function<std::pair<Status, ValueType>()> callback_create_item);
+  void RemoveItem(KeyType key);
+  void RemoveItem(KeyType key,
+                  std::function<void(ValueType)> callback_destroy_item);
+  void RemoveAll(std::function<void(ValueType)> callback_destroy_item);
+
+ private:
+  std::unordered_map<KeyType, ValueType> m_ng_items_map;
+  std::deque<KeyType> m_lru;
```
shresthamalik: Does it make sense to have an assert soemwhere( in LookUpOrCreate/RemoveItmes) to check that m_ng_items_map.size()== m_lru.size().  they should always be the same. and we might find error cases

shresthamalik: @sayantan-nervana @avijit-nervana

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,180 @@
...
+}
+
+template <typename KeyType, typename ValueType>
+void NgraphDataCache<KeyType, ValueType>::RemoveAll(
+    std::function<void(ValueType)> callback_destroy_item) {
+  absl::MutexLock lock(&m_mutex);
+  for (auto it = m_ng_items_map.begin(); it != m_ng_items_map.end(); it++) {
+    callback_destroy_item(it->second);
+  }
+  m_ng_items_map.erase(m_ng_items_map.begin(), m_ng_items_map.end());
```
shresthamalik: erase m_lru? all keys from m_lru should also be deleted?

test/test_ngraph_data_cache.cpp
shresthamalik: Assert the item returned is 3. atleast at one place

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,180 @@
...
+    absl::MutexLock lock(&m_mutex);
+    auto it = m_ng_items_map.find(key);
+    found_in_cache = (it != m_ng_items_map.end());
+    if (found_in_cache) {
+      return std::make_pair(Status::OK(), m_ng_items_map.at(key));
+    }
+  }
+  // Item not found in cache, create item
+  ValueType item;
+  auto status_item_pair = callback_create_item();
```
shresthamalik: try + catch the invocation of callbacks . we can throw errors like "Failed to create an item. Invalid Callback to Create"

From here:
https://en.cppreference.com/w/cpp/utility/functional/function
The stored callable object is called the target of std::function. If a std::function contains no target, it is called empty. Invoking the target of an empty std::function results in std::bad_function_call exception being thrown.

test/test_ngraph_data_cache.cpp
vishakha-nervana: done

ngraph_bridge/ngraph_data_cache.h
vishakha-nervana: Done

ngraph_bridge/ngraph_data_cache.h
vishakha-nervana: Done

vishakha-nervana: Done.

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,205 @@
...
+    }
+  }
+  // Item not found in cache, create item
+  ValueType item;
+  pair<Status, ValueType> status_item_pair;
+  try {
+    status_item_pair = callback_create_item();
+  } catch (std::bad_function_call& exception) {
+    return std::make_pair(
+        errors::Internal(
```
shresthamalik: all the callbacks need to be try catched

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -0,0 +1,205 @@
...
+}
+
+template <typename KeyType, typename ValueType>
+Status NgraphDataCache<KeyType, ValueType>::RemoveAll(
+    std::function<void(ValueType)> callback_destroy_item) {
+  absl::MutexLock lock(&m_mutex);
+  for (auto it = m_ng_items_map.begin(); it != m_ng_items_map.end(); it++) {
+    callback_destroy_item(it->second);
+  }
+  if (m_ng_items_map.size() != m_lru.size()) {
```
shresthamalik: the if needs to be after the clear

shresthamalik: sorry my bad, this is fine

ngraph_bridge/ngraph_data_cache.h
vishakha-nervana: All other callbacks are calling destroy_item(),
not sure what is the right thing to do if destoy_item in invalid,
In case of LookUPOrCreate(), should I still continue to create the item or not,
and In case of RemoveItem(), should I still be destroying item from cache or not.

PR #294, opened 2019-10-22
Sarkars/upgrade to ngcore master

sayantan-nervana:
1. Cherry picks sindhu's changes of upgrading to ngcore 26 (in bridge r20)
2. Upgrades build_ngtf and CMakeLists to bleeding edge master SHA of ngcore
3. Make changes in parallel executor related code to make it compat with read/write API changes
4. Upgrades mkldnn to 0.21.2 (from 0.21.2)


ngraph_bridge/version.cc
```diff
@@ -24,15 +24,15 @@
 // nGraph-TensorFlow bridge uses semantic versioning: see http://semver.org/
 
 #define NG_TF_MAJOR_VERSION 0
-#define NG_TF_MINOR_VERSION 19
+#define NG_TF_MINOR_VERSION 20
```
avijit-nervana: This should be 21.0-rc0 as 20 is already on r20 branch. But we won't cur the release branch until we need to move too r22.

PR #296, opened 2019-10-24
sarkars/Fix NgraphSerialize

sayantan-nervana:
`NgraphSerialize` is dumping empty jsons right now.


ngraph_bridge/ngraph_utils.cc
```diff
@@ -347,7 +347,7 @@ Status NgraphSerialize(const std::string& file_name,
         "Passed a null pointer as ng function to serialize");
   }
   try {
-    ngraph::serialize(ng_function, json_indentation);
+    serialized = ngraph::serialize(ng_function, json_indentation);
```
shresthamalik: good catch! 👍

avijit-nervana: So - how did it ever worked? Are there any unit tests to catch this?

PR #301, opened 2019-10-28
gopoka/Implementation of UnsortedSegmentSum op.

gopoka:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -4825,6 +4825,55 @@ static Status TranslateUnpackOp(const Node* op,
...
+  def unsorted_segment_sum(inputs, segment_ids, num_segments):
+    inputs = tf.convert_to_tensor(inputs)
+    segment_ids = tf.convert_to_tensor(segment_ids)
+
+    shape = inputs.get_shape().as_list()
+    seg_ids = segment_ids.get_shape().as_list()
+    result = tf.Variable(tf.zeros([num_segments] + shape[len(seg_ids):],
+                                  dtype=inputs.dtype))
+
+    return tf.scatter_add(result, segment_ids, inputs)
```
sayantan-nervana: lets add this as a pytest.
Examples can be found here: https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/test_prod.py

test/test_math_ops.cpp
```diff
@@ -1632,6 +1632,30 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
...
+// Test op: UnsortedSegmentSum
+TEST(MathOps, UnsortedSegmentSum) {
+  Scope root = Scope::NewRootScope();
+  Tensor A(DT_FLOAT, TensorShape({3, 4}));
+  Tensor B(DT_INT32, TensorShape({3}));
+  Tensor C(DT_INT32, TensorShape({}));
+
+  AssignInputValues(A, std::vector<float>{1.f, 2.f, 3.f, 4.f, 5.f, 6.f, 7.f,
+                                          8.f, 4.f, 3.f, 2.f, 1.f});
+  AssignInputValues(B, std::vector<int>{0, 1, 0});
```
sayantan-nervana: Please add tests where:
1. Segment IDs are > 1, so for example seg ids 0,1,2,3
2. Some missing seg ids, such as: 0,1,3,4

ngraph_bridge/ngraph_builder.cc
```diff
@@ -4825,6 +4825,55 @@ static Status TranslateUnpackOp(const Node* op,
...
+
+  shared_ptr<ng::Node> ng_input, ng_segment_ids, ng_num_segments;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_segment_ids,
+                                   &ng_num_segments));
+
+  int num_segments;
+  std::vector<int64> tmp_num_segments;
+  TF_RETURN_IF_ERROR(
+      GetStaticInputVector(op, 2, static_input_map, &tmp_num_segments));
+  num_segments = tmp_num_segments[0];
```
sayantan-nervana: Do we need to check and return a bad status if `num_segments<max(ng_num_segments)`.. or maybe TF itself throws an error if we try to do that?

ngraph_bridge/ngraph_builder.cc
```diff
@@ -4825,6 +4825,55 @@ static Status TranslateUnpackOp(const Node* op,
...
+    seg_ids = segment_ids.get_shape().as_list()
+    result = tf.Variable(tf.zeros([num_segments] + shape[len(seg_ids):],
+                                  dtype=inputs.dtype))
+
+    return tf.scatter_add(result, segment_ids, inputs)
+  */
+
+  shared_ptr<ng::Node> ng_input, ng_segment_ids, ng_num_segments;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_segment_ids,
+                                   &ng_num_segments));
```
sayantan-nervana: Input number 2 is actually read in by `GetStaticInputVector` in the next line, so we dont need to read in `ng_num_segments` here.

We can pass:
```
TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_segment_ids, nullptr));
```
to indicate that there are 3 inputs, of which get the first 2, but we dont care about the third

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -650,7 +657,7 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
       set_attributes_map["TopKV2"] = SetStaticInputs({1});
       set_attributes_map["Tile"] = SetStaticInputs({1});
       set_attributes_map["Transpose"] = SetStaticInputs({1});
-
+      set_attributes_map["UnsortedSegmentSum"] = SetStaticInputs({1, 2});
```
shresthamalik: Both input index 1 and 2 are marked as static inputs. but in the translation only index 2 is being read as static input

test/test_math_ops.cpp
gopoka: Done.

ngraph_bridge/ngraph_builder.cc
gopoka: Fixed, thanks.

ngraph_bridge/ngraph_builder.cc
gopoka: I don't see why this should be a pytest? The comment was there only to show how we can replace `UnsortedSegmentSum` with `ScatterAdd`.

ngraph_bridge/ngraph_mark_for_clustering.cc
shresthamalik: If you reading index 1 as static only to check for errors then we should not be doing that, coz it will affect the performance

sayantan-nervana: marking a node as static is not free. it will cause graph splits (and hence perf issues). We do not want to mark nodes static until absolutely necessary

gopoka: Ok, removed `input 1` from static map.

test/test_math_ops.cpp
```diff
@@ -1677,6 +1677,122 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
...
+  AssignInputValues(C, 2);
+
+  vector<int> static_input_indexes = {1, 2};
```
shresthamalik: please change the static_input_indexes here also

ngraph_bridge/ngraph_builder.cc
kanvi-nervana: Should we add a test for the same like a test for the failing case?

test/test_math_ops.cpp
shresthamalik: @gopoka Please change the static_input_indexes

gopoka: Done.

ngraph_bridge/ngraph_builder.cc
gopoka: Can you point me to any example of C++ test in bridge that tests for failure?

PR #271, opened 2019-09-30
Kanvi/remove files

kanvi-nervana:
ngtf_graph_viewer.py can now be removed because we can use the script from tensorflow import_pb_to_tensorboard.py + our tf2ngraph to do the same thing. So all that code is not required. I updated the README with the instructions on how to use it.
And the tools/run_tf2ngraph_model.py was uploaded accidentally with a PR for the aot change. It was not reviewed and is not required.

PR #303, opened 2019-10-29
Kanvi/partial revert remove files

kanvi-nervana:
This PR reverts Kanvi/remove files (#271) since the new way of protobuf visualization that was introduced in #271 would only work with grappler builds. Hence, the old way needs to kept as is and new instructions are only added in the diagnostics/README. 
Also, the tools/run_tf2ngraph_model.py is removed(that was as part of #271) since that was uploaded accidentally with the PR for aot change and is not required.


diagnostics/README.md
```diff
@@ -42,6 +42,16 @@ To get a **full** dump use the following set of flags
...
+* Change directory to this diagnostics folder
+* Run this script to parse the dumped graphs to know which encapsulate a node belongs to. At this step nodemap.pkl is created: ```python get_node_encapsulate_map.py ./path/to/folder/where/run_TF_network.py/exists/where/the/dumps/were/created/in/the/last/step/ nodemap.pkl```
+* Modify the graphdef and dump TB file in ```./vis``` using encapsulate information in ```nodemap.pkl```: ```python ngtf_graph_viewer.py -c nodemap.pkl ./path/to/original_network_pbtxtfile.pbtxt ./vis```. If you do not have the pbtxt of the original tensorflow graph, you can dump it from your script using [write_graph](https://www.tensorflow.org/api_docs/python/tf/train/write_graph)
+*  View the original network with encapsulate information by running tensorboard, using the files created in ```./vis```.
+* Note: you may need to preprocess your protobuffers (pbtxt) to remove the `_class` attribute if any of the above steps result in an error such as `{NodeType} expects to be colocated with unknown node ...`. In that case, first run ```python remove_protobuf_class_attribute.py -d /path/to/pbtxt/to/process/``` and then follow the above steps again.
```
sayantan-nervana: Lets move the `Another way of protobuf visualization` section here and add info that in case we have a grappler build we can do it the easy way, or if we dont, we need to do it the hard way (as mentioned in the green parts being re-added here)

PR #304, opened 2019-10-30
Vishakh1/mergingto executor

vishakha-nervana:
Plugging in data cache into ngraph_executor.cc.

In ngraph_executor.cc:

GetNgExecutable() is now replaced by GetExecutableAndTensors().

Got Rid of GetTensorsFrom Pipeline() as GetExecutableAndTensor() does that too now.

 GetNgFunction() and ReturnPipelinedTensors() not needed anymore.

GetExecutableAndTensors() calls LookUpOrCreate(), which later calls CreateCallback() to create executable and pipelinedtensor, if item is not found in cache.

Ngraph Data cache now contains:
 ..........Sig............> ----ng_executable
Signature---->-----serailized_ng_function
 ......Sig.......... >------pipeline_tensors

So removed the all the maps from ngraph_executor.h, that were holding above mentioned data.


ngraph_bridge/ngraph_executor.h
```diff
@@ -47,36 +48,33 @@ class NGraphExecutor {
...
-      std::shared_ptr<ngraph::runtime::Executable> ng_exec, size_t idx) {
-    lock_guard<mutex> lock(m_mutex);
-    m_executable_pipelined_tensors_map.at(ng_exec)->return_tensors(idx);
-  }
+  // Callback function called from NgraphDataCache's LookUpOrCreateItem() method
+  // Creates ng_executable, serialized_ng_function, and initializes I/O
+  // TensorPipeline
+  std::pair<Status, std::tuple<std::shared_ptr<ngraph::runtime::Executable>,
+                               std::string, shared_ptr<PipelinedTensorsStore>>>
+  CreateCallback(std::string signature, std::vector<TensorShape> input_shapes,
```
shresthamalik: Rename this function. It sounds generic when it is not

shresthamalik:
Please add more description to the changes. Bullet items would do.

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -405,31 +405,29 @@ void NGraphEncapsulateOp::ComputeUsingParallelExecutor(OpKernelContext* ctx) {
...
-
+  OP_REQUIRES_OK(
+      ctx, m_parallel_executor->GetExecutableAndTensors(
+               tf_input_tensors, ng_exec, pipelined_tensor_store, cache_hit));
+  io_tensors = pipelined_tensor_store.get()->get_tensors();
+  auto status = Status::OK();
+  if (std::get<0>(io_tensors) < 0) {
+    status = errors::Internal("No free tensor available");
+  }
+  OP_REQUIRES_OK(ctx, status);
```
shresthamalik: There is another utility OP_REQUIRES use that to throw an error

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -405,31 +405,29 @@ void NGraphEncapsulateOp::ComputeUsingParallelExecutor(OpKernelContext* ctx) {
...
-  // Get ngraph executable and inputs information
+  // Get ngraph executable and inputs information and Pipelined tensors
   bool cache_hit;
-  OP_REQUIRES_OK(ctx, m_parallel_executor->GetNgExecutable(tf_input_tensors,
-                                                           ng_exec, cache_hit));
-
+  OP_REQUIRES_OK(
+      ctx, m_parallel_executor->GetExecutableAndTensors(
+               tf_input_tensors, ng_exec, pipelined_tensor_store, cache_hit));
+  io_tensors = pipelined_tensor_store.get()->get_tensors();
```
shresthamalik: No need to do .get(). Directly calling get_tensors() should work

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +219,169 @@ Status NGraphExecutor::GetNgExecutable(
...
+                op_backend);
+  // Get NgItems i.e. ng_executable, serialized ng_functions from Data Cache
+  auto status_ng_item_pair = m_ng_data_cache.LookUpOrCreate(
+      signature, create_ng_items_callback, destroy_ng_items_callback);
+
+  if (status_ng_item_pair.first == Status::OK()) {
+    ng_exec = std::get<0>(status_ng_item_pair.second);
+    pts = std::get<2>(status_ng_item_pair.second);
+  }
+  std::cout << "Value of cache hit is :" << cache_hit << std::endl;
```
shresthamalik: remove std::cout

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +219,169 @@ Status NGraphExecutor::GetNgExecutable(
...
-    } else {
-      auto itr = m_aot_functions.find(signature);
-      if (itr == m_aot_functions.end()) {
-        return errors::Internal(
-            "Expected to find AOT precompiled ng function of signature: ",
-            signature);
-      }
-      serialized_ng_func = itr->second;
-    }
+  cache_hit = true;
```
sayantan-nervana: We don't have to set `cache_hit = true;`. The callback should be setting it to whatever needs to be set

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +219,169 @@ Status NGraphExecutor::GetNgExecutable(
...
+      std::bind(&NGraphExecutor::DestroyCallback, this, std::placeholders::_1,
+                op_backend);
+  // Get NgItems i.e. ng_executable, serialized ng_functions from Data Cache
+  auto status_ng_item_pair = m_ng_data_cache.LookUpOrCreate(
+      signature, create_ng_items_callback, destroy_ng_items_callback);
+
+  if (status_ng_item_pair.first == Status::OK()) {
+    ng_exec = std::get<0>(status_ng_item_pair.second);
+    pts = std::get<2>(status_ng_item_pair.second);
+  }
```
sayantan-nervana: get<1> gives the function, I guess? We should be returning that as well. And using that to try-catch the `ng_exec->call` and dump the serialized function in case of errors: https://github.com/tensorflow/ngraph-bridge/blob/vishakh1/MergingtoExecutor/ngraph_bridge/ngraph_encapsulate_op.cc#L459

currently the `ng_exec->call` is not try-catched in master, but maybe we can use this opportunity to add back the try-catch, similar to the `impl` case: https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_encapsulate_op.cc#L761

ngraph_bridge/ngraph_executor.h
```diff
@@ -118,25 +121,12 @@ class NGraphExecutor {
...
-  std::unordered_map<
-      std::shared_ptr<ngraph::runtime::Executable>,
-      std::vector<std::pair<void*, shared_ptr<ng::runtime::Tensor>>>>
-      m_ng_exec_input_cache_map;
-
-  std::unordered_map<
-      std::shared_ptr<ngraph::runtime::Executable>,
-      std::vector<std::pair<void*, shared_ptr<ng::runtime::Tensor>>>>
-      m_ng_exec_output_cache_map;
+  NgraphDataCache<std::string,
```
shresthamalik: Provide description. Key is the string (computed signature). what is the string that is part of the tuple which is the value?

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +219,169 @@ Status NGraphExecutor::GetNgExecutable(
...
-  return Status::OK();
+  // Get NgExecutable
+  auto status_ng_exec_pair =
+      GetNgExecutable(signature, ng_function, serialized_ng_func, op_backend);
+  // Create PipelinedTensorStore
+  if (status_ng_exec_pair.first == Status::OK()) {
+    ng_exec = status_ng_exec_pair.second;
+    auto status_ng_pts_pair = InitializeIOTensorPipeline(ng_exec);
+    pts = status_ng_pts_pair.second;
+    return std::make_pair(status_ng_pts_pair.first,
```
sayantan-nervana: This `return` can be removed. Only 1 return at the very end of the function takes care of both cases I think

ngraph_bridge/ngraph_executor.cc
vishakha-nervana: If item is found in the cache then callback will never be called, hence setting true by default

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +219,169 @@ Status NGraphExecutor::GetNgExecutable(
...
-  // Lookup the function from the Map
-  auto it = m_serialized_ng_function_map.find(ng_exec);
-  if (it == m_serialized_ng_function_map.end()) {
-    return errors::Internal("Function not found for this executable");
-  }
+std::pair<Status, std::shared_ptr<ngraph::runtime::Executable>>
+NGraphExecutor::GetNgExecutable(std::string signature,
+                                std::shared_ptr<ngraph::Function>& ng_function,
+                                string serialized_ng_func,
+                                ng::runtime::Backend*& op_backend) {
```
sayantan-nervana: `op_backend` is a `const` I think.

Also `serialized_ng_func` can be made `const string&`

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +219,169 @@ Status NGraphExecutor::GetNgExecutable(
...
+  // Generate forwarding call to Callback functions
+  // CreateCallback and DestroyCallback
+  auto create_ng_items_callback =
+      std::bind(&NGraphExecutor::CreateCallback, this, std::placeholders::_1,
+                input_shapes, static_input_map, op_backend, &cache_hit);
+  auto destroy_ng_items_callback =
+      std::bind(&NGraphExecutor::DestroyCallback, this, std::placeholders::_1,
+                op_backend);
+  // Get NgItems i.e. ng_executable, serialized ng_functions from Data Cache
+  auto status_ng_item_pair = m_ng_data_cache.LookUpOrCreate(
```
shresthamalik: Interesting LookUpOrCreate doesnt return that signal, whether the item was created or not.

vishakha-nervana: Adding to it.

ngraph_bridge/ngraph_executor.cc
vishakha-nervana: No it is needed because if Status of InitializeIOTensorPipeline() is not Okay,
then we return its status otherwise other status.

ngraph_bridge/ngraph_executor.h
vishakha-nervana: Please suggest some name

ngraph_bridge/ngraph_encapsulate_op.cc
vishakha-nervana: @shresthamalik I used OP_REQUIRES_OK, becuase it is I am getting status, can you please provide 
an example for this particular case on how to use OP_REQUIRES here?

ngraph_bridge/ngraph_executor.h
vishakha-nervana: Done

ngraph_bridge/ngraph_executor.h
```diff
@@ -88,11 +87,17 @@ class NGraphExecutor {
   }
 
  private:
+  // This method is called from CreateCallback(), It compiles ngraph
+  // Or load ng_executable from backend in case of AOT
+  std::pair<Status, std::shared_ptr<ngraph::runtime::Executable>>
+  GetNgExecutable(std::string signature,
```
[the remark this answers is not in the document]
vishakha-nervana: This is not true, CreateCallback(), creates serailized_ng_function,
It calls GetNgExecutable() only to get ng_executable, for PipleinedTensors 
a another method InitializeIOTemsor() is called from CreateCallback(),
so the name GetNgExecutable() makes sense.

ngraph_bridge/ngraph_executor.cc
```diff
@@ -223,14 +192,25 @@ Status NGraphExecutor::ComputeSignature(
...
-Status NGraphExecutor::GetNgExecutable(
+Status NGraphExecutor::GetExecutableFunctionAndTensors(
     const std::vector<Tensor>& tf_input_tensors,
-    std::shared_ptr<ngraph::runtime::Executable>& ng_exec, bool& cache_hit) {
+    std::shared_ptr<ngraph::runtime::Executable>& ng_exec,
+    std::string& serialized_ng_func, shared_ptr<PipelinedTensorsStore>& pts,
+    bool& cache_hit) {
+  std::stringstream signature_ss;
+  std::vector<TensorShape> input_shapes;
+  std::vector<const Tensor*> static_input_map;
```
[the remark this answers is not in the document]
vishakha-nervana: In previous implementation of ngraph_executor.cc it was locked,
but I don't think it is needed anymore. Others please provide your input.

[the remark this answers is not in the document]
sindhu-nervana: @avijit-nervana ?

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -458,7 +453,34 @@ void NGraphEncapsulateOp::ComputeUsingParallelExecutor(OpKernelContext* ctx) {
 
   // And execute
   ngraph::Event event_execute_graph("Execute Graph", "", "");
-  ng_exec->call(get<2>(io_tensors), get<1>(io_tensors));
+  BackendManager::LockBackend(m_parallel_executor->GetOpBackendName());
+  NGRAPH_VLOG(4) << "NGraphEncapsulateOp::Compute call starting for cluster "
```
shresthamalik: why it this diff showing up?

ngraph_bridge/ngraph_executor.cc
```diff
@@ -223,14 +192,25 @@ Status NGraphExecutor::ComputeSignature(
 }
 
 //---------------------------------------------------------------------------
-//  NGraphExecutor::GetNgExecutable
+//  NGraphExecutor::GetExecutableAndTensors
```
shresthamalik: obsolete comment

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -516,7 +538,7 @@ void NGraphEncapsulateOp::ComputeUsingParallelExecutor(OpKernelContext* ctx) {
 
   // Now return them to the cache
   ngraph::Event event_return_tensor("Return Tensor", "", "");
-  m_parallel_executor->ReturnPipelinedTensors(ng_exec, get<0>(io_tensors));
+  pipelined_tensor_store->return_tensors(get<0>(io_tensors));
```
shresthamalik: m_parallel_executor->ReturnPipelinedTensors used to be guarded by the mutex, which is not required now. We created using the cache's lock. Each thread will get a different id and then can return it without the mutex. Is that the right understanding?

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +220,167 @@ Status NGraphExecutor::GetNgExecutable(
...
-
-    // caching ng_function to serialize to ngraph if needed
-    m_serialized_ng_function_map[ng_exec] = serialized_ng_func;
-
-    m_lru.push_front(signature);
-    // Memory after
-    MemoryProfile(vm, rss);
-    auto delta_vm_mem = vm - vm0;
-    auto delta_res_mem = rss - rss0;
-    NGRAPH_VLOG(1) << "NGRAPH_TF_CACHE_PROFILE: OP_ID: " << m_instance_id
```
shresthamalik: We decided to get rid of the memory profiling?

sayantan-nervana: We added this for the LRU cache. But now we have a fixed size cache.

I am not sure if this is still effective or correct. We can keep this around I guess.

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +220,167 @@ Status NGraphExecutor::GetNgExecutable(
...
-        return errors::Internal(
-            "Expected to find AOT precompiled ng function of signature: ",
-            signature);
-      }
-      serialized_ng_func = itr->second;
-    }
+  // Generate forwarding call to Callback functions
+  // CreateCallback and DestroyCallback
+  auto create_ng_items_callback =
+      std::bind(&NGraphExecutor::CreateCallback, this, std::placeholders::_1,
```
shresthamalik: std::placeholders::_1 with bind, enables passing the signature?

shresthamalik: https://en.cppreference.com/w/cpp/utility/functional/placeholders

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +220,167 @@ Status NGraphExecutor::GetNgExecutable(
...
     }
-    ng_exec = it->second;
-    cache_hit = true;
-    NGRAPH_VLOG(1) << "Compilation cache hit: " << m_node_name;
+#endif
   }
-  return Status::OK();
+  // Get NgExecutable
+  auto status_ng_exec_pair =
+      GetNgExecutable(signature, ng_function, serialized_ng_func, op_backend);
```
shresthamalik: why are we passing both ng_functiona and serialized_ng_func?

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +220,167 @@ Status NGraphExecutor::GetNgExecutable(
...
+      stringstream serialized_exec_read;
+      serialized_exec_read << (itr->second);
+      ng_exec = op_backend->load(serialized_exec_read);
+    } else {
+      ng_exec = op_backend->compile(ng_function);
+    }
   } catch (const std::exception& exp) {
-    return errors::Internal("Failed to deserialize ngraph function");
+    BackendManager::UnlockBackend(m_op_backend_name);
+    Status st = StringToFile("tf_function_error_" + m_node_name + ".json",
```
shresthamalik: we can move this dumping of serialized ngfunction in the calling function. we dont need to send both ng-func and serialized ng-func in that case

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +220,167 @@ Status NGraphExecutor::GetNgExecutable(
...
-    cache_hit = true;
-    NGRAPH_VLOG(1) << "Compilation cache hit: " << m_node_name;
+#endif
   }
-  return Status::OK();
+  // Get NgExecutable
+  auto status_ng_exec_pair =
+      GetNgExecutable(signature, ng_function, serialized_ng_func, op_backend);
+  // Create PipelinedTensorStore
+  if (status_ng_exec_pair.first == Status::OK()) {
```
shresthamalik: Later Refactoring: we can look at changing the signatures from tuples to just Status. That way we can use TF_RETURN_IF_ERROR and avoid the series of checking for Statuses.

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +220,167 @@ Status NGraphExecutor::GetNgExecutable(
...
+//---------------------------------------------------------------------------
+//  NGraphExecutor::DestroyCallback
+//---------------------------------------------------------------------------
+void NGraphExecutor::DestroyCallback(
+    std::tuple<std::shared_ptr<ngraph::runtime::Executable>, std::string,
+               shared_ptr<PipelinedTensorsStore>>
+        evicted_ng_item,
+    ng::runtime::Backend*& op_backend) {
+  std::shared_ptr<ngraph::runtime::Executable> evicted_ng_exec;
+  std::tie(evicted_ng_exec, std::ignore, std::ignore) = evicted_ng_item;
```
shresthamalik: i see we are ignoring the other two. String might be fine. But pipelined_tensor_store might need a reset

test/test_parallel_executor.cpp
```diff
@@ -194,18 +140,16 @@ TEST(ParallelExecutor, ExecuteOnSingleThread) {
...
-
+  shared_ptr<PipelinedTensorsStore> pts;
+  std::tuple<int, PipelinedTensorVector, PipelinedTensorVector> io_tensors;
   // Call the Executor to compile the funcion
   bool cache_hit = false;
-  ASSERT_OK(executor.GetNgExecutable(tf_input_tensors, ng_exec, cache_hit));
+  std::string ser_ng_func;
+  ASSERT_OK(executor.GetExecutableFunctionAndTensors(
+      tf_input_tensors, ng_exec, ser_ng_func, pts, cache_hit));
+  io_tensors = pts.get()->get_tensors();
```
shresthamalik: It seems a few tests have been removed. Have we replaced them? Are they no longer required?

test/test_parallel_executor.cpp
```diff
@@ -194,18 +140,16 @@ TEST(ParallelExecutor, ExecuteOnSingleThread) {
 
   std::vector<Tensor> tf_input_tensors{x, y};
   shared_ptr<ngraph::runtime::Executable> ng_exec;
-
+  shared_ptr<PipelinedTensorsStore> pts;
```
shresthamalik: We dont have any tests for this PR specifically (integration of cache). Can we run RESNET50 and check there is no cache miss. If we can create a test like this it will be even better. Suggestions welcome.

ngraph_bridge/ngraph_encapsulate_op.cc
vishakha-nervana: I don't think its needs to be locked, Got rid of ReturnPipelinedTensor(), pts inside the cache is locked and safe, outside it is local to the thread, so do not need protection.

ngraph_bridge/ngraph_executor.cc
vishakha-nervana: Yes are not doing profiling anymore.

ngraph_bridge/ngraph_executor.cc
vishakha-nervana: both are needed inside GetExecutable

ngraph_bridge/ngraph_encapsulate_op.cc
vishakha-nervana: Added try cacth around ng_exec-> call(), which was missing, should have been added at the time of parallel_executor execution.

ngraph_bridge/ngraph_executor.cc
vishakha-nervana: I understand what you are saying, but I don't think we need to reset pipelined_tensor_store here, at it get erased in the cache.

ngraph_bridge/ngraph_executor.cc
vishakha-nervana: Idea is too keep  executable creation logic separate from calling, I think what you are suggesting would defeat the purpose. If you want we make discuss it.

test/test_parallel_executor.cpp
```diff
@@ -340,248 +282,6 @@ TEST(ParallelExecutor, ExecuteOnSingleThread8Bit) {
...
-  std::thread thread0(worker, 0);
-  std::thread thread1(worker, 1);
-  // std::thread thread2(worker, 33);
-
-  thread0.join();
-  thread1.join();
-  // thread2.join();
-}
-
-TEST(ParallelExecutor, ExecuteOnMultipleThreads) {
```
kanvi-nervana: Are these tests not needed anymore?

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +220,167 @@ Status NGraphExecutor::GetNgExecutable(
...
+                                          m_graph.get(), ng_function);
+    if (status != Status::OK()) {
+      return std::make_pair(status,
+                            std::make_tuple(ng_exec, serialized_ng_func, pts));
     }
-    // Evict the cache if the number of elements exceeds the limit
-    const char* cache_depth_specified =
-        std::getenv("NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH");
-    if (cache_depth_specified != nullptr) {
-      my_function_cache_depth_in_items = atoi(cache_depth_specified);
```
shresthamalik: Incorporate this flag.

ngraph_bridge/ngraph_executor.cc
shresthamalik: Please change as discussed.

ngraph_bridge/ngraph_executor.cc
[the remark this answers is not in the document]
sindhu-nervana: @vishakha-nervana - can you comment on why we don't need a lock here?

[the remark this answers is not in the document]
shresthamalik: I think we dont need lock for compute signature.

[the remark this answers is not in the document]
vishakha-nervana: > @vishakha-nervana - can you comment on why we don't need a lock here?

ComputeSignature() creates signature, We want to be able to create multi signature parallel, signature is local to thread so each thread will have its own copy of signature, hence it is thread safe.

test/test_parallel_executor.cpp
```diff
@@ -110,72 +110,18 @@ TEST(ParallelExecutor, CompilerTest) {
...
-  std::shared_ptr<ngraph::Function> ng_function;
-  ASSERT_EQ(executor.GetNgFunction(ng_exec, ng_function),
-            tensorflow::Status::OK());
-
-  // Validate the nGraph Function
-  const auto& parameters = ng_function->get_parameters();
-  ASSERT_EQ(2, parameters.size());
-}
-
-TEST(ParallelExecutor, PipelinedTensorCreate) {
```
vishakha-nervana: This test is not needed anymore, as I got rid of GetTensorPipeline()

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -148,9 +148,17 @@ void NGraphEncapsulateOp::CreateParallelExecutor(OpKernelConstruction* ctx,
   int graph_id{-1};
   OP_REQUIRES_OK(ctx, ctx->GetAttr("ngraph_graph_id", &graph_id));
 
+  int my_function_cache_depth_in_items = 16;
```
shresthamalik: can change 16 to a const, you can use something like this, which is property of cache
https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_prefetch_shared_data.h#L60

test/test_parallel_executor.cpp
vishakha-nervana: There were seperate tests for GetTensorFromPipeline(), which we dont need anymore as this function is merged into GetExecutableFunctionand Tensors() About multi-threading tests, since we are testing multithreading  for data cache, there are suffiecient tests there
We removed locks from executor so not much to testsbBut if anyone has more ideas please let me know

ngraph_bridge/ngraph_data_cache.h
```diff
@@ -53,13 +53,14 @@ class NgraphDataCache {
   // it will create item, put it in the cache and returns item and status
   std::pair<Status, ValueType> LookUpOrCreate(
       KeyType key,
-      std::function<std::pair<Status, ValueType>()> callback_create_item,
-      std::function<void(ValueType)> callback_destroy_item);
+      std::function<std::pair<Status, ValueType>(KeyType)> callback_create_item,
+      std::function<void(ValueType)> callback_destroy_item, bool& cache_hit);
```
sayantan-nervana: note-to-self: `callback_create_item` is a function that now accepts a `KeyType` (instead of void)

ngraph_bridge/ngraph_encapsulate_op.cc
sayantan-nervana: `my`->`m`

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -466,9 +470,7 @@ void NGraphEncapsulateOp::ComputeUsingParallelExecutor(OpKernelContext* ctx) {
       // Get the set of IO tensors for the next iteration
       std::tuple<int, PipelinedTensorVector, PipelinedTensorVector>
           io_tensors_next_iter;
-      OP_REQUIRES_OK(ctx, m_parallel_executor->GetTensorsFromPipeline(
-                              ng_exec, io_tensors_next_iter));
-
+      io_tensors_next_iter = pipelined_tensor_store->get_tensors();
```
sayantan-nervana: there is a bunch of get<0>, get<1> etc in the code.
Lets use `struct IoTensorBundle` or std::tie to extract the elements of the tuple (or maybe even change the code of `pipelined_tensor_store->get_tensors` to return `IoTensorBundle`, but that may be beyond the scope of this PR)

ngraph_bridge/ngraph_executor.cc
```diff
@@ -240,190 +222,164 @@ Status NGraphExecutor::GetNgExecutable(
...
+                op_backend);
+  // Get NgItems i.e. ng_executable, serialized ng_functions from Data Cache
+  auto status_ng_item_pair =
+      m_ng_data_cache.LookUpOrCreate(signature, create_ng_items_callback,
+                                     destroy_ng_items_callback, cache_hit);
+
+  if (status_ng_item_pair.first == Status::OK()) {
+    ng_exec = std::get<0>(status_ng_item_pair.second);
+    serialized_ng_func = std::get<1>(status_ng_item_pair.second);
+    pts = std::get<2>(status_ng_item_pair.second);
```
sayantan-nervana: maybe use an `std::tie` to do it in 1 line

test/test_parallel_executor.cpp
sayantan-nervana: Looks good, we can remove this I think

But refer to my comment above about an end-to-end test with 3 threads that is expected to fail

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -413,31 +421,27 @@ void NGraphEncapsulateOp::ComputeUsingParallelExecutor(OpKernelContext* ctx) {
...
-  OP_REQUIRES_OK(ctx, m_parallel_executor->GetNgExecutable(tf_input_tensors,
-                                                           ng_exec, cache_hit));
-
+  std::string serialized_ng_function;
+  OP_REQUIRES_OK(ctx, m_parallel_executor->GetExecutableFunctionAndTensors(
+                          tf_input_tensors, ng_exec, serialized_ng_function,
+                          pipelined_tensor_store, cache_hit));
+  io_tensors = pipelined_tensor_store->get_tensors();
+  OP_REQUIRES(ctx, !(std::get<0>(io_tensors) < 0),
+              errors::Internal("No free tensor available"));
```
sayantan-nervana: This is an important behaviour. Do we have a test for this? A test would be:
start running a graph with 3 threads, the first 2 threads will work, the last one is expected to throw an error.
so the end-to-end test would expect an error out of one of the 3 threads.

test/test_parallel_executor.cpp
```diff
@@ -340,248 +282,6 @@ TEST(ParallelExecutor, ExecuteOnSingleThread8Bit) {
   // Compare(tf_output_tensor, expected_val, 0);
 }
 
-TEST(ParallelExecutor, ExecuteOnMultipleThreads8Bit) {
```
sayantan-nervana: These tests needn't be removed right. They will look different under the new circumstances, but they should be essentially testing the things they were testing earlier since these tests are kind of end-to-end tests

sayantan-nervana: You could do this (or the next) test with 1, 2 and 3 threads. the 3 threads one is expected to fail (of course we may have to barrier it so that we ensure that all 3 threads are running at the same time so that the PTS is empty for atleast one of them.

ngraph_bridge/ngraph_encapsulate_op.cc
sayantan-nervana: Please file a ticket for this

test/test_parallel_executor.cpp
vishakha-nervana: Add tests

PR #305, opened 2019-10-31
Sarkars/add training resume test

sayantan-nervana:
The examples we have completes a training and dumps saved models. But afaik, we do not have a test where we resume training from a semi-trained model. We know it works, but its best to assert it in a small scale E2E test

1. Modifications in mnist_deep_simplified.py
Added an option `input_model_dir` which allows it to load semi-trained weights to initialize with
#Sample code for resuming training:
#python mnist_deep_simplified.py --output_model_dir ./save_loc/train --train_loop_count 50
#python mnist_deep_simplified.py --output_model_dir ./save_loc1/train --train_loop_count 50 --input_model_dir ./save_loc

2. Extending the current pytest:
What the test originally did
#In this test, we first run 10 iterations on ngraph and dump the semi trained model
#Then we run TF for 10 iterations and compare the values with ngraph
What is added in this PR
#Then we load back the semitrained nodels in both cases and train them for 10 iterations
#and again make sure TF and NG agree.


test/python/test_mnist_training.py
```diff
@@ -38,31 +40,64 @@ class TestMnistTraining(NgraphTest):
 
     @pytest.mark.parametrize(("optimizer"), ("adam", "sgd", "momentum"))
     def test_mnist_training(self, optimizer):
+        # In this test, we first run 20 iterations on ngraph and dump the semi trained model
+        # Then we run TF for 20 iterations and compare the values with ngraph
+        # Then we load back teh semitrained nodels in both cases and train them for 20 iterations
```
sayantan-nervana:
```suggestion
        # Then we load back the semitrained models in both cases and train them for 10 iterations
```

test/python/test_mnist_training.py
```diff
@@ -38,31 +40,64 @@ class TestMnistTraining(NgraphTest):
...
-                                     make_deterministic, optimizer)
+        # compare values
+        assert np.allclose(
+            ng_values, tf_values,
+            atol=1e-3), "Loss or Accuracy values don't match"
 
+        tf.reset_default_graph()
+
+        # Now resume training from output_model_dir
+        # and dump new model in output_model_dir_ng_2
```
shresthamalik: enable ngraph here. it was disabled for the last test

test/python/test_mnist_training.py
```diff
@@ -38,31 +40,64 @@ class TestMnistTraining(NgraphTest):
...
+        output_model_dir_ng_2 = './save_loc_ng_2_' + optimizer + '/'
```
shresthamalik: we need to enable ngraph before this

test/python/test_mnist_training.py
```diff
@@ -38,31 +40,64 @@ class TestMnistTraining(NgraphTest):
...
+        FLAGS = mnist_training_flags(data_dir, output_model_dir_ng_2,
+                                     train_loop_count, batch_size,
+                                     test_image_count, make_deterministic,
+                                     optimizer, output_model_dir_ng_1)
```
shresthamalik: I was thinking if we can break it into two tests, but i see there is a dependency on the checkpoint being dumped

adk9:
Thanks for the PR, but we're closing this because it's no longer needed; training support is being removed from the bridge.

PR #306, opened 2019-10-31
Add MNIST example that uses Keras

sayantan-nervana:
Adding an example showing how to enable ngraph_bridge for keras models


adk9:
@googlebot I fixed it.

adk9:
Thanks for the PR. We're closing this because it fails the CLA. Please reopen if necessary after signing the CLA.

PR #307, opened 2019-11-01
Sindhu/frozen models

sindhu-nervana:


diagnostics/model_test/verify_model.py
```diff
@@ -312,8 +312,8 @@ def read_inputs_from_file(param_dict):
     assert (len(tf_skipped_tensors) == len(ngraph_skipped_tensors)) and all(
         [i == j for i, j in zip(tf_skipped_tensors, ngraph_skipped_tensors)])
 
-    print("Skipping comparison of the output tensors below:")
```
sayantan-nervana: Now it will print:
```
Skipping comparison of the output tensor:
[import/a:0]
Skipping comparison of the output tensor:
[import/b:0]
```

maybe you wanted:
```
if len(tf_skipped_tensors) > 0:
    print("Skipping comparison of the output tensor:")
```

sindhu-nervana: yeah, i forgot to change this. I moved this around while running models so I won't see the print in non skipped cases.

PR #287, opened 2019-10-15
Kanvi/implement ng var update ng tensor op

kanvi-nervana:
This PR does the following:
1. Implements NGraphVariableUpdateNGTensor : which syncs the NG tensor with TF tensor
2. Adds a unit test : to test the compute function for newly implemented op

PR #309, opened 2019-11-05
Kanvi/use ng var update ng tensor op

kanvi-nervana:
This PR (based off of #287 ) does the following:

- Adds the NGraphVariableUpdateNGTensor custom op after a TF optimizer is found updating a variable

- Removes is_tf_just_looking attribute from all op registrations and the corresponding helper functions required to keep the ng-tensor and tf-tensor in sync since that will now be done by the custom op

- Adds a simple test to make sure the custom op gets attached at the right place

- Adds an end-to-end test as well making sure the sync op works as intended


test/graph_rewrites/test_ng_var_update_ng_tensor.cc
```diff
@@ -40,85 +41,125 @@ namespace tensorflow {
...
-  ng_t->read(dst_ptr, 0, output_tensor.TotalBytes());
+  ASSERT_OK(ng_session.Run({applyadagrad_t}, &ng_outputs2));
 
-  Compare(output_tensor, input_tf_tensor, 0);
+  DeactivateNGraph();
 
+  // Run on TF
+  ClientSession tf_session(root, options);
+  std::vector<tensorflow::Tensor> tf_outputs1;
+  std::vector<tensorflow::Tensor> tf_outputs2;
```
shresthamalik: The end to end test can be added to test_variables.cc or a new file outside this folder. We can use this for only testing the graph rewrites.

test/graph_rewrites/test_ng_var_update_ng_tensor.cc
```diff
@@ -40,85 +41,125 @@ namespace tensorflow {
...
+  g.AddEdge(source, Graph::kControlSlot, var_node, Graph::kControlSlot);
+  g.AddEdge(assign, Graph::kControlSlot, sink, Graph::kControlSlot);
+
+  ASSERT_OK(RewriteForTracking(&g, 0));
+
+  map<string, Node*> node_map;
+  for (auto node : g.op_nodes()) {
+    node_map[node->name()] = node;
+  }
+  ASSERT_EQ(node_map.find("var_node_sync_node")->second->type_string(),
```
shresthamalik: You can assert on the right connections of the edges as that is where complexity lies.

kanvi-nervana: good point

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -71,34 +69,75 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
-            is_tf_just_looking = false;
+            NGRAPH_VLOG(1) << DebugNode(edge->dst())
+                           << "needs reference from ( "
+                           << DebugNode(edge->src())
+                           << " ), adding a "
+                              "NGraphVariableUpdateNGTensor node here";
+
+            // If the number of outputs for the TF optimizer > 1
+            // we do not handle it and error out
+            if (node->num_outputs() > 1) {
```
shresthamalik: I think the TF Optimizer is the edge->dst() and not the node

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -71,34 +69,75 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
+            TF_RETURN_IF_ERROR(status);
+            sync_node->set_assigned_device_name(node->assigned_device_name());
+
+            // Connect output edges from the TF optimizer to the sync node
+            // which will be for the input index 0 since that is the
+            // output we want.
+            // This should not replace the control edges.
+            if (edge->dst_input() == 0) {
+              TF_RETURN_IF_ERROR(
+                  ReplaceOnlyDataOutputEdges(graph, edge->dst(), sync_node));
```
shresthamalik: I would think all the ctrl + data output edges outgoing of optimizer node need to be moved to the sync_node

test/graph_rewrites/test_ng_var_update_ng_tensor.cc
```diff
@@ -40,85 +41,125 @@ namespace tensorflow {
...
+  ClientSession ng_session(root, options);
+  std::vector<tensorflow::Tensor> ng_outputs1;
+  std::vector<tensorflow::Tensor> ng_outputs2;
+  ASSERT_OK(ng_session.Run({{var_assign, accum_assign}}, &ng_outputs1));
 
-  shared_ptr<ngraph::runtime::Tensor> ng_t = var->ng_tensor();
-  Tensor output_tensor(DT_FLOAT, TensorShape({2}));
-  void* dst_ptr = DMAHelper::base(&output_tensor);
-  ng_t->read(dst_ptr, 0, output_tensor.TotalBytes());
+  ASSERT_OK(ng_session.Run({applyadagrad_t}, &ng_outputs2));
```
shresthamalik: Run it in a loop for a few times.

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -71,34 +69,75 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
+            }
```
shresthamalik: I am not sure if TF will optimize out the sync nodes that are not connected to any other node. Are all the sync nodes (directly or indirectly) connected to the SINK node ?

test/graph_rewrites/test_ng_var_update_ng_tensor.cc
```diff
@@ -40,85 +41,125 @@ namespace tensorflow {
...
-  Compare(output_tensor, input_tf_tensor, 0);
+  DeactivateNGraph();
 
+  // Run on TF
+  ClientSession tf_session(root, options);
+  std::vector<tensorflow::Tensor> tf_outputs1;
+  std::vector<tensorflow::Tensor> tf_outputs2;
+  ASSERT_OK(tf_session.Run({{var_assign, accum_assign}}, &tf_outputs1));
+
+  ASSERT_OK(tf_session.Run({applyadagrad_t}, &tf_outputs2));
```
shresthamalik: We can think of two more tests where the output of applyadagrad goes to
1. NGraphEncap lets say (node that does not take as ref . For e.g. Add/mul)
2. Another optimizer we support (SGD/Assign ADD)
3. Another optimizer we do not support (RMSPRop or something)

kanvi-nervana: Added tests for the first 2 but the 3rd scenario where we have consecutive TF optimizers is a future optimization that is not currently supported.

test/test_variable_ops.cpp
```diff
@@ -502,6 +502,273 @@ TEST(VariableTest, SmallGraph6) {
...
+      ->mutable_rewrite_options()
+      ->set_constant_folding(tensorflow::RewriterConfig::OFF);
+
+  // Run on nGraph
+  ActivateNGraph();
+  ClientSession ng_session(root, options);
+  std::vector<tensorflow::Tensor> ng_outputs1;
+  std::vector<tensorflow::Tensor> ng_outputs2;
+
+  for (int i = 0; i < 10; i++) {
```
shresthamalik: we are assigning in a loop

test/test_variable_ops.cpp
```diff
@@ -502,6 +502,273 @@ TEST(VariableTest, SmallGraph6) {
...
+    ASSERT_OK(tf_session.Run({assign_add}, &tf_outputs3));
+  }
+
+  for (int i = 0; i < 10; i++) {
+    ASSERT_OK(tf_session.Run({var1}, &tf_outputs4));
+  }
+
+  Compare(tf_outputs1, ng_outputs1);
+  Compare(tf_outputs2, ng_outputs2);
```
shresthamalik: We don't need the string comparison coz of the sync optimizers. is that correct?

test/test_variable_ops.cpp
kanvi-nervana: This is a mistake

test/test_variable_ops.cpp
kanvi-nervana: This would need string comparison too for the initialization. I will update it now that the other PR that has that change is merged.

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -122,10 +125,65 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
+          // node to the output/s of the TF op.
+          if (IsRefType(node->output_type(0))) {
+            TF_RETURN_IF_ERROR(ReplaceOutputEdges(graph, node, sync_node));
+          } else {
+            for (auto out_edge : node->out_edges()) {
+              graph->AddEdge(sync_node, Graph::kControlSlot, out_edge->dst(),
+                             Graph::kControlSlot);
+            }
+          }
+        }
```
sayantan-nervana: Lets consider the var+accumulator leading into a TFoptimizer. The accumulator's sync is topologically a "last" node. Will TF optimize it away?
do we need a ctrl edge from accumulator's sync to the sync with outgoing edges (variable's sync), so that sync-accumulator remains connected to the output fetch node

Probably something we can look into later.

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -122,10 +125,65 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
+
```
sayantan-nervana: Note to self:

in case of IsVariableInitialized:

`Var--(ref edge)-->IsVariableInitialized--(nonref edge)-->X`


This will become:

```Var--(ref edge)-->IsVariableInitialized--(nonref edge)-->X
                                            |
                                            v
                                        sync
```

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -924,12 +916,6 @@ void NGraphEncapsulateOp::ComputeUsingLegacyExecutor(OpKernelContext* ctx) {
             ng_encap_impl_.SetNumberOfCopies(copies++);
             ng_encap_impl_.AppendCopyLog(" COPY_TO_TF ");
           }
-          if (!NGraphCatalog::GetIsTFJustLookingFromEncapOutputInfoMap(
-                  output_key)) {
-            // Some tf op might update the ng-tensor value so mark it stale
-            ng_encap_impl_.AppendCopyLog(" SET_SYNC ");
-            var->set_sync_ng_tensor(true);
-          }
```
sayantan-nervana: note to self: changes only in legacy executor. since parallel executor does not have varopts yet

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
shresthamalik: I had the same concern, but looks like the end to end tests are passing. Worth exploring later.

PR #310, opened 2019-11-06
sindhu/Implement Sin and Cos ops

sindhu-nervana:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -4991,9 +5017,10 @@ const static std::map<
       {"Rsqrt", TranslateRsqrtOp}, {"RsqrtGrad", TranslateRsqrtGradOp},
       {"Select", TranslateSelectOp}, {"Shape", TranslateShapeOp},
       {"Sigmoid", TranslateSigmoidOp}, {"SigmoidGrad", TranslateSigmoidGradOp},
-      {"Size", TranslateSizeOp}, {"Sign", TranslateUnaryOp<ngraph::op::Sign>},
-      {"Slice", TranslateSliceOp}, {"Snapshot", TranslateIdentityOp},
-      {"Softmax", TranslateSoftmaxOp}, {"Softplus", TranslateSoftplusOp},
+      {"Sin", TranslateSinOp}, {"Size", TranslateSizeOp},
```
sayantan-nervana: I think we can just call `TranslateUnaryOp` for these 2

test/test_math_ops.cpp
```diff
@@ -855,6 +855,27 @@ TEST(MathOps, Cast2D) {
...
 
+// Test op: Cos
+TEST(MathOps, Cos) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 4;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2}));
+
+  AssignInputValues(A, 3.8f);
```
sayantan-nervana: This is testing for all values = 3.8f. For mathematical functions we should probe it at different points that might fail.
I'd suggest use `void AssignInputValues(Tensor& A, T x)` https://github.com/tensorflow/ngraph-bridge/blob/sarkars/resize_bilinear/test/test_utilities.h#L84

and test for `{0, -0, pi/2, pi, 1.0, 3.8, 4.2, -3.9, ...(some other float values)... , -1.0, std::numeric_limits<float>::infinity(), -std::numeric_limits<float>::infinity(), std::numeric_limits<float>::quiet_NaN(), std::numeric_limits<float>::signaling_NaN()}`

sindhu-nervana: infinity does not work, it fails while comparing TF and ngraph values. i updated the inputs with the other values.

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -566,6 +569,7 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
       type_constraint_map["Sigmoid"]["T"] = NGraphNumericDTypes();
       type_constraint_map["SigmoidGrad"]["T"] = NGraphNumericDTypes();
       type_constraint_map["Sign"]["T"] = NGraphNumericDTypes();
+      type_constraint_map["Sin"]["T"] = NGraphNumericDTypes();
```
sayantan-nervana:
```suggestion
      type_constraint_map["Sin"]["T"] = NGraphRealDTypes();
```

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -443,6 +445,7 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
       type_constraint_map["Conv2D"]["T"] = NGraphNumericDTypes();
       type_constraint_map["Conv2DBackpropInput"]["T"] = NGraphNumericDTypes();
       type_constraint_map["Conv3D"]["T"] = NGraphNumericDTypes();
+      type_constraint_map["Cos"]["T"] = NGraphNumericDTypes();
```
sayantan-nervana:
```suggestion
      type_constraint_map["Cos"]["T"] = NGraphRealDTypes();
```

PR #318, opened 2019-11-08
Sarkars/fix race

sayantan-nervana:
In general, Set is called before any call to Get.

But in case a Get is called before Set, it will return with incorrect value (0 or -1). So when Get is called it should wait till a Set is called and then Get can return.


avijit-chakroborty:
This is a good find and great fix. This will solve the race condition. 

### ⚠️ Issues with the current design

On a separate note,  reviewing the design a bit during my truly `free` time, I came to the conclusion that the double buffering design has a serious flaw. The main issue is that the signal to start writing to the device is provided by the NGEncapsuateOp but it should be the opposite i.e., started by the **`NGraphPrefetchDatasetOp`**. The reasons are the following:

- A data pipeline may not even contain `Prefetch`. The NGEncapsulateOp won't know this. So in that case no one will call the `SetBufferDepth()` but NGEncapsulateOp will call `GetBufferDepth()` and the execution will block. 
    - This can be easily tested right now - just remove the `prefetch` from [this line](https://github.com/tensorflow/ngraph-bridge/blob/a14c0b063798f26572c85e3d105ae86d4b4a32ee/examples/axpy_pipelined.py#L45) and watch the execution stall
- For a real network, there are multiple nGraph clusters and each executes by a different NGEncapsulateOp instances. In the current design - each instances of the NGEncapsulateOp will create the shared data `NGraphPrefetchSharedResouce` with the same resource name. During the prefetch operation, the single `NGraphPrefetchOp` will pull the resource object out and try to write to the nGraph backend device but will end up using the wrong device tensor or even segfault due to other reasons.  
   - This can be observed by running this [simple example](https://github.com/cs230-stanford/cs230-code-examples/blob/master/tensorflow/vision/evaluate.py) with nGraph and setting the NGRAPH_TF_BACKEND=INTERPRETER and NGRAPH_TF_USE_PREFETCH=1 
   - Note:This is the evaluation example. Before running the evaluation - run the training without nGraph by following the instructions in the README here: https://github.com/cs230-stanford/cs230-code-examples/tree/master/tensorflow/vision

🔥 **So - is everything broken now?** 
No. Since the implementation is guarded by `NGRAPH_TF_USE_PREFETCH` environment variable, unless this is set, nothing will break i.e., all the existing features/tests will work as is.

### 🔨 Proposed design change

The basic building blocks are all good - it's just their usage will have to be readdressed. The general design change would be the following:

- Modify the **`NGraphPrefetchSharedResouce`** to add two new boolean flags 
   - *`m_prefetch_ready`* - when set will indicate that writing to device tensor during prefetch can start. 
   - *`m_write_to_device`* - when set the **`NGraphPrefetchDatasetOp`** will write to device tensor
- Let the **`NGraphPrefetchDatasetOp`** create the **`NGraphPrefetchSharedResouce`** object first and set the data member **`m_prefetch_ready`**
   - Each time the input is perfected, the **`NGraphPrefetchDatasetOp`** will check whether the **`m_write_to_device`** is set and write to device using the device tensors provided 
- The **`NGraphEncapsulateOp`** will check whether the **`NGraphPrefetchSharedResouce`** exists. 
   - If not - proceed as usual.
   - If available (that means the data prefetching is being used and the prefetched thread is ready to write the device tensor) - do the following:
       - Determines which inputs are needed to be pipelined. Note that depending on the computation graph - the index of the input (i.e., not the variables) could be anywhere in the input set. See the encapsulated graph produced by the `Hand sign recognition` example mentioned above. 
       - Copy the input tensor (that is being fetched by the prefetcher) to the **`NGraphPrefetchSharedResouce`** object. The actual procedure is same as what's currently done.
       - Also, for future iterations, see if the device tensor is ready (i.e., copied by the pre-fetcher) and skip writing to this device tensor when writing other device tensors (such as variables for the first time).

#### How to determine which input is from data pipeline

Among many other inputs to the **`NGraphEncapsulateOp`** only a few will be provided by the data pipeline as needed by the network. For example in the `Hand sign recognition` example above, only two tensors - image and label - are from the data pipeline. Rest of the input are training weights (variables that are updated during the training or read during evaluation). However, the input tensor list will contains all of them and therefore the actual tensors provided to the pre-fetcher need to be identified during the **`Compute(...)`** call. 

This can be done in a few ways. For example, during the graph rewriting, the common Op between the two sub graphs can be identified and the specific edges may be noted down and passed as attributed to the encapsulate op of something. The key is the identification of which tensors are being fetch via the data pipeline.

Also, the design of pipelined tensors as provided by specific backend needs to be revisited. For example, the variables do not need to be pipelined as they would reside within the backend but the input for training and inference will need to be pipelined - but the current backend design doesn't have this distinction. Or does it?

Finally, the implementation needs to be tested against a data pipeline that **doesn't** use a **`prefetch`** node. The rest should be identical except that the timing plots will show the device copy by the same thread during the iteration as opposed to from a different thread. 

@shresthamalik and @sayantan-nervana - these are just some thoughts that crossed my mind. Hope they help to perfect this design 😄

ngraph_bridge/ngraph_prefetch_shared_data.h
```diff
@@ -92,8 +92,26 @@ class NGraphPrefetchSharedResouce : public ResourceBase {
...
+    // TODO assert m_prefetch_buffer_depth == -1 || m_prefetch_buffer_depth ==
+    // depth
+    // To make sure we never try to set a different depth once it is set
+    m_prefetch_buffer_depth = depth;
+    m_cv.SignalAll();
+    m_mutex.Unlock();
+  }
+  int GetBufferDepth() {
+    // Locking GetBufferDepth till SetBufferDepth is called
+    // In case of races where Get is called before Set,
```
vishakha-nervana: just curious, will making this variable m_prefetch_buffer_depth
atomic work?

PR #320, opened 2019-11-08
Sindhu/more ops

sayantan-nervana:
scatternd op

Also upgrades ngcore version as it includes a bug fix and cropandresize


test/test_array_ops.cpp
```diff
@@ -902,6 +902,75 @@ TEST(ArrayOps, Rank) {
...
+  OpExecuter opexecuter(root, "ScatterNd", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}  // end of test op ScatterNd
+
+// TODO: add unit test for duplicate indices here and 3D,4D
+// these are done
+// TODO any other test? something broadcasty?
+
```
sindhu-nervana: 2d to 3d? like here? https://github.com/NervanaSystems/ngraph/blob/5465677fb1abd2e8a78d4544ff2d248a7ee3ea17/test/backend/scatter.in.cpp#L255

sayantan-nervana: Tried adding those tests, but they do not seem to be right shapes for TF.

test/python/tensorflow/python_tests_list_mac.txt
```diff
@@ -397,6 +397,74 @@ relu_op_test.ReluTest.testGradientFloat32
 relu_op_test.ReluTest.testGradientFloat64
 relu_op_test.ReluTest.testGradientScalar
 
+scatter_nd_ops_test.ScatterNdNonAliasingAddTest.testBool
```
kanvi-nervana: have these been run on mac?

test/python/tensorflow/python_tests_list_gpu.txt
```diff
@@ -389,6 +389,73 @@ relu_op_test.ReluTest.testGradientFloat32
 relu_op_test.ReluTest.testGradientScalar
 relu_op_test.ReluTest.testNumbers
 
+scatter_nd_ops_test.ScatterNdNonAliasingAddTest.testBool
```
kanvi-nervana: Have these been run on GPU?

sayantan-nervana: I thought of removing them (since they are removed on CPU), but kept them there since there is no CI for GPU. If/when GPU ci is run, I suspect some other tests will fail too, and we can comment them out then.

For mac, CI will run now, and if it fails (which I think it will) I'll remove it.

test/python/tensorflow/python_tests_list.txt
```diff
@@ -397,6 +397,76 @@ relu_op_test.ReluTest.testGradientFloat32
...
+scatter_nd_ops_test.StatefulScatterNdTest.testRank3InvalidShape2
+scatter_nd_ops_test.StatefulScatterNdTest.testRank3ValidShape
+scatter_nd_ops_test.StatefulScatterNdTest.testResVarInvalidOutputShape
+scatter_nd_ops_test.StatefulScatterNdTest.testScatterOutOfRangeCpu
+scatter_nd_ops_test.StatefulScatterNdTest.testScatterRepeatIndices
+scatter_nd_ops_test.StatefulScatterNdTest.testSimple
+scatter_nd_ops_test.StatefulScatterNdTest.testSimple2
+scatter_nd_ops_test.StatefulScatterNdTest.testSimple3
+scatter_nd_ops_test.StatefulScatterNdTest.testSimpleResource
+scatter_nd_ops_test.StatefulScatterNdTest.testVariableRankAdd
```
shresthamalik: these tests were not run before?

test/test_array_ops.cpp
```diff
@@ -902,6 +902,95 @@ TEST(ArrayOps, Rank) {
...
+  auto R = ops::ScatterNd(root, indices, updates, {10});
+  std::vector<Output> sess_run_fetchoutputs = {R};
+
+  OpExecuter opexecuter(root, "ScatterNd", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}
+
+// This test runs on TF but not on Ng for now
```
shresthamalik: how are we doing that. From the test it seems it is running on NG

test/python/tensorflow/python_tests_list_gpu.txt
sayantan-nervana: Removed from mac as they fail there

test/test_array_ops.cpp
sayantan-nervana: Stale comment. Removed it

test/python/tensorflow/python_tests_list.txt
sayantan-nervana: ScatterND is being added as a new op, so searched for TF python tests that might correspond to scatternd and added them here.

PR #323, opened 2019-11-09
Shrestha/add unit test mixed tensors

shresthamalik:
Added Unit Tests that test forming of ng_inputs and ng_outputs with few tensors created by backend and others by executable. 
 - (NGraphExecTest, MixedTensors)
 - (NGraphExecTest, MixedTensorsPipelined)

1. Renamed file graph_exec.cc to test_ngraph_exec.cc (in line with other files)
2. Created a test class NGraphExecTest to host some utility functions that are used by tests
3. Tests (NGraphExecTest, Axpy)  and (NGraphExecTest, Axpy8bit) have no changes except for using the utility functions at some places. The diff is mainly due to the new file.


test/test_ngraph_exec.cpp
```diff
@@ -0,0 +1,499 @@
...
+#include "tensorflow/core/graph/algorithm.h"
+#include "tensorflow/core/graph/graph.h"
+#include "tensorflow/core/graph/graph_constructor.h"
+#include "tensorflow/core/platform/env.h"
+
+#include "ngraph_bridge/ngraph_builder.h"
+#include "ngraph_bridge/ngraph_pipelined_tensors.h"
+#include "ngraph_bridge/ngraph_utils.h"
+
+#include "test_utilities.h"
```
kanvi-nervana: Change to `test/test_utilities.h`

test/test_ngraph_exec.cpp
```diff
@@ -0,0 +1,499 @@
...
+    DeactivateNGraph();
+    Status status = session->Run(feed_dict, out_node_names, {}, &out_tensors);
+    ActivateNGraph();
+    return status;
+  }
+};
+
+TEST_F(NGraphExecTest, Axpy) {
+  Graph input_graph(OpRegistry::Global());
+  ASSERT_OK(LoadGraph("test_axpy_launchop.pbtxt", &input_graph));
```
vishakha-nervana: I think there is a lot of common code between  these tests, and It can be moved to SetUp().

test/test_ngraph_exec.cpp
```diff
@@ -0,0 +1,499 @@
...
+    DumpNGTensor<int8>(cout, ng_function->get_output_op(i)->get_name(),
+                       outputs[i]);
+    cout << endl;
+  }
+  // Add the validation logic
+  // TODO
+}
+
+TEST_F(NGraphExecTest, MixedTensors) {
+  Graph input_graph(OpRegistry::Global());
```
vishakha-nervana: I am not sure if I understand what mixed tensors are, 
can you please add this to general description?

test/test_ngraph_exec.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+  // Run on nGraph
+  // Translate Graph
+  std::vector<TensorShape> tf_input_shapes(num_inputs, tf_shape);
+  shared_ptr<ng::Function> ng_function;
+  ASSERT_OK(
+      TranslateTFGraphNoStatic(tf_input_shapes, input_graph, ng_function));
+
+  // Create the nGraph backend
+  string backend_name = "INTERPRETER";
```
vishakha-nervana: I noticed that we are are testing a lot of our tests with Interpreter as backend,
do you think it would be good idea to get backend in user input, through environment variable 
or flag, and in case input not provided Interpreter will be default backend.

test/test_ngraph_exec.cpp
shresthamalik: Its already there. There are some tensors that are created using backend api and others using execuatable api, that's what i mean by mixed tensors

test/test_ngraph_exec.cpp
shresthamalik: OverrideBackendFromEnv(&backend_name) this does that

test/test_ngraph_exec.cpp
shresthamalik: I have not looked too much into the existing tests Axpy and Axpy8bit. Changing them are beyond the scope of this PR. It is showing as major change mainly due to renaming of the file

PR #329, opened 2019-11-15
Kanvi/upgrade ngraph to v0.26.1-rc0

kanvi-nervana:


kanvi-nervana:
Don't update branch from here

PR #330, opened 2019-11-15
sarkars/resize bilinear for r19

sayantan-nervana:
resizebilinear translation. uses ngraph op "interpolate", but that has no implementation in any backend, so gtest is disabled for now.

There are some TODOs in the PR, to be done when we have backends start supporting interpolate.

### DIFF
--- a/ngraph_bridge/ngraph_builder.cc
+++ b/ngraph_bridge/ngraph_builder.cc
@@ -26,6 +26,7 @@
 #include "ngraph/builder/quantization.hpp"
 #include "ngraph/op/argmax.hpp"
 #include "ngraph/op/argmin.hpp"
+#include "ngraph/op/experimental/layers/interpolate.hpp"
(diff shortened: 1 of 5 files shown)


ngraph_bridge/ngraph_builder.cc
```diff
@@ -3696,6 +3697,33 @@ static Status TranslateReshapeOp(
...
+  shared_ptr<ng::Node> images, size;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &images, &size));
+
+  bool align_corners;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "align_corners", &align_corners));
+
+  ngraph::op::InterpolateAttrs attrs;
+  attrs.align_corners = align_corners;
+  attrs.mode = "linear";
+  attrs.antialias = false;
```
kanvi-nervana: Is it always supposed to be false?

PR #336, opened 2019-11-19
Sindhu/op by op capability framework

sindhu-nervana:
1. Added a mapping of {TF OP, list of nGraph ops}
2. For each TF op, loop through the corresponding nGraph Ops and check if Op is supported by a backend or not
3. If it is not supported in #2 above, then do not mark for clustering 
4. Add unit tests to test the functionality


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -136,6 +138,43 @@ static ConfirmationFunction SimpleConfirmationFunction() {
...
+  if (op_name == "Divide")
+    ng_dummy_node = std::make_shared<ngraph::op::Divide>();
+  if (op_name == "Subtract")
+    ng_dummy_node = std::make_shared<ngraph::op::Subtract>();
+  if (op_name == "Multiply")
+    ng_dummy_node = std::make_shared<ngraph::op::Multiply>();
+  else
+    cout << "Cannot create a dummy ngraph node for this given op: " << op_name
+         << std::endl;
+  return ng_dummy_node;
```
sindhu-nervana: TODO: For testing purpose, should we move this map to header? To be discussed

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -192,6 +231,14 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
 
   static std::map<std::string, SetAttributesFunction> set_attributes_map;
 
+  // Map:: TF ops to NG Ops to track if all the Ngraph ops
+  // are supported by backend
+  // Update this Map if a new TF Op translation is
+  // implemented or a new Ngraph Op has been added
+  std::map<std::string, std::vector<std::string>> TFtoNgraphOpMap{
+      {"FloorMod", {"Floor", "Divide", "Subtract", "Multiply"}},
```
sindhu-nervana: TODO: Change this map to ```std::map<std::string, std::vector<ng::node>> TFtoNgraphOpMap{
      {"FloorMod", {ng::op::Floor, etc, etc, etc}}```

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -136,6 +138,30 @@ static ConfirmationFunction SimpleConfirmationFunction() {
   return cf;
 };
 
+// Check if op is supported by backend using is_supported API
+static Status IsSupportedByBackend(
+    const Node* node,
+    std::map<std::string, std::vector<shared_ptr<ng::Node>>>& TFtoNgraphOpMap,
+    const string& backend_name, bool& is_supported) {
```
sayantan-nervana: as discussed we could pass in an actual backend, rather than a string name. That will help in writing tests

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -765,6 +810,19 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
...
+      // check if op is supported by backend
+      bool is_supported = false;
+
+      // Create nGraph backend
+      string ng_backend_type;
+      BackendManager::GetCurrentlySetBackendName(&ng_backend_type);
+      TF_RETURN_IF_ERROR(IsSupportedByBackend(node, TFtoNgraphOpMap,
+                                              ng_backend_type, is_supported));
+      if (!is_supported) {
+        NGRAPH_VLOG(5) << "TF Op is not supported by backend:" << node->name();
```
sayantan-nervana: print name of backend, op type (and op name which is currently being printed)

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -765,6 +1049,26 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
...
       }
 
+      // Check if op is supported by backend
+      bool is_supported = false;
+
+      // Create nGraph backend
+      string ng_backend_type;
+      BackendManager::GetCurrentlySetBackendName(&ng_backend_type);
+      // Create backend to query
+      Status status = BackendManager::CreateBackend(ng_backend_type);
```
shresthamalik: Wrap CreateBackend with TF_RETURN_IF_ERROR

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -136,6 +138,25 @@ static ConfirmationFunction SimpleConfirmationFunction() {
...
+    bool& is_supported) {
+  is_supported = true;
+
+  auto ng_op = TFtoNgraphOpMap[node->type_string()];
+  // Loop through the ngraph op list to query
+  for (auto it = ng_op.begin(); it != ng_op.end(); it++) {
+    // Pass ngraph node to check if backend supports this op
+    auto ret = op_backend->is_supported(**it);
+    if (!ret) {
+      is_supported = false;
```
shresthamalik: you can return here, dont need to loop over the rest

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -192,6 +213,269 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
...
+  // node
+  // Commented few other ops like GetOutputElement, Or Relubackprop until
+  // default constructors are added
+  auto constant = ngraph::op::Constant::create(ngraph::element::f32,
+                                               ngraph::Shape{}, {2.0f});
+  // Map:: TF ops to NG Ops to track if all the Ngraph ops
+  // are supported by backend
+  // Update this Map if a new TF Op translation is
+  // implemented or a new Ngraph Op has been added
+  std::map<std::string, std::set<shared_ptr<ng::Node>>> TFtoNgraphOpMap{
```
shresthamalik: make this map static

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -136,6 +138,25 @@ static ConfirmationFunction SimpleConfirmationFunction() {
...
 };
 
+// Check if op is supported by backend using is_supported API
+static Status IsSupportedByBackend(
+    const Node* node, const ng::runtime::Backend* op_backend,
+    std::map<std::string, std::set<shared_ptr<ng::Node>>>& TFtoNgraphOpMap,
+    bool& is_supported) {
+  is_supported = true;
+
+  auto ng_op = TFtoNgraphOpMap[node->type_string()];
```
sayantan-nervana: Check if item exists in the map, else return bad status

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -765,6 +1049,26 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
...
+      // Create nGraph backend
+      string ng_backend_type;
+      BackendManager::GetCurrentlySetBackendName(&ng_backend_type);
+      // Create backend to query
+      Status status = BackendManager::CreateBackend(ng_backend_type);
+      ng::runtime::Backend* op_backend =
+          BackendManager::GetBackend(ng_backend_type);
+      TF_RETURN_IF_ERROR(IsSupportedByBackend(node, op_backend, TFtoNgraphOpMap,
+                                              is_supported));
+      BackendManager::ReleaseBackend(ng_backend_type);
```
sayantan-nervana: Maybe create backend once and check for all nodes and then release, instead of creating-releasing for each node

ngraph_bridge/ngraph_mark_for_clustering.cc
shresthamalik: using map.at() preferred instead of []

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -136,6 +138,25 @@ static ConfirmationFunction SimpleConfirmationFunction() {
...
+  auto ng_op = TFtoNgraphOpMap[node->type_string()];
+  // Loop through the ngraph op list to query
+  for (auto it = ng_op.begin(); it != ng_op.end(); it++) {
+    // Pass ngraph node to check if backend supports this op
+    auto ret = op_backend->is_supported(**it);
+    if (!ret) {
+      is_supported = false;
+      return Status::OK();
+    }
+  }
```
shresthamalik: need to still return status outside

ngraph_bridge/ngraph_mark_for_clustering.cc
sindhu-nervana: > using map.at() preferred instead of []

why?

shresthamalik: [] adds a new key to the map if it does not exist
.at() checks if the key exists otherwise throws an error
http://www.cplusplus.com/reference/map/map/

sindhu-nervana: thanks, will make the change

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -136,6 +138,26 @@ static ConfirmationFunction SimpleConfirmationFunction() {
...
 };
 
+// Check if op is supported by backend using is_supported API
+static Status IsSupportedByBackend(
+    const Node* node, const ng::runtime::Backend* op_backend,
+    std::map<std::string, std::set<shared_ptr<ng::Node>>>& TFtoNgraphOpMap,
+    bool& is_supported) {
+  is_supported = true;
+
+  auto ng_op = TFtoNgraphOpMap.at(node->type_string());
```
sayantan-nervana: Use find to locate if `node->type_string()` is in the map. if not return a bad status.

otherwise the `at` could fail ungracefully, and then a TF Status would not be bubbled up well.

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -765,6 +1128,27 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
         break;
       }
 
+      // Check if op is supported by backend
+      bool is_supported = false;
+
+      // Create nGraph backend
+      BackendManager::GetCurrentlySetBackendName(&ng_backend_type);
+      if (ng_backend_type != "NOP") {
```
sayantan-nervana: Need to check if we can avoid backend specific ifs
We may want to modify the unit test. NOP backend indeed does not support "Mul" therefore you are getting `TF Op Mul of type Mul is not supported by backend: NOP`

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -765,6 +1128,27 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
...
+        // Create backend to query
+        TF_RETURN_IF_ERROR(BackendManager::CreateBackend(ng_backend_type));
+        ng::runtime::Backend* op_backend =
+            BackendManager::GetBackend(ng_backend_type);
+        TF_RETURN_IF_ERROR(IsSupportedByBackend(node, op_backend,
+                                                TFtoNgraphOpMap, is_supported));
+
```
shresthamalik: Release backend is missing

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -781,6 +1165,11 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
     }
   }
 
+  // Release backend created to query is_supported
+  if (ng_backend_type != "NOP") {
```
shresthamalik: why is it here, is the backend used somewhere else

sindhu-nervana: no, instead of doing it for each node doing it for all the nodes and then releasing outside the for loop.

ngraph_bridge/ngraph_mark_for_clustering.cc
sindhu-nervana: yeah, that is the reason i added condition. @shresthamalik @kanvi-nervana - can we modify this unit test? If so, how? I can make the change if it is minor

sindhu-nervana: Also, NOP indeed does not have `is_supported` API

ngraph_bridge/ngraph_mark_for_clustering.cc
shresthamalik: you can create outside the for loop too, it is confusing since you are creating inside, but releasing outside

sindhu-nervana: ya, could do that too.

test/graph_rewrites/op_by_op_capability_test.cc
```diff
@@ -0,0 +1,173 @@
...
+// CPU, INTERPRETER should supports these ops, NOP does not.
+
+TEST(OpByOpCapability, Backend1) {
+  // Create Graph
+  Scope root = Scope::NewRootScope();
+  auto A = ops::Const(root.WithOpName("A"), {3.f, 2.f});
+  auto B = ops::Const(root.WithOpName("B"), {3.f, 2.f});
+  auto Add = ops::Add(root.WithOpName("Add"), A, B);
+  auto C = ops::Const(root.WithOpName("C"), {3.f, 2.f});
+  auto Mul = ops::Mul(root.WithOpName("Mul"), Add, C);
```
sayantan-nervana: We do not need these TF ops I think.

test/graph_rewrites/config_for_grappler_test.cc
```diff
@@ -276,7 +276,7 @@ TEST(GrapplerConfig, RConfig3) {
   // Even though the backend is set via config-writer, the one specified
   // by the env. var takes effect. So though we set this to CPU
   // the backend should point to NOP as set via env. var
-  ASSERT_EQ(ng_backend, "NOP");
+  ASSERT_EQ(ng_backend, "CPU");
```
shresthamalik: why did we have to change this?

test/graph_rewrites/op_by_op_capability_test.cc
```diff
@@ -0,0 +1,101 @@
...
+  // Map with all the backends, and what the boolean is_supported should be
+  std::map<std::string, bool> backend_map{
+      {"CPU", true}, {"INTERPRETER", true}, {"NOP", false}};
+
+  // Tests three cases of (Backend, is_supported=true/false)
+  // 1.CPU,true
+  // 2.INTERPRETER,true
+  // 3.NOP,false
+  for (auto it = backend_map.begin(); it != backend_map.end(); it++) {
+    ASSERT_OK(BackendManager::SetBackendName(it->first));
```
shresthamalik: this will set the backends for all the tests as NOP, when run in the suite.

Need to store, restore the backend. Do BackendManager::GetCurrentlySetBackendName before starting the test and then set it back BackendManager::SetBackendName

sindhu-nervana: thanks for pointing this out. made the change @shresthamalik

test/graph_rewrites/op_by_op_capability_test.cc
```diff
@@ -0,0 +1,102 @@
...
+  ASSERT_OK(BackendManager::GetCurrentlySetBackendName(&ng_backend_type));
+
+  // Map with all the backends, and what the boolean is_supported should be
+  std::map<std::string, bool> backend_map{
+      {"CPU", true}, {"INTERPRETER", true}, {"NOP", false}};
+
+  // Tests three cases of (Backend, is_supported=true/false)
+  // 1.CPU,true
+  // 2.INTERPRETER,true
+  // 3.NOP,false
```
sayantan-nervana:
```suggestion
  // 3. NOP,false
```

test/graph_rewrites/op_by_op_capability_test.cc
```diff
@@ -0,0 +1,102 @@
...
+  string ng_backend_type;
+  ASSERT_OK(BackendManager::GetCurrentlySetBackendName(&ng_backend_type));
+
+  // Map with all the backends, and what the boolean is_supported should be
+  std::map<std::string, bool> backend_map{
+      {"CPU", true}, {"INTERPRETER", true}, {"NOP", false}};
+
+  // Tests three cases of (Backend, is_supported=true/false)
+  // 1.CPU,true
+  // 2.INTERPRETER,true
```
sayantan-nervana:
```suggestion
  // 2. INTERPRETER, true
```

test/graph_rewrites/op_by_op_capability_test.cc
```diff
@@ -0,0 +1,102 @@
...
+  bool is_supported;
+  string ng_backend_type;
+  ASSERT_OK(BackendManager::GetCurrentlySetBackendName(&ng_backend_type));
+
+  // Map with all the backends, and what the boolean is_supported should be
+  std::map<std::string, bool> backend_map{
+      {"CPU", true}, {"INTERPRETER", true}, {"NOP", false}};
+
+  // Tests three cases of (Backend, is_supported=true/false)
+  // 1.CPU,true
```
sayantan-nervana:
```suggestion
  // 1. CPU, true
```

test/graph_rewrites/op_by_op_capability_test.cc
```diff
@@ -0,0 +1,102 @@
...
+  ASSERT_OK(BackendManager::GetCurrentlySetBackendName(&ng_backend_type));
+
+  // Map with all the backends, and what the boolean is_supported should be
+  std::map<std::string, bool> backend_map{
+      {"CPU", true}, {"INTERPRETER", true}, {"NOP", false}};
+
+  // Tests three cases of (Backend, is_supported=true/false)
+  // 1. CPU, true
+  // 2. INTERPRETER, true
+  // 3. NOP,false
```
sayantan-nervana:
```suggestion
  // 3. NOP, false
```

PR #344, opened 2019-11-21
Sarkars/make mem profiling optional

sayantan-nervana:


ngraph_bridge/ngraph_api.cc
```diff
@@ -127,6 +127,12 @@ void SetDisabledOps(set<string> disabled_ops_set) {
   disabled_op_types = disabled_ops_set;
 }
 
+bool IsProfilingMemory() {
```
kanvi-nervana: Suggestion: IsProfilingMemory --> IsMemoryProfilingEnabled

PR #349, opened 2019-11-25
MBubula/Added RandomUniformOp 

maxbubula:


test/python/test_random_uniform.py
```diff
@@ -0,0 +1,42 @@
...
+        shape = [10000]
+        data = tf.random.uniform(shape)
+        out = tf.math.reduce_sum(data, axis=0, keep_dims=True)
+
+        sess_fn = lambda sess: sess.run([out], feed_dict={})
+        assert np.allclose(
+            self.with_ngraph(sess_fn),
+            self.without_ngraph(sess_fn),
+            rtol=1.e-2,
+            atol=1.e-3)
```
sayantan-nervana: This test is going to run the computaion twice, and because of that it might produce 2 different random numbers in the 2 runs probably. Do we need to set some TF seed or something like that?

ngraph_bridge/ngraph_builder.cc
```diff
@@ -3186,6 +3186,36 @@ static Status TranslateRankOp(const Node* op, const std::vector<const Tensor*>&,
   return Status::OK();
 }
 
+static Status TranslateRandomUniformOp(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  // std::cout<<"--------------- begin\n";
```
sayantan-nervana: stray comment

ngraph_bridge/ngraph_builder.cc
```diff
@@ -3186,6 +3186,36 @@ static Status TranslateRankOp(const Node* op, const std::vector<const Tensor*>&,
...
+      op->name(), ng::element::i64, ng::Shape{shape.size()}, shape);
+
+  auto const_use_fixed_seed = ConstructNgNode<ng::op::Constant>(
+      op->name(), ng::element::boolean, ng::Shape{}, std::vector<bool>{false});
+
+  auto random_uniform = ConstructNgNode<ng::op::RandomUniform>(
+      op->name(), const_min, const_max, const_shape, const_use_fixed_seed, 0);
+
+  SaveNgOp(ng_op_map, op->name(), random_uniform);
+  // std::cout<<"--------------- end\n";
```
sayantan-nervana: stray comment

ngraph_bridge/ngraph_builder.cc
```diff
@@ -3186,6 +3186,36 @@ static Status TranslateRankOp(const Node* op, const std::vector<const Tensor*>&,
...
+  TF_RETURN_IF_ERROR(GetStaticInputVector(op, 0, static_input_map, &shape));
+
+  auto const_shape = ConstructNgNode<ng::op::Constant>(
+      op->name(), ng::element::i64, ng::Shape{shape.size()}, shape);
+
+  auto const_use_fixed_seed = ConstructNgNode<ng::op::Constant>(
+      op->name(), ng::element::boolean, ng::Shape{}, std::vector<bool>{false});
+
+  auto random_uniform = ConstructNgNode<ng::op::RandomUniform>(
+      op->name(), const_min, const_max, const_shape, const_use_fixed_seed, 0);
```
sayantan-nervana: Maybe we need to read the Seed from the TF op and pass it to ngraph's `RandomUniform` instead of `0`.

Even then do we expect TF's random number and ngraph's random number to be same?

test/python/test_random_uniform.py
maxbubula: That is why I compute the sum over the random vector, but to be sure I added setting a seed.

ngraph_bridge/ngraph_builder.cc
maxbubula: Cleaned up.

ngraph_bridge/ngraph_builder.cc
maxbubula: Cleaned up.

ngraph_bridge/ngraph_builder.cc
maxbubula: Changed, initially I wasn't working on that op, but after deeper insight I found the seed in attrs, I think it suits the need well.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -3186,6 +3186,38 @@ static Status TranslateRankOp(const Node* op, const std::vector<const Tensor*>&,
...
+
+  auto const_use_fixed_seed = ConstructNgNode<ng::op::Constant>(
+      op->name(), ng::element::boolean, ng::Shape{}, std::vector<bool>{true});
+
+  tensorflow::int64 seed{};
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "seed", &seed));
+
+  auto random_uniform = ConstructNgNode<ng::op::RandomUniform>(
+      op->name(), const_min, const_max, const_shape, const_use_fixed_seed,
+      seed);
```
sayantan-nervana: @shresthamalik
I wonder if we should mark input0 as static and create a new `const_shape`. Maybe we can just read it as a node and pass it to `RandomUniform`'s constructor.

Of course, if the input node is not const, it will produce dynamic shape output. (but with static input marking it will be a graph split)
https://github.com/NervanaSystems/ngraph/blob/ab440246afb5278347f1cbc76f87fb7ec6968570/src/ngraph/op/experimental/random_uniform.cpp#L84

bani-intelaipg: @maxbubula and @shresthamalik : has this comment been resolved?

sayantan-nervana: yes. marking as resolved

PR #351, opened 2019-11-26
Shrestha/refactor pipeline

shresthamalik:
Mostly a refactoring PR.
- The parallel executor assumes that all inputs and outputs are pipelined. And if NGRAPH_TF_USE_PREFETCH=1 , all inputs and outputs are also prefetched.
This will be changed to some inputs and outputs are pipelined. Out of the pipelined inputs, some are prefetched.

Because of the current assumption, the pipelined_tensors and Prefetch::IOTensorBundle were all used interchangeably, making it hard to start building on top of it
The changes are mainly in the names of the data structures and some reordering of code.

- Changed Prefetch IO TensorBundle to have only input tensors and ID
- Added checks in the encapsulate op for pipeline depth used currently and the one being shared to prefetch
- Added Prefetch Capture to var build

Tested with examples/axpy_pipelined.py


ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -427,23 +425,37 @@ void NGraphEncapsulateOp::ComputeUsingParallelExecutor(OpKernelContext* ctx) {
...
+
+  NGRAPH_VLOG(2)
+      << "NGraphEncapsulateOp::Compute got ngraph executable for cluster id: "
+      << m_parallel_executor->GetNgraphClusterId();
+
+  event_get_ng_item.Stop();
+  ngraph::Event::write_trace(event_get_ng_item);
+
+  // Error check for pipelined tensors and pipeline depth
+  OP_REQUIRES(ctx, m_parallel_executor->GetTensorPipelineDepth() == 2,
```
vishakha-nervana: instead of calling m_parallel_executor->GetTensorPipelineDepth() 
two times, can you store its value in some variable and then use that.

ngraph_bridge/enable_variable_ops/ngraph_capture_variables.cc
```diff
@@ -105,6 +187,14 @@ Status CaptureVariables(Graph* graph, std::set<string> skip_these_nodes) {
     NGRAPH_VLOG(4) << "Removing: " << node->name();
     graph->RemoveNode(node);
   }
+
+  if (std::getenv(NGraphPrefetchSharedResouce::NGRAPH_TF_USE_PREFETCH) !=
+      nullptr) {
+    if (prefetch_node != nullptr) {
+      return ReplacePrefetch(graph, prefetch_node);
+    }
```
vishakha-nervana: Should we add a VLOG in the else  of if (prefetch_node != nullptr) {
saying "prefetch_node = nullptr", No prefetch op created.....
something like this, so atleast we know.

shresthamalik: Good idea. There is no else, and the else will be created only for VLOG. I can add VLOG saying replacing this prefetch op with NGraphPrefetch

vishakha-nervana: That works too.

ngraph_bridge/ngraph_encapsulate_op.cc
shresthamalik: not storing in a variable coz its used only in error checking, not used anywhere else

PR #354, opened 2019-11-27
Shrestha/tensor manager reads prefetched inputs

shresthamalik:
The prefetched inputs are available in catalog. 
- Tensor Manager reads the catalog to get prefetched input indexes also
- Added tests


ngraph_bridge/ngraph_tensor_manager.cc
```diff
@@ -63,6 +63,33 @@ void NGraphTensorManager::Initialize() {
       FindComplement(m_number_of_inputs, m_input_indexes_from_variables);
   m_pipelined_output_indexes =
       FindComplement(m_number_of_outputs, m_output_indexes_assigning_variable);
+
+  string node_prefetch_key =
```
kanvi-nervana: prefetch_node_key

ngraph_bridge/ngraph_tensor_manager.cc
```diff
@@ -63,6 +63,33 @@ void NGraphTensorManager::Initialize() {
...
+      NGraphCatalog::CreateNodeKey(m_ng_encap_graph_id, m_ng_encap_node_name);
+
+  if (NGraphCatalog::ExistsInPrefetchedInputIndexMap(node_prefetch_key)) {
+    auto prefetch_indexes =
+        NGraphCatalog::GetIndexesFromPrefetchedInputIndexMap(
```
kanvi-nervana: The NGraphCatalog::GetIndexesFromPrefetchedInputIndexMap function signature can be changed to accept the node key rather than creating itself to avoid mistakes and also we have the key already created.

ngraph_bridge/ngraph_tensor_manager.cc
```diff
@@ -63,6 +63,33 @@ void NGraphTensorManager::Initialize() {
       FindComplement(m_number_of_inputs, m_input_indexes_from_variables);
   m_pipelined_output_indexes =
       FindComplement(m_number_of_outputs, m_output_indexes_assigning_variable);
+
+  string node_prefetch_key =
+      NGraphCatalog::CreateNodeKey(m_ng_encap_graph_id, m_ng_encap_node_name);
```
shresthamalik: I would rather not have the user create the key. So I will add utility for NGraphCatalog::ExistsInPrefetchedInputIndexMap(graph_id, name)

kanvi-nervana: The utility already exists https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_catalog.h#L154

ngraph_bridge/ngraph_tensor_manager.cc
shresthamalik: changed this, not creating a key

PR #358, opened 2019-12-02
sarkars/ashoke/changing tracing mechanism

sayantan-nervana:
Gets rid of earlier style of tracing. Also some other profiling/vlog stuff

Now tracing is RAII and looks like:

```
{
     NG_TRACE("NGTF_Call", "");
    // heavy computation here
}
```


shresthamalik:
Is there a plan to pull it into master also?

sayantan-nervana:
> Is there a plan to pull it into master also?

The tracing mechanism, maybe yes. Removing logs, maybe no.

PR #360, opened 2019-12-02
Vishakh1/datapipeline test final

vishakha-nervana:
Hopefully this PR will pass for data pipeline and other tests, let see


test/python/common.py
```diff
@@ -109,7 +109,7 @@ def set_env_variable(self, env_var, env_var_val):
 
     # unset the env variable
     def unset_env_variable(self, env_var):
-        os.environ.pop(env_var, None)
+        os.unsetenv(env_var)
```
shresthamalik: Please change this to use unset too
https://github.com/tensorflow/ngraph-bridge/pull/360/files#diff-c3448a8c2134a492e0f08f2131d5552dR130

vishakha-nervana: I wish suggest commit would have worked on unchanged lines as well

vishakha-nervana: @shresthamalik Should I change here and other similar place too?https://github.com/tensorflow/ngraph-bridge/blob/931d732ab1d5cc3c8ff2cad31a83e23418bf66f4/test/python/common.py#L58

sayantan-nervana: So are we replacing the environ.pop with unsetvar in the couple of places it shows up on tests: https://github.com/tensorflow/ngraph-bridge/search?utf8=%E2%9C%93&q=os.environ.pop&type=

vishakha-nervana: No we are not doing it in this PR. It is outside the scope of this PR.

PR #367, opened 2019-12-05
Added support for Where Op for NNPI

jiminha:


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -610,6 +610,7 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
       type_constraint_map["Transpose"]["T"] = NGraphDTypes();
       type_constraint_map["Transpose"]["Tperm"] = NGraphIndexDTypes();
       type_constraint_map["Unpack"]["T"] = NGraphDTypes();
+      type_constraint_map["Where"]["T"] = NGraphDTypes();
```
sayantan-nervana: right now input type of "where" is bool only, right?
maybe we can say `type_constraint_map["Where"]["T"] = {DT_BOOL}`

jiminha: Yes, bool only, will change

sayantan-nervana:
> @jiminha Possible to add some C++ unit tests, and/or Py tests?

Where can only run on a specific backend, and not on CPU, and our CI is CPU based, hence the lack of UTs

PR #356, opened 2019-11-27
shrestha/upgrade ng-core to master

shresthamalik:
ugrade ngraph_version = "e543688981ea2083a260929a70420f5ba6984d95" 
 Nagy Mostafa [MLIR] Fused Ops dialect declaration (#3860)
  
- Fixed Parallel Executor to take in the node name
- PTS asserts out if depth>0 but no of inputs and outputs is 0. Behavior to be fixed in another PR. Unit tests that have NoOp make a cluster with no inputs and outputs. disabled support of `NoOp` for them.

- All pipelines except --var go to Parallel Executor.

PR #372, opened 2019-12-10
Kanvi/Transpose PTS

kanvi-nervana:
To convey information about "depth" even if an encapsulate has no inputs/outputs


ngraph_bridge/ngraph_pipelined_tensors.cc
```diff
@@ -91,30 +91,23 @@ bool IndexLibrary::is_free(size_t id) {
...
-    : m_in_tensors(in),
-      m_out_tensors(out),
-      m_num_inputs(in.size()),
-      m_num_outputs(out.size()) {
+    : m_in_tensors(in), m_out_tensors(out) {
+  auto m_depth_in = in.size();
+  auto m_depth_out = out.size();
+
+  // We assume that input and output depths are same
+  m_depth = std::min(m_depth_in, m_depth_out);
```
sayantan-nervana: Lets throw an error here

ngraph_bridge/ngraph_pipelined_tensors.cc
```diff
@@ -91,30 +91,23 @@ bool IndexLibrary::is_free(size_t id) {
...
-      m_depth = out[0].size();
-    } else {
-      m_depth = 0;  // The executable has no inputs and outputs
-    }
+  bool has_inputs = in[0].size() > 0;
+  bool has_outputs = out[0].size() > 0;
+  if (has_inputs && !has_outputs) {
+    m_depth = in.size();
+  } else if (!has_inputs && has_outputs) {
+    m_depth = out.size();
```
sayantan-nervana: Maybe we dont need this if-else here. We can just check `m_depth_in== m_depth_out`. and pick that as m_depth

ngraph_bridge/ngraph_pipelined_tensors.cc
```diff
@@ -132,8 +125,10 @@ void PipelinedTensorsStore::return_tensors(size_t id) {
 PipelinedTensorVector PipelinedTensorsStore::get_group(bool is_input,
                                                        size_t i) {
   PipelinedTensorVector group;
-  for (size_t idx = 0; idx < (is_input ? m_num_inputs : m_num_outputs); idx++) {
-    group.push_back((is_input ? m_in_tensors : m_out_tensors)[idx][i]);
+  if (is_input) {
+    group = m_in_tensors[i];
+  } else {
+    group = m_out_tensors[i];
```
sayantan-nervana: Lets get rid of `PipelinedTensorVector group` and return in the if/else itself to get rid of an extra copy

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -558,13 +558,20 @@ Status NGraphEncapsulateImpl::UpdatePipelinedTensorCache(
     // Create these pipelined ng tensors only if needed, else reuse from cache
```
shresthamalik: Need to remove this
https://github.com/tensorflow/ngraph-bridge/pull/356/files#diff-a499ec3819d5249725dc073c6d8d7af7R11
and this https://github.com/tensorflow/ngraph-bridge/pull/356/files#diff-984639fceaea8131336f5fd39edd6450R165

we were disabling NoOp to get by the error. Will look into rest of the PR later

shresthamalik: Are there any unit tests that can be added?

sayantan-nervana: maybe a unit test is we create a PTS with 2 PipelinedTensorMatrix, with different depths. The construction should throw an error

kanvi-nervana: Added the simple test

test/python/test_variableops_static_input.py
```diff
@@ -159,11 +159,6 @@ def test_variable_static_input_variables_share_buffer(self):
         env_var_map = self.store_env_variables([buffer_sharing_env])
         self.set_env_variable(buffer_sharing_env, "1")
 
-        # disable NoOp it forms a single cluster with no inputs or outputs
-        # Pipelined Tensor Store thinks it is depth 0 and errors out
```
shresthamalik: there are two tests in this file. please do it for both and in clean up we are also setting ngraph_bridge.set_disabled_ops('')

I think there will be 4 lines change in this file

test/test_pipelined_tensor_store.cc
```diff
@@ -0,0 +1,48 @@
...
+
+TEST(PipelinedTensorStoreTest, SimpleTest) {
+  int in_depth = 2;
+  int out_depth = 3;
+  PipelinedTensorMatrix pipelined_input_tensors(in_depth);
+  PipelinedTensorMatrix pipelined_output_tensors(out_depth);
+  bool test_passed = false;
+  try {
+    PipelinedTensorsStore* pts = new PipelinedTensorsStore(
+        pipelined_input_tensors, pipelined_output_tensors);
```
shresthamalik: you can ASSERT_THROW()

test/python/test_variableops_static_input.py
sindhu-nervana: ya, you need to remove here as well - https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/test_variableops_static_input.py#L118

ngraph_bridge/ngraph_executor.cc
```diff
@@ -477,32 +477,30 @@ NGraphExecutor::InitializeIOTensorPipeline(
         nullptr);
   }
   // Create these pipelined ng tensors only if needed, else reuse from cache
-  size_t num_pipelined_inputs = pipelined_input_indexes.size();
-  size_t num_pipelined_outputs = pipelined_output_indexes.size();
+  size_t num_inputs = ng_exec->get_parameters().size();
```
shresthamalik: this seems incorrect. please take the incoming change
size_t num_pipelined_inputs = pipelined_input_indexes.size();
  size_t num_pipelined_outputs = pipelined_output_indexes.size();

ngraph_bridge/ngraph_executor.cc
```diff
@@ -477,32 +477,30 @@ NGraphExecutor::InitializeIOTensorPipeline(
...
-                 << " No. of Pipelined Pipelined Outputs: "
-                 << num_pipelined_outputs;
-  PipelinedTensorMatrix pipelined_input_tensors(num_pipelined_inputs);
-  PipelinedTensorMatrix pipelined_output_tensors(num_pipelined_outputs);
-
-  for (size_t i = 0; i < num_pipelined_inputs; i++) {
-    int input_index = pipelined_input_indexes[i];
-    pipelined_input_tensors[i] =
-        ng_exec->create_input_tensor(input_index, m_depth);
+  NGRAPH_VLOG(5) << "InitializeIOTensorPipeline: In: " << num_inputs
```
shresthamalik: Same comment here.

PR #377, opened 2019-12-11
Kanvi/Create cc files

kanvi-nervana:
Divide ngraph_var and ngraph_find_replace_prefetchdataset into .h and .cc respectively


bazel/BUILD
```diff
@@ -81,6 +81,7 @@ cc_library(
         "ngraph_bridge/ngraph_enter_prefetch_in_catalog.cc",
         "ngraph_bridge/ngraph_executor.cc",
         "ngraph_bridge/ngraph_encapsulate_op.cc",
+        "ngraph_bridge/ngraph_find_replace_prefetchdataset.cc",
```
kanvi-nervana: This should be removed since we do not do the prefetch for grappler build and bazel uses grappler.

shresthamalik: prefetch can be done for any build. it is guarded by the env flag

kanvi-nervana: ok then I'll leave it here

PR #381, opened 2019-12-12
Shrestha/tag rc2

shresthamalik:


ngraph_bridge/ngraph_executor.cc
```diff
@@ -308,7 +308,8 @@ NGraphExecutor::CreateCallback(const std::string signature,
         "tf_function_" + m_node_name + "_" + to_string(rank_id) + ".json",
         serialized_ng_func);
     if (status != Status::OK()) {
-      return std::make_pair(status, std::make_pair(ng_exec, ng_function));
+      return std::make_pair(status,
```
kanvi-nervana: Isn't this the distributed build fix change that went in already?

shresthamalik: that went in the master, this is on r0.21 to create a tag

kanvi-nervana: May be you want to cherry-pick it separately so that tagging can be a stand-alone PR

shresthamalik: Its fine. Not a big change

PR #387, opened 2019-12-16
mbubula/BatchMatMul fix

maxbubula:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -976,7 +976,8 @@ static Status TranslateBatchMatMulV2Op(
           "for each dimension. Failed to match dimension ",
           i, " where lhs was ", ng_lhs_shape[i], " and rhs was ",
           ng_rhs_shape[i]);
-    } else if ((ng_lhs_shape[i] == 1) || (ng_rhs_shape[i] == 1)) {
+    } else if (((ng_lhs_shape[i] == 1) || (ng_rhs_shape[i] == 1)) &&
+               (ng_lhs_shape[i] != ng_lhs_shape[i])) {
       return errors::Unimplemented(
           "ng_lhs_shape and ng_rhs_shape must be the same for each dimension, "
           "for current implementation of BatchMatMulV2."
```
bani-intelaipg: Suggesting to add error message, " or either should be 1 dim, "

bani-intelaipg: Also, put  a space at the beginning of the next line: "Failed to match dimension ",

maxbubula: Done.

maxbubula: What do you mean? If you mean to add following line to previous error line I think `broadcastable` is enough. Or if it is about current line it would be something like:  `ng_lhs_shape and ng_rhs_shape must be the same for each dimension or either should be 1 dim, for current implementation of BatchMatMulV2. Failed to match dimension...`, but if either dimensions are one they are also same.

PR #392, opened 2019-12-18
sindhu/register backends

sindhu-nervana:
This PR fixes static build and static linking with recent ngraph core level changes


test/graph_rewrites/op_by_op_capability_test.cc
```diff
@@ -62,8 +62,7 @@ TEST(OpByOpCapability, Backend) {
   ASSERT_OK(BackendManager::GetCurrentlySetBackendName(&ng_backend_type));
 
   // Map with all the backends, and what the boolean is_supported should be
-  std::map<std::string, bool> backend_map{
-      {"CPU", true}, {"INTERPRETER", true}, {"NOP", false}};
+  std::map<std::string, bool> backend_map{{"CPU", true}, {"INTERPRETER", true}};
```
sindhu-nervana: removed NOP case here because we do not have static library available for this backend, so we cannot register NOP with static builds - so this test will fail.

test/test_ngraph_exec.cpp
```diff
@@ -25,6 +25,7 @@
 #include "ngraph_bridge/ngraph_builder.h"
 #include "ngraph_bridge/ngraph_pipelined_tensors.h"
 #include "ngraph_bridge/ngraph_utils.h"
+#include "ngraph_bridge/version.h"
```
kanvi-nervana: this is not required anymore

examples/cpp/inference/infer_multiple_networks.cc
```diff
@@ -164,6 +164,11 @@ int main(int argc, char** argv) {
     return -1;
   }
 
+// Register backends for static linking
```
kanvi-nervana:
```suggestion
// Register cpu backend for static linking
```

examples/cpp/inference/infer_single_network.cc
```diff
@@ -160,6 +160,11 @@ int main(int argc, char** argv) {
     return -1;
   }
 
+// Register backends for static linking
```
kanvi-nervana:
```suggestion
// Register cpu backend for static linking
```

examples/cpp/inference/infer_multiple_networks.cc
```diff
@@ -164,6 +164,11 @@ int main(int argc, char** argv) {
     return -1;
   }
 
+// Register cpu backend for static linking
+#if defined(NGRAPH_BRIDGE_STATIC_LIB_ENABLE)
+  ngraph_register_cpu_backend();
+#endif
+
```
sindhu-nervana: Right now we create only CPU backend here, hence the ifdef is only for CPU. But suggestions for future is we can query the backend we are trying to set/create and register just that backend if NGRAPH_BRIDGE_STATIC_LIB_ENABLE is enabled

examples/cpp/inference/infer_multiple_networks.cc
```diff
@@ -164,6 +164,11 @@ int main(int argc, char** argv) {
     return -1;
   }
 
+// Register cpu backend for static linking
```
shresthamalik: do we need to register here, since we have a registration in create backend?

examples/cpp/inference/infer_single_network.cc
```diff
@@ -160,6 +160,11 @@ int main(int argc, char** argv) {
     return -1;
   }
 
+// Register cpu backend for static linking
+#if defined(NGRAPH_BRIDGE_STATIC_LIB_ENABLE)
```
shresthamalik: same comment

bazel/BUILD
```diff
@@ -134,6 +134,7 @@ cc_library(
         "-Wno-strict-overflow",
         '-D NGRAPH_BRIDGE_STATIC_LIB_ENABLE',
         '-D NGRAPH_TF_USE_GRAPPLER_OPTIMIZER',
+        '-D NGRAPH_CPU_STATIC_LIB_ENABLE',
```
shresthamalik: Do we build only CPU for bazel? INTERPRETER backend is not build?

examples/cpp/inference/infer_multiple_networks.cc
sindhu-nervana: yeah it was failing otherwise because we are using set backend here

bazel/BUILD
sindhu-nervana: We did not need to for the application that use bazel like hello_tf an infer_multi. But if needed in future we could add a rule for INTERPRETER as well

examples/cpp/inference/infer_multiple_networks.cc
sindhu-nervana: TODO: Revisit this to see if we can remove registering here and register only in BackendManager.

shresthamalik:
```suggestion
// Register cpu backend for static linking
// [TODO]: Revisit this to see if we can remove registering here and register only in BackendManager.
```

PR #395, opened 2019-12-20
Bani / Const Op handling for empty and shorter vector values NGTF-2442

bani-intelaipg:


test/python/test_const.py
```diff
@@ -0,0 +1,87 @@
...
+        reason="NGraph will throw error, but TF will fill in zeros")
+    def test_const_empty(self):
+        log = logging.getLogger('test_const_empty')
+        zz = tf.constant([], dtype=float, shape=[2, 3])
+
+        def run_test(sess):
+            log.debug('Invoking sess.run(zz)')
+            return sess.run(zz)
+
+        assert (
```
shresthamalik: instead of skipping the test we could assert on error being thrown?

ngraph_bridge/ngraph_utils.h
```diff
@@ -140,32 +140,38 @@ Status ValuesFromConstNode(const NodeDef& node,
       n_elements *= shape.dim(i).size();
     }
     values->resize(n_elements);
+
+    auto val_lastsaved = 0;
+#define SET_VALS(val_size, val_vec)                        \
```
shresthamalik: can we use a templated function instead?

bani-intelaipg: Restructured the code

bani-intelaipg:
@shresthamalik please review the latest changes before merging.

PR #401, opened 2019-12-20
Shrestha/var in compute

shresthamalik:
- Enabled --var build to use parallel executor integrating weights-on-device and data pipelining
-moved ngraph_var files outside the var build


kanvi-nervana:
@googlebot I consent.

PR #402, opened 2019-12-20
Add asserts in some python tests

kanvi-nervana:


kanvi-nervana:
we will need to create a new one, so closing this

PR #404, opened 2019-12-20
Kanvi/fused BMMT support

kanvi-nervana:
NGraph now supports a fused BatchMatMulTranspose. This PR is make changes in the translation for BatchMatMul to accommodate that.


ngraph_bridge/ngraph_builder.cc
```diff
@@ -812,10 +812,7 @@ static Status TranslateBatchMatMulOp(
 
   auto ng_lhs_axes = out_axes;
   auto ng_rhs_axes = out_axes;
-
-  // Get the backend name, if the backend is CPU and n_dims >= 3
-  // then use the BatchMatMul op supported by nGraph
-  if (n_dims >= 3 && backend_name == "CPU") {
+  if (n_dims == 2) {
```
sayantan-nervana: Maybe we do not need the else parts:
```
else {
      ng_lhs_axes.push_back(n_dims - 2);
      ng_lhs_axes.push_back(n_dims - 1);
    }
```

sayantan-nervana: I'd say we can construct `ng_lhs_axes` and `ng_rhs_axes` in each of the cases using iota instead of push_back (which is slower probably) and get rid of `out_axes`

for example in case n>3, only 1 needs to be constructed. we do not need both `ng_lhs_axes` and `ng_rhs_axes`

sayantan-nervana: Computing `ng_lhs_shape = ng_lhs->get_shape();` is not needed in n==2 case. There is some leftover/legacy code from the previous implementation that we can clean up

```
ng_lhs = ng::builder::numpy_transpose(ng_lhs, tf_adj_x ? {1,0} : {0,1}); // or ng::AxisVector{0,1}
Builder::SetTracingInfo(op->name(), ng_lhs);
ng_rhs = ng::builder::numpy_transpose(ng_lhs, tf_adj_y ? {1,0} : {0,1});
Builder::SetTracingInfo(op->name(), ng_rhs);
SaveNgOp(ng_op_map, op->name(),
             ConstructNgNode<ngraph::op::Dot>(op->name(), ng_lhs, ng_rhs));
```

kanvi-nervana: Done

ngraph_bridge/ngraph_builder.cc
```diff
@@ -786,170 +786,83 @@ static Status TranslateBatchMatMulOp(
...
     // Transpose X if AdjX = true
     if (tf_adj_x) {
-      ng_lhs_axes.push_back(n_dims - 1);
-      ng_lhs_axes.push_back(n_dims - 2);
-      ng_lhs = ng::builder::numpy_transpose(ng_lhs, ng_lhs_axes);
+      ng_lhs = ng::builder::numpy_transpose(ng_lhs, {1, 0});
```
sayantan-nervana: these 2 are not needed/used for n_dims == 2
ng_lhs_shape = ng_lhs->get_shape();
ng_rhs_shape = ng_rhs->get_shape();

PR #413, opened 2019-12-24
Fix for NGRAPH_DISTRIBUTED

vishakha-nervana:


ngraph_bridge/ngraph_executor.cc
```diff
@@ -295,12 +295,7 @@ NGraphExecutor::CreateCallback(const std::string signature,
   // Serialize to nGraph if needed
   if (std::getenv("NGRAPH_ENABLE_SERIALIZE") != nullptr) {
     std::string file_name = "tf_function_" + m_node_name + ".json";
```
shresthamalik: is this file_name being used anywhere?

vishakha-nervana: No, it is not.

shresthamalik: Please remove it then.

vishakha-nervana: Done

PR #422, opened 2020-01-06
sarkars/Move all maps

sayantan-nervana:
Need to add some new functionality to MarkForClustering(MFC) (in a separate PR). MFC is currently monolithic, with all the maps defined inside it. Therefore we cannot access the maps outside MFC. To add the new functionality, I need access to the maps outside MFC. Hence moving the maps outside.

Also added #ifdef DISTRIBUTED to TFtoNgraphOpMap where needed

Now that the map has moved outside, also added test in opexecutor to ensure that the translated function contains ops present in the TfToNg map. With that test, found that a bunch of gtests fail, which have been fixed now.

Also added "MaximumBroadcasting", (MinimumBroadcasting was present already)


ngraph_bridge/ngraph_mark_for_clustering.h
```diff
@@ -30,14 +30,27 @@ Status MarkForClustering(Graph* graph, std::set<string> skip_these_nodes,
...
 bool InputIsStatic(const Node* node, int index);
 Status GetNodeBackend(const Node* node, string* backend_name);
 void SetNodeBackend(Node* node, const string& backend_name);
+
+using SetAttributesFunction = std::function<Status(Node*)>;
+const std::map<std::string, SetAttributesFunction>& GetAttributeSetters();
+
+using TypeConstraintMap =
+    std::map<std::string, std::map<std::string, gtl::ArraySlice<DataType>>>;
+const TypeConstraintMap& GetTypeConstraints();
```
kanvi-nervana: minor GetTypeConstraints ---> GetTypeConstraintMap since that is what it is returning

sayantan-nervana: Done.

PR #423, opened 2020-01-06
Kanvi/update TFtoNgraphOpMap

kanvi-nervana:


sayantan-nervana:
Added changes of this PR in https://github.com/tensorflow/ngraph-bridge/pull/422

PR #426, opened 2020-01-07
Sarkars/add analysis pass mode to encapsulate cluster

sayantan-nervana:
`EncapsulateCluster` has 3 main jobs:
- Analysis
- Rewriting
- AOT

Added an `Encapsulator` class, which splits the workload into 2 functions `AnalysisPass` (`ClusterManager` is populated here) and `RewritePass` (`Encapsulate` op is created, and spliced into the graph, and the nodes it replaces are deleted). The AOT business is moved out to a stand-alone function `PerformAOTOnEncapsulates`

Note pass 5 has moved ahead to analysis phase and it has been slightly modified


test/graph_rewrites/encapsulate_clusters_test.cc
```diff
@@ -32,6 +32,196 @@ namespace ngraph_bridge {
 
 namespace testing {
 
+// Test that calls the functions of encapsulator in the wrong order
+// Non-OK statuses are axpected
```
sayantan-nervana:
```suggestion
// Non-OK statuses are expected
```

ngraph_bridge/ngraph_encapsulate_clusters.h
```diff
@@ -35,10 +39,107 @@ typedef std::map<std::string, std::vector<int>> ShapeHintMap;
...
+// Given a TF graph, it scans it for inputs and finds what TF is saying about
+// their shapes (in the .pb itself)
+// Creates a map between input node names and PartialShape information we get
+// from the TF graph
+std::map<std::string, PartialShape> GetShapesFromTFInputnodes(
+    Graph* graph, const string& input_node_type);
+
+// Given an encapsulate node, and the input shapes,
+// performs TranslateGraph and returns an ng function and a signature
+Status PerformTranslation(Node* node,
```
shresthamalik: This function should be taking in a TFGraph rather than a Node. Can be addressed in some other PR

ngraph_bridge/ngraph_encapsulate_clusters.cc
```diff
@@ -529,382 +809,185 @@ Status EncapsulateClusters(
...
+      // empty, and in that case the empty shape and the rank!=0 hint fuse
+      // to give an invalid shape according to our current logic. have to
+      // modify that
+      PartialShape partial_shape_from_node;
+      if (shape_field != nullptr) {
+        // Get shape from the node
+        partial_shape_from_node = PartialShape(shape_field->shape());
       }
-    }  // end of for (ShapeHintMap single_hint : node_shapes_hints_sets)
+      NGRAPH_VLOG(5) << "For node " << node->name() << " got shape from nose: "
```
shresthamalik:
```suggestion
      NGRAPH_VLOG(5) << "For node " << node->name() << " got shape from node: "
```

shresthamalik: 😆 " got shape from nose: "

ngraph_bridge/ngraph_encapsulate_clusters.h
```diff
@@ -35,10 +39,107 @@ typedef std::map<std::string, std::vector<int>> ShapeHintMap;
...
+class Encapsulator {
+ public:
+  Encapsulator(Graph* g);
+  // Populate ClusterManager with the subgraphs for each potential encapsulate
+  Status AnalysisPass();
+  // Perform the actual graph surgery
+  Status RewritePass(
+      FunctionDefLibrary* fdeflib, int graph_id,
+      const std::unordered_map<std::string, std::string>& device_config);
+  Status NewClusterIds(std::set<int>& result);
```
shresthamalik: add a comment on what this function does

ngraph_bridge/ngraph_encapsulate_clusters.h
sayantan-nervana: This takes a node, which is expected to be an encapsulate (it throws an error otherwise). Given that it is an encapsulate, it finds its corresponding sub graph in the clustermanager and translates that.

Not the best API for a function as you mentioned. I could extract from the clustermanager beforehand and pass that to the function. 'll probably simplify it in the next PRs.

ngraph_bridge/ngraph_encapsulate_clusters.h
```diff
@@ -35,10 +39,107 @@ typedef std::map<std::string, std::vector<int>> ShapeHintMap;
...
+// shape hint, combine all that information
+PartialShape CombineNodeInfoAndHint(Node* node,
+                                    PartialShape partial_shape_from_node,
+                                    const ShapeHintMap& single_hint);
+
+// Given a TF graph, it scans it for inputs and finds what TF is saying about
+// their shapes (in the .pb itself)
+// Creates a map between input node names and PartialShape information we get
+// from the TF graph
+std::map<std::string, PartialShape> GetShapesFromTFInputnodes(
```
shresthamalik: Similar comment for this. It should work only for a placeholder as it is invalid for any other node type? For eg, if i pass in  input_node_type as Add, the function will probably happily work. There is no check whether input_node_type is really a boundary/input node. 

GetShapesFromTFInputnodes could be GetShapesFromPlaceholder

Can be addressed in another PR

sayantan-nervana: Yes, this might change in later PRs. Right now it only accepts placeholders etc. But soon users might provide hints about "max" in the middle of the graph (say for the where op), so this section might need a redo

shresthamalik: ok, so node_names will be provided rather than type?

sayantan-nervana: the function `GetShapesFromTFInputnodes` takes a graph and a string. This string is basically "placeholder". So this function scans the graph for "placeholder" type nodes and finds what TF is saying about their shapes. Note `const string& input_node_type` as a separate input is probably not needed, Originally I had that as an argument as we get "placeholders" in case of grappler and "args" in case of normal pass.

AOT is supposed to work with grappler only, hence for now the only value input_node_type accepts is the string "placeholder". Maybe I can remove the argument. But its there for historical reasons, in case we want AOT to work with normal pass as well

ngraph_bridge/ngraph_encapsulate_clusters.cc
```diff
@@ -75,30 +75,287 @@ static void AddInput(NodeDef* dst, StringPiece src_name, int src_slot) {
...
+  Encapsulator enc(graph);
+  NGRAPH_VLOG(3) << "Running AnalysisPass in EncapsulateClusters";
+  TF_RETURN_IF_ERROR(enc.AnalysisPass());
+  NGRAPH_VLOG(3) << "Running RewritePass in EncapsulateClusters";
+  TF_RETURN_IF_ERROR(enc.RewritePass(fdeflib, graph_id, device_config));
+  NGRAPH_VLOG(3) << "Performing AOT in EncapsulateClusters";
+  TF_RETURN_IF_ERROR(PerformAOTOnEncapsulates(graph, aot_info));
+
+  set<int> newly_created_cluster_ids;
+  TF_RETURN_IF_ERROR(enc.NewClusterIds(newly_created_cluster_ids));
```
vishakha-nervana: Will CreateClusterIDs be a better name?

sayantan-nervana: It does not created new ids. it just returns the new ones that Analysis pass created.
Its more of a "GetNewlyCreatedClusterIDs"

sayantan-nervana: Come to think of it, thats what I named my argument (newly_created_cluster_ids) 🤔

vishakha-nervana: So may be GetNewClusterIDs(newly_created_cluster_ids)

sayantan-nervana: Changed to new name.

PR #430, opened 2020-01-09
Vishakha/ngraph device support

vishakha-nervana:


adk9:
Closing because we've deprecated support for nGraph as a device. Thanks.

PR #441, opened 2020-01-16
Jitendra/floordiv fix

jitendra42:
For `floordiv` op: If input is integer it will be using `div` otherwise `floor(div())`

### DIFF
--- a/ngraph_bridge/ngraph_builder.cc
+++ b/ngraph_bridge/ngraph_builder.cc
@@ -1977,13 +1977,25 @@ static Status TranslateFillOp(
 static Status TranslateFloorDivOp(
     const Node* op, const std::vector<const Tensor*>& static_input_map,
     Builder::OpMap& ng_op_map) {
-  auto ng_floordiv = [&op](std::shared_ptr<ng::Node> ng_input1,
-                           std::shared_ptr<ng::Node> ng_input2) {
-    return ConstructNgNode<ng::op::Floor>(
-        op->name(),
-        ConstructNgNode<ng::op::Divide>(op->name(), ng_input1, ng_input2));
-  };
-  return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_floordiv);
+  DataType dtype;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "T", &dtype));
+  if(dtype == DT_INT32) {
+    std::cout <<"###JP: its DT_INT32" << dtype << std::endl;
+    auto ng_div = [&op](std::shared_ptr<ng::Node> ng_input1,
+                            std::shared_ptr<ng::Node> ng_input2) {
+      return ConstructNgNode<ng::op::Divide>(op->name(), ng_input1, ng_input2);
+    };
+    return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_div);
+  } else {
+    std::cout <<"###JP: Its DT_FLOAT: "<< std::endl;
+    auto ng_floordiv = [&op](std::shared_ptr<ng::Node> ng_input1,
+                            std::shared_ptr<ng::Node> ng_input2) {
+      return ConstructNgNode<ng::op::Floor>(
+          op->name(),
+          ConstructNgNode<ng::op::Divide>(op->name(), ng_input1, ng_input2));
+    };
+    return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_floordiv);
+  }
 }
 
 static Status TranslateFloorModOp(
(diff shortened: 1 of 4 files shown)


ngraph_bridge/ngraph_builder.cc
```diff
@@ -1977,13 +1977,23 @@ static Status TranslateFillOp(
...
-  auto ng_floordiv = [&op](std::shared_ptr<ng::Node> ng_input1,
-                           std::shared_ptr<ng::Node> ng_input2) {
-    return ConstructNgNode<ng::op::Floor>(
-        op->name(),
-        ConstructNgNode<ng::op::Divide>(op->name(), ng_input1, ng_input2));
-  };
-  return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_floordiv);
+  DataType dtype;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "T", &dtype));
+  if (dtype == DT_INT32) {
```
sayantan-nervana: you could use `NGraphIndexDTypes`
https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_utils.cc#L344

ngraph_bridge/ngraph_builder.cc
```diff
@@ -1977,13 +1977,23 @@ static Status TranslateFillOp(
...
+    return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_div);
+  } else {
+    auto ng_floordiv = [&op](std::shared_ptr<ng::Node> ng_input1,
+                             std::shared_ptr<ng::Node> ng_input2) {
+      return ConstructNgNode<ng::op::Floor>(
+          op->name(),
+          ConstructNgNode<ng::op::Divide>(op->name(), ng_input1, ng_input2));
+    };
+    return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_floordiv);
+  }
```
sayantan-nervana: I wonder if a static analyzer like klocwork or some compiler might complain "not returning anything from a non-void function"

we can do a `return errors::Internal("We should never reach this part in translating floordiv");`

or we can declare function outside the ifelse `Status std::function<std::shared_ptr<ng::Node>, std::shared_ptr<ng::Node>>`, populate it inside the ifelse and then have a single line outside the if-else that says `return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_bin_fn);`

jitendra42:
@sayantan-nervana Updated PR with code review changes.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -1977,13 +1977,27 @@ static Status TranslateFillOp(
...
-  auto ng_floordiv = [&op](std::shared_ptr<ng::Node> ng_input1,
-                           std::shared_ptr<ng::Node> ng_input2) {
-    return ConstructNgNode<ng::op::Floor>(
-        op->name(),
-        ConstructNgNode<ng::op::Divide>(op->name(), ng_input1, ng_input2));
-  };
-  return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_floordiv);
+  DataType dtype;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "T", &dtype));
+  std::initializer_list<DataType> int_types = {DT_INT8, DT_UINT16, DT_INT16,
```
shresthamalik: you can create Int types in utils so that it can be reused for some other op
https://github.com/tensorflow/ngraph-bridge/blob/master/ngraph_bridge/ngraph_utils.cc#L354

sayantan-nervana: Probably a set instead of initializer_list.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -1977,13 +1977,27 @@ static Status TranslateFillOp(
...
+  } else {
+    auto ng_floordiv = [&op](std::shared_ptr<ng::Node> ng_input1,
+                             std::shared_ptr<ng::Node> ng_input2) {
+      return ConstructNgNode<ng::op::Floor>(
+          op->name(),
+          ConstructNgNode<ng::op::Divide>(op->name(), ng_input1, ng_input2));
+    };
+    return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_floordiv);
+  }
+  return errors::Internal(
```
shresthamalik: Please remove this error
" We should never reach this part in translating floordiv");"

And move this out of the if-else
return TranslateBinaryOp(op, static_input_map, ng_op_map, ng_floordiv);

jitendra42:
updated PR with review changes

PR #456, opened 2020-01-22
ashahba/Cherry-pick TF v1.15.0 changes from r0.19 into master

ashahba:
Upgrade default `TensorFlow` for `ngraph-bridge` to `v1.15.0`

Now we need `GCC 5.3` or better to build the code.

This also addresses the fix for #492


sayantan-nervana:
@ashahba : for fixing the binary tf pipeline maybe we can just change the queue to cpu (from cpu-gcc4.8" in the binary tf yaml?

Later on, we can eliminate the cpu-gcc4.8 queue, and repurpose that machine for the cpu queue?

kanvi-nervana:
Just noticed that tf-binary queue fails on cpu queue, might need to be looked at since we were expecting it to pass, right?

rsketine:
@ashahba we need this to be reviewed by packaging team before including this. Number of implication in making this change. here is why

TF is releasing their software with GCC 4.8.5 and ABI = 0
   --- they moved from manylinux1 to manylinux2010 starting from their 1.15.0 release. 
   --- https://www.python.org/dev/peps/pep-0571/#the-manylinux2010-policy (All the versions of the libraries used are listed here.)
The option -std=c++14 as an option which requires this package to be added  as this feature was not added before this. https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7/
Explanation in this bug: https://github.com/tensorflow/tensorflow/issues/32677

This change has lots of implications on packaging, distribution and compatibility with TF itself. Please look into this and let us know we can also discuss.

test/ci/buildkite/ngtf-cpu_ubuntu-bin-build.yaml
```diff
@@ -5,33 +5,36 @@
...
-    - "queue=cpu_gcc_4.8"
+    - "queue=cpu"
     parallelism: 1
 
   - wait
 
   - command: |
       source /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/venv/bin/activate 
+      git clone https://github.com/tensorflow/tensorflow.git /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/tensorflow
+      ln -sf /localdisk/buildkite-agent/prebuilt_tensorflow_1_15_0/artifacts/tensorflow/tensorflow-1.15.0-cp36-cp36m-linux_x86_64.whl /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/tensorflow
```
sayantan-nervana: Can you please add some comments about these lines and why they are needed

.gitignore
```diff
@@ -12,3 +12,5 @@ cpu_codegen/
 
 .*.swp
 .nfs*
+build_cmake
```
shresthamalik: please remove the .gitignore

README.md
```diff
@@ -131,13 +131,13 @@ Verify that `ngraph-bridge` installed correctly:
...
-    nGraph version used for this build: b'0.18.0+c5d52f1'
-    TensorFlow version used for this build: <v1.14.0-...>
-    CXX11_ABI flag used for this build: 0
-    nGraph bridge built with Grappler: False
-    nGraph bridge built with Variables and Optimizers Enablement: False
+        TensorFlow version:  <1.15.0>
+        nGraph bridge version: <b'0.14.0'>
+        nGraph version used for this build: b'0.18.0+c5d52f1'
+        TensorFlow version used for this build: <v1.15.0-...>
+        CXX11_ABI flag used for this build: 0
```
shresthamalik: should this be 1 `CXX11_ABI flag used for this build: 0` ?

.gitignore
sayantan-nervana: the build_cmake is a reasonable gitignore i guess? we dont want to commit anything we generate in build_cmake

the test/**/dump.sh is generated from model tester. Again it never meant to be committed. So might as well add them to gitignore?

build_ngtf.py
```diff
@@ -237,9 +237,9 @@ def main():
     print("Target Arch: %s" % target_arch)
 
     # The cxx_abi flag is translated to _GLIBCXX_USE_CXX11_ABI
-    # For gcc 4.8 - this flag is set to 0 and newer ones, this is set to 1
+    # For gcc 7.3 - this flag is set to 0 and newer ones, this is set to 1
     # The specific value is determined from the TensorFlow build
-    # Normally the shipped TensorFlow is built with gcc 4.8 and thus this
+    # Normally the shipped TensorFlow is built with gcc 7.3 and thus this
     # flag is set to 0
     cxx_abi = "0"
```
shresthamalik: cxx_abi  set to 1?

test/python/test_tf2ngraph_script.py
```diff
@@ -163,6 +163,8 @@ def test_command_line_api(self, inp_format, inp_loc, out_node_name,
             # Comparing with expected value
             assert np.isclose(res1, exp).all()
 
+    #TODO: Unskip this test
+    @pytest.mark.skip(reason="Does not work with tf v1.15.0")
```
shresthamalik: @sayantan-nervana fyi. This test is being skipped.

build_ngtf.py
```diff
@@ -291,9 +291,24 @@ def main():
             os.chdir(pwd_now)
 
             # Next install the tensorflow python packge
-            command_executor(
-                ["pip", "install", "-U", "tensorflow==" + tf_version])
-            cxx_abi = get_tf_cxxabi()
+            if ('1.15' not in tf_version):
```
shresthamalik: Not sure about this change. In the option `use prebuilt tf` we should pip install tf 1.15 and built. 
But it seems we are building tf from source first.

test/python/test_tf2ngraph_script.py
sayantan-nervana: @ashahba : If its working now lets reenable it. If not, can you please create a ticket to reenable it, so that it is tracked.

build_ngtf.py
ashahba: It's get's generated and overwritten correctly later on, and this is just the initialization.

ashahba: But sure, I'll change that.

shresthamalik: yes, saw that it is overwritten. Basically, the comment was not aligned
```
 # Normally the shipped TensorFlow is built with gcc 7.3 and thus this
    # flag is set to 0
```

build_ngtf.py
ashahba: The prebuilt TF v1.15.0 from `Google` is badly broken and we can use it with either GCC `4.8` or newer to build the bridge. This is what this whole PR is about.
Hopefully for `TF v2.1` and after there is only one GCC we have to deal with which is `7.3` and we should be good to use upstream wheels again.

shresthamalik: You mean, we cant use it to build with gcc 4.8 or higher ?

ashahba: Yes we can't use it on either `4.8` or newer to build bridge so we have to rebuilt it with right flags first.

test/python/test_tf2ngraph_script.py
ashahba: Let me try uncommenting this and running it and if it fails, I'll file a ticket or look into a fix.

build_ngtf.py
ashahba: Good catch. I fixed it

build_ngtf.py
shresthamalik: Wondering if we should error out in that case rather than building from source. We have few other options:
1. vanilla build : builds tf fom source + builds all our cpp applications
2. --use_tf_from_location : re-uses the .whl and certain tf libraries to builds bridge -> builds our cpp applications
3. --use_prebuilt_tf : pip installs tf -> cannot build our cpp applications using this method

shresthamalik: If we error out then our pipeline that tests this will not work. Cant think of any other place where this would be a problem.

ashahba: Options `1` and `2` are still working and it's only the option `3` that's problematic.

I'm Ok with erring out but then do we ask user to build TF themselves and pass it down to us so we can use their wheel?
If we don't have control over how they build the wheel, we could run into the same issue with their wheel as well.

Now as far as pipeline test is concerned we can leave it the way it is I guess.

test/ci/buildkite/ngtf-cpu_ubuntu-bin-build.yaml
ashahba: Done

ashahba:
@shresthamalik and @sayantan-nervana I addressed all the review requests I guess.

test/python/test_tf2ngraph_script.py
ashahba: It's fixed now

build_ngtf.py
shresthamalik: We can error out and ask the user to build using options 1 or 2 temporarily. 
In either case [erroring out or building from src and then using the .whl], It would be good to document the reason in a comment in the else section.

sayantan-nervana: Lets error out in this case. We are still testing the "pip install" pipeline, but for actual usage since we do not support "pip install tensorflow" we should error out. For example, this TF building has bazel dependency etc. Someone expecting to do "pip install tensorflow" may not have bazel dependency etc.

ashahba:
For TF v1.15.0 there is no `--config=v1` and the default and the only supported option is `--config=v2` to built TF. Luckily for TF v1.15.2 they fixed that.
So I’m going to do a minor patch fix once this PR is merged

build_ngtf.py
ashahba: Sounds good. I fixed it.

sayantan-nervana: Notes for myself:
the way its set up right now is that if an user tries to run it it will fail as artifacts_dir/tensorflow/whl will not exist

But for buildkite it will exist as we are copying in the whl in the BK yaml. So BK will proceed

ashahba: 💯

PR #467, opened 2020-01-24
Sarkars/modify mfc to query backend

sayantan-nervana:


ngraph_bridge/ngraph_encapsulate_clusters.cc
```diff
@@ -116,10 +116,61 @@ Status EncapsulateClusters(
...
+      }
+    }  // end of if (node->type_string() == input_node_type)
+  }    // End of for loop that goes through all nodes
+
+  // Did we manage to concretize all input shapes?
+  for (auto itr : node_partial_shape_map) {  // iterate over all inputs
+    if (inputs_node_shapes_for_compilation.find(itr.first) ==
+        inputs_node_shapes_for_compilation.end()) {
+      // TODO: print "this" hint
+      return errors::Internal("Cannot AOT using this hint (",
```
kanvi-nervana:
```suggestion
      return errors::Internal("Cannot use this hint (",
```

ngraph_bridge/ngraph_mark_backend_support.cc
```diff
@@ -62,8 +144,40 @@ Status GetBackendSupportInfoForNgfunction(
...
+  }
+  // Get the TF node support information
+  TF_RETURN_IF_ERROR(
+      GetTFNodeSupportInfo(ng_function, result_map, ng_result_map));
+  return Status::OK();
+}
 
+Status GetTFNodeSupportInfo(const shared_ptr<ng::Function>& ng_function,
+                            std::map<std::string, bool>& result_map,
+                            std::map<std::string, bool>& ng_result_map) {
```
sayantan-nervana: we can add a const probably on ng_result_map

PR #470, opened 2020-01-28
ashahba/Fix 'ngtf-cpu-model-test' check

ashahba:


ashahba:
The  buildkite test also passed here: https://buildkite.com/ngraph/ngtf-cpu-model-test/builds/622

sayantan-nervana:
Model tests pipeline passes: https://buildkite.com/ngraph/ngtf-cpu-model-test/builds/624

PR #471, opened 2020-01-28
Wenzhe/Stridedslicegrad

wenzhe-nrv:


sayantan-nervana:
> Will be good to have a Python level test in test/python/test_stridedslicegrad.py

Something like this: https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/test_conv2dbackpropinput.py#L65

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -665,6 +668,8 @@ const TypeConstraintMap& GetTypeConstraintMap() {
     type_constraint_map["Squeeze"]["T"] = NGraphDTypes();
     type_constraint_map["StridedSlice"]["T"] = NGraphDTypes();
     type_constraint_map["StridedSlice"]["Index"] = NGraphIndexDTypes();
+    type_constraint_map["StridedSliceGrad"]["T"] = NGraphDTypes();
```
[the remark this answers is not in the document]
wenzhe-nrv: added the comment on the translate function

wenzhe-nrv:
`slice_op_test.SliceTest.testGradientsAll` in TF python kernel test covers the strided slice grad op and it passes now. It calls a slice then do an auto diff, instead of calling the `StridedSliceGrad` op directly. I think this is a better than a bridge python test.
I'm working to refactor the code in the op translation uses `ngraph::make_slice_plan`.
@sayantan-nervana @bani-intelaipg

ngraph_bridge/ngraph_builder.cc
```diff
@@ -4725,6 +4725,101 @@ static Status TranslateStridedSliceOp(
...
+  std::vector<int64> stride_vec;
+  TF_RETURN_IF_ERROR(
+      GetStaticInputVector(op, 3, static_input_map, &stride_vec));
+  std::vector<int64_t> stride_vec_longint(stride_vec.begin(), stride_vec.end());
+
+  // get slice plan
+  auto convert_mask_to_axes = [](const int mask) {
+    ng::AxisSet axes{};
+    for (auto i = 0; i < sizeof(int) * 8; ++i) {
+      if ((unsigned char)(mask >> i & 0x01) == 1) {
```
sayantan-nervana: Minor: Please add brackets to make the order of operation more readable.

sayantan-nervana: A lot of the code of TranslateStridedSlice and TranslateStridedSliceGrad is similar (for example convert_mask_to_axes). Can we use a single function or a helper function or somehow reuse the code?

wenzhe-nrv: agree

sayantan-nervana:
Are there any tensorflow python tests possible for this?

you can discover TF tests like this: https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/tensorflow/README.md and then add them here: https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/tensorflow/python_tests_list.txt


maybe these tests here...
https://github.com/tensorflow/tensorflow/blob/66fb5a1d938de51b66d89501313c49dc1c56ee7b/tensorflow/python/kernel_tests/array_ops_test.py#L906

wenzhe-nrv:
@sayantan-nervana thanks for finding the test. Yes, I had tried this test `StridedSliceGradTest.testGradient`, it covers quite some cases that is failing now. mostly negative stride values or reverse. 
I'll spend some time to check if this test could pass or not.

sayantan-nervana:
> I'll spend some time to check if this test could pass or not.
Even if they cant be fixed we can add them to the test list and hash them out

test/python/tensorflow/python_tests_list.txt
```diff
@@ -20,6 +20,10 @@ array_ops_test.StridedSliceTest.testShrink
 array_ops_test.StridedSliceTest.testTensorIndexing
 #array_ops_test.StridedSliceTest.test_basic_slice
 
+#array_ops_test.StridedSliceGradTest.testGradient
+array_ops_test.StridedSliceGradTest.testGradientZero
+#array_ops_test.StridedSliceGradTest.testInt64Indices
```
sayantan-nervana: These tests are hashed out.

Also some other possible tests here: https://github.com/tensorflow/tensorflow/blob/66fb5a1d938de51b66d89501313c49dc1c56ee7b/tensorflow/python/kernel_tests/array_ops_test.py#L955

ngraph_bridge/ngraph_builder.cc
```diff
@@ -373,6 +373,82 @@ Builder::PerformNgBroadcast(const string& prov_tag,
...
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "begin_mask", &attrs.begin));
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "end_mask", &attrs.end));
+  TF_RETURN_IF_ERROR(
+      GetNodeAttr(op->attrs(), "new_axis_mask", &attrs.new_axis));
+  TF_RETURN_IF_ERROR(
+      GetNodeAttr(op->attrs(), "shrink_axis_mask", &attrs.shrink_axis));
+  TF_RETURN_IF_ERROR(
+      GetNodeAttr(op->attrs(), "ellipsis_mask", &attrs.ellipsis));
+
+  NGRAPH_VLOG(5) << "Strieded Slice Mask Attributes: "
```
sayantan-nervana:
```suggestion
  NGRAPH_VLOG(5) << "Strided Slice Mask Attributes: "
```

ngraph_bridge/ngraph_builder.cc
```diff
@@ -373,6 +373,82 @@ Builder::PerformNgBroadcast(const string& prov_tag,
...
+  }
+
+  NGRAPH_VLOG(5) << "Slice_plan: begin: " << ng::join(slice_plan.begins)
+                 << ", end: " << ng::join(slice_plan.ends)
+                 << ", stride: " << ng::join(slice_plan.strides)
+                 << ", reshape input shape: " << slice_plan.reshape_in_shape
+                 << ", reshape output shape: " << slice_plan.reshape_out_shape
+                 << ", reverse axis: " << slice_plan.reverse_axes;
+
+  return Status::OK();
```
sayantan-nervana: This function returns only Status::OK() at the end. So maybe we can make it void or just make it return slice_plan directly

wenzhe-nrv: I wondered about this and should probably ask first. Some of the functions in ngraph_builder.cc return status::OK() and pass returns as reference in the functions, such as `GetInputNodes`, `GetStaticInputVector`, etc.., while others don't. I wonder what's the standard for bridge to decide which function needs to return a status while others could return a value.

sayantan-nervana: Sometimes we may not necesarily call any TF function. We may just do some normal cpp op. But if we expect it might throw an error we can return a TF status:

```
Status my_divide(int y, node* n, int& z) {
int x;
TF_RETURN_IF_ERROR(n->get_attribute("foo", &x));
// If "foo" doesnt exist return teh status TF function get_attribute throws

// Got x.
if (y == 0){
  return errors::Internal("Bad denominator");
  // we generate an error message. This error message is not coming from a TF function call
} else {
  z = x/y;
}

return Status::OK();
}
```

PR #475, opened 2020-01-31
Jitendra/Event tracing to chrome tracing

jitendra42:
Moved from event tracing to chrome tracing which is in RAII style

Converts
```
ngraph::Event event("Name", "Category", "Args");
//code
event.Stop()
event.write_trace(event)
```
to
```
{
   NG_TRACE("Name", "Category", "Args")
   //code
}
```


jitendra42:
I tested using `test/test_thread_safe_queue.cc` in `gtest_ngtf` has enable chrome tracing and it is generating chrome trace file. However the name of trace file is different `runtime_event_trace.json` instead of `ngraph_event_trace.json`

Let me test for mnist/axpy examples as well and also will add test.

examples/cpp/inference/infer_multiple_networks.cc
```diff
@@ -264,7 +262,7 @@ int main(int argc, char** argv) {
     // Run the inference loop
     //-----------------------------------------
     for (int i = 0; i < iteration_count; i++) {
-      ngraph::Event evt_iteration(oss.str(), to_string(i), "");
+      NG_TRACE(oss.str(), to_string(i), "");
```
shresthamalik: Do you need braces for this trace?

shresthamalik: Checked not required. Will use the scope of for loop.

examples/cpp/inference/infer_multiple_networks.cc
```diff
@@ -33,10 +33,10 @@
 #include "tensorflow/core/public/session.h"
 #include "tensorflow/core/util/command_line_flags.h"
 
-#include "ngraph/event_tracing.hpp"
-
+#include "ngraph/runtime/chrome_trace.hpp"
```
shresthamalik: Do we need to include this?

test/test_thread_safe_queue.cc
```diff
@@ -44,53 +45,53 @@ TEST(ThreadSafeQueue, Simple) {
   atomic<bool> consumer_do_wait{true};
   atomic<int> item_count{0};
 
-  ngraph::Event::enable_event_tracing();
+  ngraph::runtime::event::Manager::enable_event_tracing();
```
shresthamalik: This may not be required. Is probably here as a stray comment. 
We can remove this if the test is not using the trace.

shresthamalik:
We can do a before and after with mnist and see whether high level the traces look similar.

ngraph_bridge/ngraph_utils.h
```diff
@@ -27,13 +27,16 @@
...
 
-#include "ngraph/event_tracing.hpp"
 #include "ngraph/ngraph.hpp"
+#include "ngraph/runtime/chrome_trace.hpp"
 #include "ngraph/serializer.hpp"
 
 #include "logging/ngraph_log.h"
 #include "logging/tf_graph_writer.h"
 
+#define NG_TRACE(name, category, args) \
```
bani-intelaipg: How about we add a comment for the users of this MACRO?
// Activates event logging until the end of the current code-block scoping; Automatically writes log data as soon as the the scoping ends.

ngraph_bridge/ngraph_utils.h
```diff
@@ -27,13 +27,16 @@
...
+  ngraph::runtime::event::Duration dx__ { name, category, args }
```
sayantan-nervana: Also from good macro writing practice point of view, maybe we wrap these in brackets: `ngraph::runtime::event::Duration dx__ { (name), (category), (args) }`

PR #477, opened 2020-02-01
Properly format all Bazel files

ashahba:


ashahba:
I'm using `buildifier` here: https://github.com/bazelbuild/buildtools/tree/master/buildifier
and here is how I updated the files:

```
$ buildifier $(find . -type f \( -iname *.BUILD -or -iname BUILD.bazel \))
```

ashahba:
@sayantan-nervana If we want to automate this, we should add `buildifier` to the the install list too.

PR #483, opened 2020-02-07
Shrestha/Var_static_input

shresthamalik:
Added handling for case when Variable is a static input to Encapsulate (relevant for --enable_var build) + some refactoring 
- Added utility that GetsStaticInputs indexes of a graph 
- Attach Attribute "_ngraph_static_inputs" to NGraphEncapsulate Op in ngraph_encapsulate_clusters.cc . This information is used in the algorithm in RewritForTracking
- For enable-var build, Rewrite for Tracking now checks if var is a static input to encapsulate and accordingly marks the Var Node to update Var's TF Tensor. Python Test test_variable_op_static_inputs.py now asserts that the results from NG and TF run are same, earlier was asserting that results are not same


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -1301,6 +1301,35 @@ bool InputIsStatic(const Node* node, int index) {
   return std::find(inputs.begin(), inputs.end(), index) != inputs.end();
 }
 
+Status GetStaticInputs(Graph* graph, std::vector<int32>* static_input_indexes) {
```
sayantan-nervana: Minor. We generally use references for outputs. Then we wont have to check if the input pointer is null

test/python/test_variableops_static_input.py
```diff
@@ -129,24 +121,20 @@ def test_variable_static_input_variables_dont_share_buffer(self):
...
         assert np.allclose(ng_var_init_val, tf_var_init_val)
 
-        # 1st iteration mean value will match, 2nd and 3rd wont
-        assert np.allclose(ng_mean_values[0], tf_mean_values[0])
+        # Computed Mean Values
+        assert np.allclose(ng_mean_values, tf_mean_values)
 
-        if ngraph_bridge.are_variables_enabled():
-            assert (np.allclose(ng_mean_values[1], tf_mean_values[1]) == False)
-            assert (np.allclose(ng_mean_values[2], tf_mean_values[2]) == False)
```
sayantan-nervana: Note to self: this test was "failing" before. With this PR it passes

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -46,25 +73,22 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
-
-      bool outputs_ng_supported = true;
+      NGRAPH_VLOG(5) << "Checking: " << DebugNode(node) << " " << node->name();
+
+      // Check if the TF Tensor of the Variable needs to be updated
+      // Update is required in two cases
+      // 1. If any of the outputs of this Op is feeding into a TF Op
+      // 2. If this is a static input to NGraphEncapsulate
+      bool update_tf_tensor =
+          !AreOutputsNGSupported(node) || NodeIsStaticInput(node);
```
sayantan-nervana: Quick ques: What if its ng supported, but is deassigned because its a trivial cluster. In that case it would be "ng supported but feeding to TF".

ngraph_bridge/ngraph_mark_for_clustering.cc
sayantan-nervana: Ok, I see, its in keeping with the already existing `void GetStaticInputs(const Node* node, std::vector<int32>* inputs)`

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
shresthamalik: deassignment happens before we encapsulate, this rewrite happens after encapsulate clusters is called. so we will only see final clusters at this stage

sayantan-nervana: ah, right.

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -27,6 +28,32 @@ namespace tensorflow {
...
+  }
+  return true;
+}
+
+// returns true if this node is a static input to any of its output nodes
+bool NodeIsStaticInput(Node* node) {
+  for (auto edge : node->out_edges()) {
+    if (edge->dst()->IsOp() && !edge->IsControlEdge() &&
+        InputIsStatic(edge->dst(), edge->dst_input())) {
+      NGRAPH_VLOG(5) << "ngraph does not support dst node ";
```
kanvi-nervana: dst node is not supported because it has a static input

shresthamalik: stray comment. will fix it

ngraph_bridge/enable_variable_ops/ngraph_rewrite_for_tracking.cc
```diff
@@ -46,25 +73,22 @@ Status RewriteForTracking(Graph* graph, int graph_id) {
...
     if (itr != REWRITE_REPLACE_OP_MAP.end()) {
-      NGRAPH_VLOG(1) << "Checking: " << DebugNode(node) << " " << node->name();
-
-      bool outputs_ng_supported = true;
+      NGRAPH_VLOG(5) << "Checking: " << DebugNode(node) << " " << node->name();
+
+      // Check if the TF Tensor of the Variable needs to be updated
+      // Update is required in two cases
+      // 1. If any of the outputs of this Op is feeding into a TF Op
+      // 2. If this is a static input to NGraphEncapsulate
```
sayantan-nervana: We can add some more comments here. In case of static input, the Compute fn will read the value from TF. which means if TF is not updated it will get stale values from TF etc.

ngraph_bridge/enable_variable_ops/ngraph_replace_op_utilities.cc
```diff
@@ -30,9 +30,8 @@ namespace ngraph_bridge {
 Status ReplaceOptimizer(Graph* graph, Node* node, Node** replacement,
                         const string replacement_node_name,
                         const string replacement_node_type,
-                        const bool just_looking,
-                        const bool outputs_ng_supported, const int graph_id,
-                        const bool is_backend_set) {
+                        const bool just_looking, const bool update_tf_tensor,
+                        const int graph_id, const bool is_backend_set) {
```
sayantan-nervana: note to self: Earlier `!outputs_ng_supported <-> needs copying`. Now that assumption has changed. In case of var feeding an encapsulate as static input, we have `outputs_ng_supported=true` but `needs copying = true` as well, thus breaking `!outputs_ng_supported <-> needs copying`.

Hence directly passing in a variable named `update_tf_tensor`

ngraph_bridge/ngraph_cluster_manager.cc
```diff
@@ -32,13 +34,26 @@ size_t NGraphClusterManager::NewCluster() {
...
   return idx < s_cluster_graphs.size() ? s_cluster_graphs[idx] : nullptr;
 }
 
 void NGraphClusterManager::EvictAllClusters() { s_cluster_graphs.clear(); }
 
+Status NGraphClusterManager::GetClusterGraph(size_t idx, Graph* graph) {
+  GraphDef* cluster_gdef = NGraphClusterManager::GetClusterGraphDef(idx);
+  if (cluster_gdef == nullptr) {
+    graph = nullptr;  // is this ok?
+    return Status::OK();
```
sayantan-nervana: If we are returning Status, we should return an error status here i think

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -115,9 +115,10 @@ void NGraphEncapsulateOp::CreateParallelExecutor(OpKernelConstruction* ctx,
 
   int cluster_id{-1};
   OP_REQUIRES_OK(ctx, ctx->GetAttr<int>("ngraph_cluster", &cluster_id));
-  graph_def = NGraphClusterManager::GetClusterGraph(cluster_id);
+  OP_REQUIRES_OK(ctx, NGraphClusterManager::GetClusterGraph(
+                          cluster_id, encap_subgraph.get()));
 
-  if (graph_def == nullptr) {
+  if (encap_subgraph.get() == nullptr) {
```
sayantan-nervana: If GetClusterGraph returns a status, then this check can go away. We can still keep it around for "onion defense" :D

ngraph_bridge/ngraph_cluster_manager.cc
shresthamalik: I reverted some changes last friday, which are showing up again. I think while addressing review comments and I changed and pushed from a different repo which did not have the reverted changes. You should not be seeing this function. :)

shresthamalik: restored the PR. No changes to cluster manager

PR #487, opened 2020-02-12
ashahba/Upgrading TensorFlow to v1.15.2

ashahba:
Usually when new releases of `TensorFlow` come out, there are a few releases on the release branch.
For `r1.15` they posted 3 releases so far and `v1.15.2` is the latest and greatest so we should go with that version.

One thing I noticed was that `--config` only supports `v2` on `v1.15.0` but for `v1.15.2` there is support for both `v1` and `v2` so my understanding is in case some day we want to experiment with `v1` vs `v2` config on `v1.15` then `v1.15.2` is the one we should go with.


tools/build_utils.py
```diff
@@ -218,7 +218,7 @@ def build_tensorflow(venv_dir, src_dir, artifacts_dir, target_arch, verbosity):
         "bazel",
         "build",
         "--config=opt",
-        "--config=v2",
+        "--config=v1",
```
sayantan-nervana: What is this config v1 vs v2 here?

ashahba: It's this guy:
https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/compat/v1/config
My understanding is for `TF v1.15.x` we should just stick with `/v1/config` and only use that for `TF v2.x` and better.

ashahba: Surprisingly for `TF v1.15.0` they only supported `--config=v2` 🤔 
But then they realized they need `--config=v1` support and they added it to newer releases of `TF v1.15.x` so hence this change.

PR #488, opened 2020-02-18
yimei/Add support for TF's Log1p op

yimeisun123:


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -553,6 +554,7 @@ const TypeConstraintMap& GetTypeConstraintMap() {
     type_constraint_map["Less"]["T"] = NGraphDTypes();
     type_constraint_map["LessEqual"]["T"] = NGraphDTypes();
     type_constraint_map["Log"]["T"] = NGraphNumericDTypes();
+    type_constraint_map["Log1p"]["T"] = NGraphNumericDTypes();
```
sayantan-nervana: the definition of "T" is real numbers here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt#L20768

We should use `NGraphRealDTypes` as `NGraphNumericDTypes` includes int types as well

test/test_math_ops.cpp
```diff
@@ -1614,6 +1614,28 @@ TEST(MathOps, Log2D) {
...
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Log1p", static_input_indexes, output_datatypes,
+                        sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}  // end of test op Log1p
+
```
sayantan-nervana: can u add a pytest as well: https://www.tensorflow.org/api_docs/python/tf/math/log1p
also please make sure using NGRAPH_TF_LOG_PLACEMENT that log1p cpp op is being created if u use the python api `tf.math.log1p`

test/test_math_ops.cpp
```diff
@@ -1614,6 +1614,28 @@ TEST(MathOps, Log2D) {
   opexecuter.RunTest();
 }  // end of test op Log
 
+TEST(MathOps, Log1p) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 3;
+  vector<float> vals = {0, 0.25, 0.5, 1, 5, 10};
```
sayantan-nervana: can u add some more numbers which have negative numbers in between -1 to 0

and if the input is < -1? how does TF behave and are we doing the same behaviour when running through ng?

test/test_math_ops.cpp
sayantan-nervana: Also, can you please try to discover some TF python tests using this: https://github.com/tensorflow/ngraph-bridge/tree/master/test/python/tensorflow

things like these: https://github.com/tensorflow/tensorflow/blob/9716a8099eafc84c479d49b62e1eb2bbf18af33c/tensorflow/python/ops/math_ops_test.py#L567

ngraph_bridge/ngraph_mark_for_clustering.cc
yimeisun123: ok, will change this data type.

test/test_math_ops.cpp
yimeisun123: @sayantan-nervana 
I tried to add the negative value to the test, i.e., -2, -1, both ngraph and TF have the same result but the test case fails as shown below. The reason is that the comparison in test_utilities.h uses _arg0 == arg1_, this won't work for nan and inf. So, can we just keep the current test data set?
`/home/yimeisun/ngraph-bridge/test/test_utilities.h:183: Failure
Value of: rt
  Actual: false
Expected: true
 TF output -nan
 NG output -nan
/home/yimeisun/ngraph-bridge/test/test_utilities.h:183: Failure
Value of: rt
  Actual: false
Expected: true
 TF output -inf
 NG output -inf
`

test/test_math_ops.cpp
yimeisun123: @sayantan-nervana - I added the TF test cases in the "test/python/tensorflow/python_tests_list.txt " file, did the make install and pip install tf_br whl,  but when I run the test_runner.py or the tf_unittest_runner.py, those test cases are output as INVALID. It seems to me that there are additional update needed when adding the TF's test cases.
Did I miss copying some files or install additional whl? Thanks

test/test_math_ops.cpp
sayantan-nervana: @noim210 was facing the same issue i think, where the equality does not work for inf/nan. Can you plz fix the comparing function as well

test/test_math_ops.cpp
sayantan-nervana: copy the build_cmake/python/dist<whlfile> to ./build_cmake/artifacts

test/test_math_ops.cpp
yimeisun123: Done in PR491.

test/test_math_ops.cpp
yimeisun123: Done. Thanks!

PR #491, opened 2020-02-22
Yimei/Add nan and infinity comparison in test utility - NGTF-2620

yimeisun123:
Handle inf and nan cases

### DIFF
--- a/test/test_math_ops.cpp
+++ b/test/test_math_ops.cpp
@@ -1920,6 +1920,34 @@ TEST(MathOps, RealDivBroadcasting) {
   opexecuter.RunTest();
 }  // end of test op RealDivBroadcasting
 
+// Test op: RealDiv for nan, inf case
+TEST(MathOps, RealDivNonfinite) {
+  Scope root = Scope::NewRootScope();
+  int dim = 3;
+
+  Tensor A(DT_FLOAT, TensorShape({dim}));
+  Tensor B(DT_FLOAT, TensorShape({dim}));
+
+  auto inf = std::numeric_limits<float>::infinity();
+
+  vector<float> dividend_vals = {0, -inf, inf};
+  vector<float> divisor_vals = {0, 1.0, 1.0};
+
+  AssignInputValues(A, dividend_vals);
+  AssignInputValues(B, divisor_vals);
+
+  vector<int> static_input_indexes = {};
+  auto R = ops::RealDiv(root, A, B);
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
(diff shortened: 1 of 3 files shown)


test/test_utilities.cpp
```diff
@@ -191,12 +191,15 @@ void Compare(Tensor& T1, Tensor& T2, float tol) {
...
     auto a = T1_data[k];
     auto b = T2_data[k];
-    if (a == 0) {
-      EXPECT_NEAR(a, b, tol);
-    } else {
-      auto rel = a - b;
-      auto rel_div = std::abs(rel / a);
-      EXPECT_TRUE(rel_div <= tol);
+    if (!(std::isnan(a) && std::isnan(b)) &&
+        !(std::isinf(a) && std::isinf(b))) {
```
sayantan-nervana: In case of nan or inf, we are not comparing them.

For positive and negative infinities, they might be comparable. hence we should check them. Not sure in the case of nan

For unit tests: you could create a tensor and populate it with Nan/inf maybe using this function: https://github.com/tensorflow/ngraph-bridge/blob/02e02434216b017a4879bfe4453312a937ef87a1/test/test_math_ops.cpp#L112

test/test_utilities.cpp
```diff
@@ -191,16 +191,24 @@ void Compare(Tensor& T1, Tensor& T2, float tol) {
...
+      continue;
+    }
+    if (std::isinf(a) && std::isinf(b)) {
+      if ((a < 0 && b < 0) || (a > 0 && b > 0)) {
+        continue;
       } else {
-        auto rel = a - b;
-        auto rel_div = std::abs(rel / a);
-        EXPECT_TRUE(rel_div <= tol);
+        ASSERT_TRUE(false) << "Mismatch inf and -inf";
```
sayantan-nervana: Minor:
We can replace the if-else with:
`EXPECT_TRUE((a < 0 && b < 0) || (a > 0 && b > 0)) << "Mismatch inf and -inf";`

test/test_math_ops.cpp
```diff
@@ -1920,6 +1920,34 @@ TEST(MathOps, RealDivBroadcasting) {
...
+  Scope root = Scope::NewRootScope();
+  int dim = 3;
+
+  Tensor A(DT_FLOAT, TensorShape({dim}));
+  Tensor B(DT_FLOAT, TensorShape({dim}));
+
+  auto inf = std::numeric_limits<float>::infinity();
+
+  vector<float> dividend_vals = {0, -inf, inf};
+  vector<float> divisor_vals = {0, 1.0, 1.0};
```
sayantan-nervana: Note to self: The first case 0/0 tests the nan case

PR #493, opened 2020-02-25
ashahba/Format bazel build command for C++ lib and add config=v2 option

ashahba:
This is needed both for consistency and readability


tools/build_utils.py
```diff
@@ -286,9 +286,16 @@ def build_tensorflow_cc(src_dir, artifacts_dir, target_arch, verbosity):
 
     # Now build the TensorFlow C++ library
     cmd = [
-        "bazel", "build", "--config=opt", "--config=noaws", "--config=nohdfs",
-        "--config=noignite", "--config=nokafka", "--config=nonccl",
-        "//tensorflow:libtensorflow_cc.so.1"
+        "bazel",
+        "build",
+        "--config=opt",
+        "--config=v2",
```
sayantan-nervana: So tf1.15 is ok to build with config=v2?

sayantan-nervana: Saw your reply regarding config=v2 in the other PR. Thanks

ashahba: Yes, it is.

PR #503, opened 2020-04-22
upgrade to recent nGraph

shresthamalik:
Upgrade NGTF to recent nGraph commit close to OV fork
Upgrade some ops to opset3
Disable certain tests


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -1232,16 +1232,18 @@ Status MarkForClustering(Graph* graph, const std::set<string> skip_these_nodes,
...
+      // TFtoNgraphOpMap,
+      //                                         is_supported));
+
+      // if (!is_supported) {
+      //   NGRAPH_VLOG(5) << "TF Op " << node->name() << " of type "
+      //                  << node->type_string()
+      //                  << " is not supported by backend: " <<
+      //                  ng_backend_type;
+      //   break;
+      // }
```
adk9, line 1246: Remove; you don't need to comment this out.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -5404,39 +5404,40 @@ Status Builder::TranslateGraph(
...
+      //     putenv(reset);
+      //     delete[] reset;
+      //   }
+
+      //   return errors::Internal(
+      //       "Found ngraph node ", n->get_name(),
+      //       " which has provenance tag set of size ", num_tags,
+      //       ". Expected all ngraph nodes created in TranslateGraph to have "
+      //       "exactly one provenance tag");
+      // }
```
adk9, line 5440: Please fix or remove.

bazel/mkl_dnn.BUILD
```diff
@@ -40,24 +56,31 @@ config_setting(
...
 
+
 cc_library(
     name = "mkl_dnn",
     srcs = glob([
         "src/common/*.cpp",
         "src/common/*.hpp",
         "src/cpu/*.cpp",
         "src/cpu/*.hpp",
+        "src/cpu/jit_utils/*.cpp",
```
shresthamalik, line 78: Question regarding changes for bazel build. Are these changes being tested anywhere? Since we removed bazel build from being tested on ci.

kanvi-nervana: No, these aren't tested. I wasn't sure what we want to do with the Bazel files wrt upgrading nGraph version. Since there were already changes in the existing PR, I decided to let them be.

shresthamalik: If it is not being tested we should not be upgrading the bazel build as part of the PR. 
The bazel build can be upgraded in a separate PR when we are able to build with latest mkl version. We can make it part of ci then.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -3003,64 +3001,32 @@ static Status TranslateReduceOp(
   std::transform(
       axes.begin(), axes.end(), ng_reduction_axes_vect.begin(),
       [input_rank](int idx) { return idx + (idx < 0 ? (int)input_rank : 0); });
-  ng::AxisSet ng_reduction_axes(ng_reduction_axes_vect);
+  auto ng_reduction_axes = ConstructNgNode<ng::op::Constant>(
```
kanvi-nervana, line 3005: Should this be opset3 ?

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -869,31 +864,20 @@ GetTFToNgOpMap() {
...
+        {"Minimum", {std::make_shared<ngraph::opset3::Minimum>()}},
         {"Mul", {std::make_shared<ngraph::opset3::Multiply>()}},
         {"Neg", {std::make_shared<ngraph::op::Negative>()}},
-        {"OneHot",
-         {std::make_shared<ngraph::op::OneHot>(),
-          std::make_shared<ngraph::op::Convert>(),
-          std::make_shared<ngraph::op::Select>(),
-          std::make_shared<ngraph::op::Reshape>(),
-          std::make_shared<ngraph::op::Broadcast>()}},
+        {"OneHot", {std::make_shared<ngraph::op::OneHot>()}},
```
kanvi-nervana, line 869: Should this be opset3?

ngraph_bridge/ngraph_builder.cc
adk9: Yes.

ngraph_bridge/ngraph_mark_for_clustering.cc
kanvi-nervana: removed

ngraph_bridge/ngraph_builder.cc
kanvi-nervana: removed

bazel/mkl_dnn.BUILD
kanvi-nervana: Done

ngraph_bridge/ngraph_mark_for_clustering.cc
adk9: What I meant was - do you really need to comment this out? What about the opset3 upgrade does it make unsupported on the current backends supported on the main branch?

ngraph_bridge/ngraph_builder.cc
```diff
@@ -538,9 +541,6 @@ static Status TranslateBinaryOp(
   std::shared_ptr<ng::Node> ng_lhs, ng_rhs;
   TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_lhs, &ng_rhs));
 
-  std::tie(ng_lhs, ng_rhs) =
```
shresthamalik, line 541: So we removed explicit broadcasting here but some ops call TranslateBinaryOp that were not upgraded.  Like Power. Do they need separate handling?

kanvi-nervana: We only upgraded ops that were necessary to pass tests at the moment, rest will be done in separate PR's if that's what you are asking.

shresthamalik: Thinking why the tests for the ops that were not upgraded are not failing. Maybe they don't check broadcasting. This op does seem like it can support broadcast since it is doing element-wise Power
https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/pow.cc 

You can add another function similar to this and call that for the upgraded ops, that way it wont break the ops that are not upgraded.

PR #510, opened 2020-04-29
Fix --use_prebuilt_tensorflow flag to use prebuilt TF

adk9:
* Fix `--use_prebuilt_tensorflow` to actually skip downloading TF repo and building TF
* Add a CLI option to build_ngtf.py (``--disable-cpp-api`) to disable building C++ API, unit tests and examples
* Fix test_ngtf.py to be able to run tests


build_ngtf.py
```diff
@@ -301,38 +301,9 @@ def main():
     else:
         if arguments.use_prebuilt_tensorflow:
             print("Using existing TensorFlow")
-            # First download the source. This will create the tensorfow directory as needed
```
shresthamalik, line 304: The use_prebuilt_tensorflow mode, was designed to get the tf repo and built the required libraries so that we can run the cpp tests. With tf 1.15.2 due to gcc/abi issues we might not be able to do it temporarily, the idea was to get it back when tf fixes it.

adk9: And how's that different from not specifying the flag? Based on the usage description [here](https://github.com/tensorflow/ngraph-bridge/blob/master/build_ngtf.py#L98), `use_prebuilt_tensorflow` is supposed to use prebuilt (downloaded or installed) TF instead of building it from source. And because the installed version doesn't distribute the cc libraries, it is supposed to disable CPP tests?

shresthamalik: yea, i guess the description is not updated.

adk9: I'm confused — updated to what? If `--use_prebuilt_tensorflow` were to download and build TF then it shouldn't exist in the first place? These changes make the flag do what it says.

shresthamalik: --use_prebuilt_tensorflow gets TF and builds only certain TF libraries that are required to build and run cpp applications. I think the default way builds a lot more TF libraries. When this mode (--use_prebuilt_tensorflow ) was first added it did not built TF at all, and hence we were not able to run cpp unit tests or cpp applications in examples. Later it was modified to be able to do that and the description might not have been updated. I understand your confusion this mode was extended to do other things one of which was to verify our ability to build and run cpp applications with the gcc/abi pre-built tf was using.

So 2 ways we can proceed:
1. Change the description of the flag
2. Take in your changes. Add another mode (which does what the current mode does) so that when the gcc/abi issues with tf pre-built is resolved we have something to quickly add and test cpp unit tests.

adk9: What are the ABI issues with prebuilt TF? Why don't we add a mode to be able to just fetch and build `libtensorflow_cc` and run C++ tests against it when these ABI issues are resolved? IMO, `--use_prebuilt_tensorflow` should do what it's supposed to do, and it's incredibly handy to have it.

adk9:
> Are the CI builds passing with these changes? Any new CI config to set up to take advantage of the prebuilt-tf option?

Not yet, but they should pass hopefully soon. It would be nice to have a BK job to test building against prebuilt TF.

tools/test_utils.py
```diff
@@ -123,6 +133,7 @@ def run_ngtf_pytests(venv_dir, build_dir):
 
     # Next run the ngraph-tensorflow python tests
     command_executor(["pip", "install", "-U", "pytest"])
+    command_executor(["pip", "install", "-U", "keras"])
```
kanvi-nervana, line 136: This will install keras 2.4

tools/test_utils.py
```diff
@@ -194,7 +205,7 @@ def run_tensorflow_pytests(venv_dir, build_dir, ngraph_tf_src_dir, tf_src_dir):
     print("CURRENT DIR: " + os.getcwd())
 
     print("Patching TensorFlow using: %s" % patch_file)
-    apply_patch(patch_file)
+    apply_patch(patch_file, level=0)
```
kanvi-nervana, line 208: What does the level do?

adk9: Specifies the patch level to use.

tools/test_utils.py
adk9: Fixed.

PR #517, opened 2020-05-15
Make friendly name unique for a node

kanvi-nervana:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -97,7 +97,7 @@ static void SaveNgOp(Builder::OpMap& ng_op_map, const std::string& op_name,
 
 void Builder::SetTracingInfo(const std::string& op_name,
                              const shared_ptr<ng::Node> ng_node) {
-  ng_node->set_friendly_name(op_name);
+  ng_node->set_friendly_name(op_name + "/" + ng_node->get_name());
```
shresthamalik, line 100: use underscore to clearly demarcate where TF name ends and ngraph begins
```suggestion
  ng_node->set_friendly_name(op_name + "_" + ng_node->get_name());
```

kanvi-nervana: _ is also used as part of the friendly name afaik,  but I can change it - ?

PR #521, opened 2020-06-04
Kanvi/Update SquaredDiffernce

kanvi-nervana:
Update SquaredDiffernce to use ngraph::op::SquaredDifference
Upgrade to use opset3 ops


ngraph_bridge/ngraph_builder.cc
```diff
@@ -4570,13 +4570,12 @@ static Status TranslateSquareOp(
...
-  return TranslateBinaryOp(
-      op, static_input_map, ng_op_map, [&op](std::shared_ptr<ng::Node> input1,
-                                             std::shared_ptr<ng::Node> input2) {
-        auto ng_diff =
-            ConstructNgNode<ng::op::Subtract>(op->name(), input1, input2);
-        return ConstructNgNode<ng::op::Multiply>(op->name(), ng_diff, ng_diff);
-      });
+  shared_ptr<ng::Node> input1;
+  shared_ptr<ng::Node> input2;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &input1, &input2));
```
shresthamalik, line 4575: Do we need to use TranslateBinaryOp, it will broadcast the inputs if required. Does our unit test check for broadcasting? If not we could add one.

kanvi-nervana: We have 2 unit tests for this op: with/without broadcasting https://github.com/tensorflow/ngraph-bridge/blob/master/test/test_math_ops.cpp#L2143

PR #524, opened 2020-06-11
Add half-precision datatype support

adk9:
* Add support for DT_HALF TF datatypes
* Minor fix to type constraint map for "OneHot" op


adk9:
@googlebot I consent.

PR #526, opened 2020-06-17
Remove support for backend-specific ops

adk9:
* Remove deprecated backend-specific ops (e.g. CombinedNonMaxSuppression)
* Use opset3 NonMaxSuppression op for NonMaxSuppressionV4
* Remove deprecated backend-specific op for Gather


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -863,6 +862,7 @@ GetTFToNgOpMap() {
         {"Minimum", {std::make_shared<ngraph::opset3::Minimum>()}},
         {"Mul", {std::make_shared<ngraph::opset3::Multiply>()}},
         {"Neg", {std::make_shared<ngraph::op::Negative>()}},
+        {"NonMaxSuppressionV4", {std::make_shared<ngraph::opset3::NonMaxSuppression>(), constant, constant, constant}},
```
kanvi-nervana, line 865: I don't think you need to specify constant multiple times in the map.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2367,25 +2272,10 @@ static Status TranslateGatherV2Op(
...
-    shared_ptr<ng::Node> ng_gather =
-        backend->get_backend_op("Gather", &ng_input, &ng_input_coords, &axis);
-    if (ng_gather == nullptr) {
-      return errors::Internal("In translating GatherV2 op ", op->name(),
-                              " backend could not return valid ngraph node");
-    }
-    Builder::SetTracingInfo(op->name(), ng_gather);
-    SaveNgOp(ng_op_map, op->name(), ng_gather);
-  }
+  auto gather_op = ConstructNgNode<ng::op::Gather>(
```
kanvi-nervana, line 2275: Shouldn't this be opset3?

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2945,27 +2835,16 @@ static Status TranslateNonMaxSuppressionV4Op(
...
-  shared_ptr<ng::Node> ng_nmsv4 = backend->get_backend_op(
-      "NonMaxSuppressionV4", &ng_boxes, &ng_scores,
-      (size_t)(max_output_size[0]), (float)(iou_threshold[0]),
-      (float)score_threshold[0], (bool)pad_to_max_output_size);
-  if (ng_nmsv4 == nullptr) {
-    return errors::Internal("In translating NonMaxSuppressionV4 op ",
-                            op->name(),
-                            " backend could not return valid ngraph node");
-  }
+  // ADK: (FIXME) are the outputs set correctly here?
```
kanvi-nervana, line 2847: Do we have tests for this op? I couldn't find any, so we should probably add one.

ngraph_bridge/ngraph_builder.cc
adk9: The intent of this PR was to primarily remove backend-specific ops, not upgrade any ops. We'll address upgrade of all Gather ops in a separate PR.

ngraph_bridge/ngraph_builder.cc
adk9: None of the existing backends support this op. And I believe that there are TF unit tests that test this op which we'll enable once we've a backend that supports this op.

shresthamalik: With this PR we have a translation that is not tested. Better to add support when it could be tested.

kanvi-nervana: I agree, we should probably only remove the backend specific ops in this PR and add/change support if we can test it.

adk9: `NonMaxSuppressionV4` is not a backend-specific op. I agree that it's untested at the moment but so are several other ops. I'd rather keep it and fix it as soon as the new backend is up (which is going to be very soon) than comment it out temporarily.

PR #527, opened 2020-06-17
Kanvi/Logical_Xor support

kanvi-nervana:
Add support for logical_xor - opset3 op and a python test for it

### DIFF
--- a/ngraph_bridge/ngraph_mark_for_clustering.cc
+++ b/ngraph_bridge/ngraph_mark_for_clustering.cc
@@ -362,6 +362,7 @@ const std::map<std::string, ConfirmationFunction>& GetConfirmationMap() {
     confirmation_function_map["LogicalAnd"] = SimpleConfirmationFunction();
     confirmation_function_map["LogicalNot"] = SimpleConfirmationFunction();
     confirmation_function_map["LogicalOr"] = SimpleConfirmationFunction();
+    confirmation_function_map["LogicalXor"] = SimpleConfirmationFunction();
     confirmation_function_map["MatMul"] = SimpleConfirmationFunction();
     confirmation_function_map["Max"] = SimpleConfirmationFunction();
     confirmation_function_map["Maximum"] = SimpleConfirmationFunction();
@@ -844,6 +845,7 @@ GetTFToNgOpMap() {
         {"LogicalAnd", {std::make_shared<ngraph::opset3::LogicalAnd>()}},
         {"LogicalNot", {std::make_shared<ngraph::op::Not>()}},
         {"LogicalOr", {std::make_shared<ngraph::opset3::LogicalOr>()}},
+        {"LogicalXor", {std::make_shared<ngraph::opset3::LogicalXor>()}},
         {"MatMul",
          {std::make_shared<ngraph::op::Reshape>(),
           std::make_shared<ngraph::op::MatMul>()}},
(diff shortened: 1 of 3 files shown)


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -361,6 +361,7 @@ const std::map<std::string, ConfirmationFunction>& GetConfirmationMap() {
     confirmation_function_map["LogicalAnd"] = SimpleConfirmationFunction();
     confirmation_function_map["LogicalNot"] = SimpleConfirmationFunction();
     confirmation_function_map["LogicalOr"] = SimpleConfirmationFunction();
+    confirmation_function_map["LogicalXor"] = SimpleConfirmationFunction();
```
shresthamalik, line 364: Other maps? Type Constraints?

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -841,6 +842,7 @@ GetTFToNgOpMap() {
         {"LogicalAnd", {std::make_shared<ngraph::opset3::LogicalAnd>()}},
         {"LogicalNot", {std::make_shared<ngraph::op::Not>()}},
         {"LogicalOr", {std::make_shared<ngraph::opset3::LogicalOr>()}},
+        {"LogicalXor", {std::make_shared<ngraph::opset3::LogicalXor>()}},
```
shresthamalik, line 845: Add a cpp unit test too.

test/python/test_logical_xor.py
```diff
@@ -0,0 +1,45 @@
...
+class TestXorOperations(NgraphTest):
+
+    def test_xor(self):
+        input_x = ((False, False, True, True))
+        input_y = ((False, True, False, True))
+        x = tf.compat.v1.placeholder(tf.bool, shape=(4,))
+        y = tf.compat.v1.placeholder(tf.bool, shape=(4,))
+        out = tf.math.logical_xor(x, y)
+
+        def run_test(sess):
```
shresthamalik, line 41: This is probably running through TF. As type constraint is missing. CPP unit test would fail since there is no fallback

ngraph_bridge/ngraph_mark_for_clustering.cc
shresthamalik: Checked its not a strict check. Type Constraint

test/python/test_logical_xor.py
kanvi-nervana: Checked Op placement and it is getting placed on nGraph.

test/python/test_logical_xor.py
```diff
@@ -0,0 +1,45 @@
...
+from common import NgraphTest
+
+
+class TestXorOperations(NgraphTest):
+
+    def test_xor(self):
+        input_x = ((False, False, True, True))
+        input_y = ((False, True, False, True))
+        x = tf.compat.v1.placeholder(tf.bool, shape=(4,))
+        y = tf.compat.v1.placeholder(tf.bool, shape=(4,))
```
shresthamalik, line 38: If the op supports broadcasting, you can add test for broadcasting

kanvi-nervana:
It turns out TF will break LogicalXor - LogicalAnd, LogicalNot, LogicalOr which we already have a translation for hence we do not need to add translation for LogicalXor separately.

PR #529, opened 2020-06-17
Upgrade unary ops to opset3

kanvi-nervana:
Upgrade unary ops to opset3
Abs, Cos, Exp, Log, LogicalNot, Negate, Relu, Reciprocal, Sin, Sign, Sqrt, Tanh
Adds missing tests for some
Upgrades Rsqrt and RsqrtGrad
Adds translation and test for Acos


test/graph_rewrites/enter_in_catalog_test.cc
```diff
@@ -105,9 +105,9 @@ TEST(CatalogTest, SmallGraph2) {
   auto init_value = ops::Const(root, {{1.f, 1.f}, {1.f, 1.f}});
   auto var_assign = ops::Assign(root.WithOpName("Var_Assign"), var, init_value);
 
-  auto acos = ops::Acos(root.WithOpName("Acos"), var);
+  auto asin = ops::Asin(root.WithOpName("Asin"), var);
```
bani-intelaipg, line 108: No Acos any more in Opset3?

bani-intelaipg: Is Asin a new Op?

kanvi-nervana: It's the other way around. Acos is now supported in nGraph so we need to change the test to use an op that is currently not supported.

adk9: `Asin` is supported too but we just don't have a translation for it yet.

kanvi-nervana: Yes, once we add the translation we'll have to update the test to use some other unsupported op for it will fail. I did not realize at the time of making the change that Asin is supported as well.

PR #530, opened 2020-06-18
Update yaml files to install keras 2.3.1

kanvi-nervana:
Keras 2.4 is not compatible with TF 1.15 so we need to install specific keras version 2.3.1 for CI


kanvi-nervana:
I know nG py tests do.

PR #532, opened 2020-06-19
Fix build_tf.py to be able to build TF 2.1+

adk9:
* Fix `build_tf.py` to be able to build versions of TF greater than 2.0\
* Fix the script to not error out when the output dir already exists

### DIFF
--- a/tools/build_utils.py
+++ b/tools/build_utils.py
@@ -220,16 +220,14 @@ def build_tensorflow(tf_version, venv_dir, src_dir, artifacts_dir, target_arch,
         "--config=opt",
         "--config=noaws",
         "--config=nohdfs",
-        "--config=noignite",
-        "--config=nokafka",
         "--config=nonccl",
     ]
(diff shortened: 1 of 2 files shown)


tools/build_utils.py
```diff
@@ -220,16 +220,14 @@ def build_tensorflow(tf_version, venv_dir, src_dir, artifacts_dir, target_arch,
         "--config=opt",
         "--config=noaws",
         "--config=nohdfs",
-        "--config=noignite",
-        "--config=nokafka",
         "--config=nonccl",
     ]
     # Build the python package
-    if (tf_version == "v2.0.0"):
+    if (tf_version.startswith("v2.")):
```
kanvi-nervana, line 226: With these changes the user will be able to build TF2.x using build_tf script and use it with --use_tf_from_location build option, correct?

adk9: Yes.

PR #537, opened 2020-06-23
Fix static build with --use_ngraph_staticlibs

adk9:
* Rename mkl_dnn target to dnnl and link against libdnnl.so rather than libmkldnn.so
* Use `IMPORTED_NO_SONAME` to import mkl as it has no soname
* Add missing dep targets, missing definitions and includes.


adk9:
These changes do not completely fix the static build which is still broken for some type of build configurations. Please do not review just yet and/or merge.

PR #541, opened 2020-06-24
Shrestha/const result remove

shresthamalik:
Possible fix to remove Const->Result in ngraph-function
- Look at Const Nodes that were assigned a different cluster number than there output nodes. Remove them from clustering. Since Const Nodes are data flow ops and have no input they should not affect deadness analysis that was done. 
- Removed Shape Op from clustering. It creates a constant op so keeping it out.
- Added a check in Translate Graph function to detect Const->Result [Just looked at the nGraph apis did not really verify if it works]
- modified the tests to tune to the new clustering


adk9:
Merged in #729

PR #551, opened 2020-06-26
Minor fix for --use_prebuilt_tensorflow when CPP API enabled

adk9:
* Build CPP API first before copying libtensorflow_framework.so to get
around an issue when using --use_prebuilt_tensorflow


kanvi-nervana:
Just curious, what was the issue that was showing up?

adk9:
> Just curious, what was the issue that was showing up?

Building with cpp support and prebuilt TF was failing as git fails to clone in an existing directory.

PR #553, opened 2020-06-27
Use subprocess.Popen in command_executor to catch errors

bani-intelaipg:
Added debug info to analyze Intermittent PyTest error during BuildKite CI pipeline.
Tried Build https://buildkite.com/ngraph/ngtf-cpu-ubuntu/builds/5011
Unit-tested using..
```
import sys, shlex, subprocess
cmd_str = "xls -l"
cmd_list = shlex.split(cmd_str)
#p = subprocess.Popen(args)
#if(subprocess.check_call(cmd_list, stderr=subprocess.STDOUT) != 0):
try:
    retcode = subprocess.call(cmd_list, stderr=subprocess.STDOUT)
    if retcode < 0:
        print("Child was terminated by signal", -retcode, file=sys.stderr)
        exit(retcode)
    elif retcode == 0:
        print("Child returned OK", retcode, file=sys.stderr)
        # OK, continue
    else:
        print("Child returned ERROR", retcode, file=sys.stderr)
        exit(retcode)
except OSError as e:
    print("Execution failed:", e, file=sys.stderr)
    #exit 1 # exit program with error
    raise

print("Back to main program again...")
```


tools/build_utils.py
```diff
@@ -56,8 +56,22 @@ def command_executor(cmd, verbose=False, msg=None, stdout=None, stderr=None):
...
+        elif retcode == 0:
+            print("Child returned OK", retcode, file=sys.stderr)
+            # OK, continue
+        else:
+            print("Child returned ERROR", retcode, file=sys.stderr)
+            exit(retcode)
+    except OSError as e:
+        print("!!! Execution failed !!!", e, file=sys.stderr)
+        #exit 1 # exit program with error
+        raise
```
adk9, line 74: This doesn't seem to use the passed in `stdout` and `stderr` arguments which would mask the build output (and errors) in many cases?

bani-intelaipg: Updated to use Popen (similar to existing scripts), also used params

PR #554, opened 2020-06-29
Refactor NgraphEncapsulateImpl

adk9:
* Move `ngraph_var.{cc,h}` to `enable_variable_ops/`
* Introduce `Compile()` and `GetCompiledString()` methods in
`NgrapEncapsulateImpl` to be able to compile nGraph functions. This fixes leaky abstractions in `ngraph_encapsulate_clusters.cc` and the tests where the backend was called directly to compile an nGraph function.


kanvi-nervana:
The build fails because of incorrect include paths.

ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -92,12 +91,48 @@ Status NGraphEncapsulateImpl::ComputeSignature(
...
+        (st.ok() ? ""
+                 : " Failed to serialize as well with error: " +
+                       st.error_message()));
+  }
+  BackendManager::UnlockBackend(backend_name);
+  return Status::OK();
+}
+
+// Compiles the ngraph function and returns ngraph executable as a string
+Status NGraphEncapsulateImpl::GetCompiledString(
```
kanvi-nervana, line 120: Does this function need locking?

PR #555, opened 2020-06-29
kanvi/upgrade _FusedConv2D to opset3

kanvi-nervana:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -2121,9 +2121,9 @@ static Status TranslateFusedConv2DOp(const Node* op,
                          ng_strides, ng_dilations, ng_padding_below,
                          ng_padding_above);
 
-    ng_conv = ConstructNgNode<ng::op::Convolution>(
+    ng_conv = ConstructNgNode<ng::opset3::Convolution>(
         op->name() + "_FusedConv2D_Conv", ng_input, ng_filter, ng_strides,
-        ng_dilations, ng_padding_below, ng_padding_above);
+        ng_padding_below, ng_padding_above, ng_dilations);
```
adk9, line 2126: If TF pad-type is "VALID", then set it to `ngraph::opset3::PadType::VALID` here.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2133,9 +2133,9 @@ static Status TranslateFusedConv2DOp(const Node* op,
     auto constant_6 = ConstructNgNode<ng::op::Constant>(
```
adk9, line 2133: `ng::opset3::Constant`?

kanvi-nervana: thanks for catching it

PR #556, opened 2020-06-30
Do not suppress stderr when running C++ unit tests

bani-intelaipg:


bani-intelaipg:
This will still help get errors printed, e.g. the gtest_ngtf (cpp unit test) printed error that TF2 so was not found. I suggest we merge it. We will analyze the the other reported error e.g. https://buildkite.com/ngraph/ngtf-cpu-ubuntu-variables/builds/2846#ea985851-bd3d-4dca-aca2-cbe70dd18d3f

PR #566, opened 2020-07-03
enable stderr in run_test() of tf_unittest_runner

bani-intelaipg:


bani-intelaipg:
Also added copy_tf_to_artifacts in build_ngtf to copy TF .so after build in case of build TF2 from source

bani-intelaipg:
We need this in master to prevent spurious CI failures.

bani-intelaipg:
Will create a JIRA issue for reduction_ops_test.SumReductionTest.testFloat32 failure

adk9:
Just as an aside: the test logs look a little verbose because of these changes. For the grappler build, we should wrap the "using backend" message (in `ngraph_optimizer.cc`) with a "NGRAPH_TF_LOG_0_DISABLED" env check as in `ngraph_rewrite_pass.cc`.

PR #571, opened 2020-07-07
Fix build_ngtf script to install TF artifacts when building from source

kanvi-nervana:
remove line that breaks the build - not required


bani-intelaipg:
In case on full/clean build (where building TF from source), in order to run the C++ unit tests exactly like BuildKite CI, we need the libtensorflow_framework.so.2 in ./build_cmake/artifacts/tensorflow/ folder. So, we need to fix it.

adk9:
I'm not sure I understand these changes. Wouldn't we install correctly to the `artifacts_location` when building TF from source? What is the location of installed `libtensorflow_framework.so` when building from source?

PR #572, opened 2020-07-08
Upgrade ops that use broadcast op

kanvi-nervana:
Upgrade FusedMatMul, Fill op which use broadcast
Upgrade some more simple ops - Log1p, ConcatV2, MatMul, Relu6, Sigmoid, Softplus
Add simple C++ test for MatMul


ngraph_bridge/ngraph_builder.cc
```diff
@@ -1634,15 +1634,12 @@ static Status TranslateFillOp(
...
-  for (size_t i = 0; i < dims_vec.size(); ++i) {
-    ng_output_shape[i] = dims_vec[i];
-    ng_axis_set.insert(i);
-  }
-  SaveNgOp(ng_op_map, op->name(),
-           ConstructNgNode<ng::op::Broadcast>(op->name(), ng_value,
-                                              ng_output_shape, ng_axis_set));
+  auto ng_output_shape = ConstructNgNode<ng::opset3::Constant>(
+      op->name(), ng::element::i64, ng::Shape{dims_vec.size()},
+      vector<size_t>(dims_vec.begin(), dims_vec.end()));
```
adk9, line 1639: Just `dims_vec` should work.

PR #575, opened 2020-07-08
Upgrade reshape helpers to use opset3::Transpose

kanvi-nervana:
Upgrade reshape helpers to use transpose when rank is same instead


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -713,14 +710,14 @@ GetTFToNgOpMap() {
       {"ConcatV2", {std::make_shared<ngraph::opset3::Concat>()}},
       {"Const", {constant}},
       {"Conv2D",
-       {std::make_shared<ngraph::op::Reshape>(),
+       {std::make_shared<ngraph::opset3::Transpose>(),
```
adk9, line 713: Would you need a constant here?

PR #580, opened 2020-07-10
Add interpreter backend CI pipeline

adk9:


test/test_nn_ops.cpp
```diff
@@ -659,7 +659,7 @@ TEST(NNOps, MaxPool3DNDHWCValid) {
   }
 }  // end of MaxPool3DNDHWCValid op
 
-TEST(NNOps, QuantizedAvgPoolEvenInput) {
+TEST(NNOps, DISABLED_QuantizedAvgPoolEvenInput) {
```
kanvi-nervana, line 662: Does this fail because of quantizedavgpool not being updated to opset3?

tools/test_utils.py
```diff
@@ -277,9 +277,13 @@ def run_tensorflow_pytests_from_artifacts(backend, ngraph_tf_src_dir,
     # Now run the TensorFlow python tests
     test_src_dir = os.path.join(ngraph_tf_src_dir, "test/python/tensorflow")
     test_script = os.path.join(test_src_dir, "tf_unittest_runner.py")
-    if backend is not None and 'GPU' in backend:
-        test_manifest_file = os.path.join(test_src_dir,
-                                          "python_tests_list_gpu.txt")
+    if backend is not None:
+        if 'GPU' in backend:
```
kanvi-nervana, line 281: This can be removed, since we don't have the support.

test/test_nn_ops.cpp
adk9: Yeah, it could be. I just noticed the changes in your PR. I'll try to rebase against it.

tools/test_utils.py
adk9: Sure, we can have a separate PR that removes GPU support.

test/ci/buildkite/setup.yml
```diff
@@ -1,3 +1,6 @@
+env:
```
bani-intelaipg, line 1: May be put a comment like: Per BuildKite doc, defining an env var at the top of the yaml file will set that var on each of the command steps in the pipeline, and is equivalent to setting the env attribute on every step.

test/ci/buildkite/setup.yml
```diff
@@ -1,3 +1,6 @@
...
+  NGRAPH_TF_BACKEND: ${NGRAPH_TF_BACKEND:-CPU}
```
kanvi-nervana, line 2: Do we need a global setting?

bani-intelaipg:
I think yes, as it sets the proper flag for all the steps in the interpreter CI.

On Jul 10, 2020, at 8:30 PM, kanvi-nervana <notifications@github.com> wrote:

﻿

@kanvi-nervana commented on this pull request.

________________________________

In test/ci/buildkite/setup.yml<https://github.com/tensorflow/ngraph-bridge/pull/580#discussion_r453149097>:

> @@ -1,3 +1,6 @@
+env:
+  NGRAPH_TF_BACKEND: ${NGRAPH_TF_BACKEND:-CPU}


Do we need a global setting?

test/ci/buildkite/test_runner.py
```diff
@@ -115,17 +111,17 @@ def main():
     elif (arguments.test_tf_python):
```
kanvi-nervana, line 111: The bazel test option can be removed

adk9: Cool, maybe remove that and the GPU tests in a separate PR?

test/ci/buildkite/setup.yml
adk9: Yes, this setting is needed to set the backend correctly for all steps in case of each pipeline.

test/python/test_tf2ngraph_shape_hints.py
```diff
@@ -141,6 +136,9 @@ class Testtf2ngraphShapehints(NgraphTest):
     @pytest.mark.skipif(
         not ngraph_bridge.is_grappler_enabled(),
         reason="Requires grappler build for tf2ngraph and AOT")
+    @pytest.mark.skipif(
+        NGTF_BACKEND != "INTERPRETER",
+        reason="Only INTERPRETER backend supports precompilation")
```
kanvi-nervana, line 141: The Interpreter pipeline does not use grappler so it will be skipped on that pipeline and it will be skipped on grappler pipeline because of setting NGRAPH_TF_BACKEND=CPU

kanvi-nervana: I checked and this test was getting skipped. I fixed it here #583

test/ci/buildkite/setup.yml
adk9: I figured that it was obvious from the context since this is in `setup.yml`. Will document the pipeline scripts some more in follow-up PRs.

PR #584, opened 2020-07-14
 Add API to check if openvino is enabled

kanvi-nervana:
Add API to check if OV is enabled and display the same after build finishes.

### DIFF
--- a/build_ngtf.py
+++ b/build_ngtf.py
@@ -508,9 +508,6 @@ def main():
         flag_string_map[arguments.use_grappler_optimizer]
     ])
 
-    if (arguments.build_openvino_backend) or (arguments.use_prebuilt_openvino):
-        ngraph_tf_cmake_flags.extend(["-DNGRAPH_TF_USE_OPENVINO=YES"])
-
     # Now build the bridge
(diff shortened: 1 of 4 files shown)


build_ngtf.py
```diff
@@ -508,6 +508,9 @@ def main():
         flag_string_map[arguments.use_grappler_optimizer]
     ])
 
+    if (arguments.build_openvino_backend) or (arguments.use_prebuilt_openvino):
+        ngraph_tf_cmake_flags.extend(["-DNGRAPH_TF_USE_OPENVINO=YES"])
+
```
adk9, line 513: Not needed. Just use the already defined variable `ENABLE_OPENVINO`.

ngraph_bridge/version.cc
```diff
@@ -67,6 +67,14 @@ bool ngraph_tf_is_grappler_enabled() {
 #endif
 }
 
+bool ngraph_tf_is_openvino_enabled() {
+#if defined(NGRAPH_TF_USE_OPENVINO)
```
adk9, line 71: Use `ENABLE_OPENVINO`.

PR #589, opened 2020-07-16
Dump IE graphs when NGRAPH_TF_DUMP_GRAPHS is set

adk9:
* Dump IE CNN network and weights when NGRAPH_TF_DUMP_GRAPHS is set
* Make sure IE exceptions are caught in case of compilation errors

### DIFF
--- a/ngraph_bridge/ie_executable.cc
+++ b/ngraph_bridge/ie_executable.cc
@@ -44,6 +44,11 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
   m_network = InferenceEngine::CNNNetwork(func);
   set_parameters_and_results(*func);
 
+  if (std::getenv("NGRAPH_TF_DUMP_GRAPHS")) {
+    auto& name = m_network.getName();
+    m_network.serialize(name + ".xml", name + ".bin");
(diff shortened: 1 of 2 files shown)


ngraph_bridge/ie_executable.cc
```diff
@@ -44,6 +44,11 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
   m_network = InferenceEngine::CNNNetwork(func);
   set_parameters_and_results(*func);
 
+  if (std::getenv("NGRAPH_TF_DUMP_GRAPHS")) {
```
kanvi-nervana, line 47: May be we can use another env var for this? NGRAPH_TF_DUMP_GRAPHS dumps a lot of graphs and you may be interested only in IE graphs.

bani-intelaipg: As this flag is used for debugging, more graphs might not be a big issue.

adk9: Yes, too many undocumented env vars get a little unwieldy so I decided to reuse `NGRAPH_TF_DUMP_GRAPHS`.

PR #590, opened 2020-07-20
Add transpose folding pass

adk9:
* Add a pass to combine or cancel out consecutive transposes
* Avoid transpose between convolution and bias-add for fused convs


ngraph_bridge/ngraph_builder.cc
```diff
@@ -4176,6 +4188,13 @@ Status Builder::TranslateGraph(
   //
   ng_function = make_shared<ng::Function>(ng_result_list, ng_parameter_list);
 
+  //
+  // Apply additional passes on the nGraph function here.
+  //
+  ngraph::pass::Manager passes;
+  passes.register_pass<TransposeElimination>();
```
kanvi-nervana, line 4195: Can we rename it to ConsecutiveTransposeElimination which is what it does

adk9: It actually combines or eliminates consecutive transposes. How about `TransposeFolding` then?

ngraph_bridge/ngraph_builder.cc
```diff
@@ -4176,6 +4177,13 @@ Status Builder::TranslateGraph(
...
```
kanvi-nervana, line 4184: Missed changing the name here

PR #592, opened 2020-07-21
Remove "static inputs" for reduction ops

adk9:


adk9:
Abandoning, because we don't have a true dynamic backend fully enabled yet.

adk9:
Closed in favor of #682.

PR #596, opened 2020-07-28
Add transpose sinking pass

adk9:
This PR adds a layout optimization pass which tries to move transposes up or down through the graph, combining and eliminating consecutive transposes in the process. The pass is based on the now-deprecated nGraph reshape sinking pass (`ngraph::pass:ReshapeSinking`) but operates on `opset3::Transpose`.


kanvi-nervana:
@googlebot I consent.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -4129,7 +4129,8 @@ Status Builder::TranslateGraph(
   // Apply additional passes on the nGraph function here.
   //
   ngraph::pass::Manager passes;
-  passes.register_pass<TransposeFolding>();
+  passes.register_pass<pass::TransposeFolding>();
```
kanvi-nervana, line 4132: Removing TransposeFolding would be a separate PR?

test/pass/transpose_sinking_test.cpp
```diff
@@ -0,0 +1,337 @@
...
+TEST(TransposeSinking, PassProperty) {
+  auto pass = std::make_shared<pass::TransposeSinking>();
+  ASSERT_TRUE(
+      pass->get_property(ngraph::pass::PassProperty::REQUIRE_STATIC_SHAPE));
+  ASSERT_FALSE(
+      pass->get_property(ngraph::pass::PassProperty::CHANGE_DYNAMIC_STATE));
+}
+
+TEST(TransposeSinking, EdgeSplitting) {
+  // checks if Transpose is pushed through ng::opset3::Abs, but stopped by
```
kanvi-nervana, line 54: Wondering if this test needs to be changed so the output shape is NHWC instead of NCHW

test/python/test_conv2d.py
```diff
@@ -0,0 +1,60 @@
...
+import pytest
+import platform
+
+import tensorflow as tf
+from tensorflow.python.ops import nn_ops
+from tensorflow.python.ops import array_ops
+from common import NgraphTest
+from tensorflow.python.framework import dtypes
+
+import numpy as np
```
kanvi-nervana, line 32:
```suggestion
import pytest
import platform
import numpy as np

import tensorflow as tf
from tensorflow.python.ops import nn_ops
from tensorflow.python.ops import array_ops
from tensorflow.python.framework import dtypes

from common import NgraphTest
```

kanvi-nervana: Minor: We have a order in which we have the includes etc : 
Standard
TF
nGraph
Bridge

ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -0,0 +1,504 @@
...
+  }
+  stringstream ss;
+  auto transpose = ngraph::as_type_ptr<T>(node);
+  auto const1 = ngraph::as_type_ptr<ngraph::opset3::Constant>(
+      transpose->get_input_node_shared_ptr(1));
+  ss << transpose->get_name() << " ( axis order = "
+     << ngraph::vector_to_string(const1->get_axis_vector_val())
+     << " , shape = " << ngraph::vector_to_string(transpose->get_shape())
+     << " ) "
+     << " , child = " << transpose->get_argument(0)->get_name();
```
kanvi-nervana, line 67: Should we change this to "child of "

ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -0,0 +1,504 @@
...
+  NGRAPH_VLOG(4) << "Default transpose: "
+                 << describe<ngraph::opset3::Transpose>(default_transpose);
+  return default_transpose;
+}
+
+// convert_binary_to_default_order is used when one of the arguments
+// of a binary op isn't in the default format (i.e. nhwc instead of nchw)
+// We have to normalize this other argument to nchw by swimming nchw towards
+// parameters
+// as far as we can
```
kanvi-nervana, line 197: Can you add what right and left mean, helps in understanding?

ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -0,0 +1,504 @@
...
+// materialize pending transposes if they can't be propagated through op
+bool TransposeSinking::run_on_function(shared_ptr<ngraph::Function> f) {
+  TransposeMap reorders, reuse_map;
+  set<shared_ptr<ngraph::Node>> transposes_to_delete;
+  unordered_map<std::string, ngraph::Shape> orig_result_out_shape;
+
+  // STEP 1 : Sink or Swim transposes away for op clusters
+  for (auto n : f->get_ordered_ops()) {
+    NGRAPH_VLOG(4) << "-----Start: Processing node----- " << n->get_name();
+    // collect all Result nodes for a sanity check
```
kanvi-nervana, line 453:
```suggestion
    // collect output shape of all Result nodes for a sanity check
```

ngraph_bridge/ngraph_builder.cc
adk9: Yes.

ngraph_bridge/pass/transpose_sinking.cc
adk9: Changed to `input`.

ngraph_bridge/pass/transpose_sinking.cc
adk9: I reworded the comment to indicate which of the arguments gets transposed/reshaped.

test/pass/transpose_sinking_test.cpp
adk9: Should it matter one way or another?

kanvi-nervana: Not sure, may be it does not since it is a unit test and we aren't comparing with TF.

PR #598, opened 2020-07-28
IE tensor ordering, Const-Result checks, remove isolated Param

bani-intelaipg:
IE tensor ordering, Const-Result checks and removal of isolated Param nodes.
Broken down from previous PR# 595


bani-intelaipg:
Let's wait to try a model which requires these fixes

PR #599, opened 2020-07-28
Add debug helpers related to ngraph::Function

bani-intelaipg:


bani-intelaipg:
ngraph::serialize will work, although for larger clusters the gdb output will be huge. So, these additional ones are meant for custom/verbose debug prints while  paused on a call stack. In gdb, we have some restrictions. Non-elementary (e.g. std::string) argument passing is a problem too.

ngraph_bridge/ngraph_utils.cc
```diff
@@ -350,6 +350,84 @@ Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
...
+          // (result, from) e.g. Result_353->Constant_673,
+          // Result_350->ngraph_output_1
+          m_map_result_to_ngnode.erase(outNode->get_friendly_name());
+          m_map_result_to_ngnode.insert(std::pair<std::string, std::string>(
+              outNode->get_friendly_name(), node->get_friendly_name()));
+        }
+      }
+    }
+  }
+  std::cout << "m_map_result_to_ngnode ==> ";
```
adk9, line 417: "results ==> "

ngraph_bridge/ngraph_utils.h
```diff
@@ -329,6 +329,16 @@ Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank);
 Status NgraphSerialize(const std::string&,
                        const std::shared_ptr<ngraph::Function>&);
 
+#if (CMAKE_BUILD_TYPE == Debug)
+// For gdb debug help...
+// Serialize a ngraph function into a file
+Status gdb_serialize_ngfunc(const char*,
```
adk9, line 335: Prefix helpers with, say, `dbg_` instead of `gdb_` since these debug helpers aren't really specific to gdb per se.

Perhaps
`dbg_print_ngraph_function`
`dbg_serialize_ngraph_function`
so on and so forth.

ngraph_bridge/ngraph_utils.cc
```diff
@@ -350,6 +350,84 @@ Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
   return Status::OK();
 }
 
+#if (CMAKE_BUILD_TYPE == Debug)
+// GDB Debug utility
+void gdb_print_ngfunc_nodes_shptr(const shared_ptr<ngraph::Function>& func) {
+  gdb_print_ngfunc_nodes_funcref(*func);
+}
+
+void gdb_print_ngfunc_nodes_vptr(void* func_addr) {
```
adk9, line 359: Not needed. This is something you can do in a debugger.

bani-intelaipg: I son't know how. Any example?

ngraph_bridge/ngraph_utils.h
bani-intelaipg: gdb branding will make it very clear that these are meant for debugger support, whereas, dbg_xxx could misguide future maintainers as useful debug calls to be used within code - which they can, but we are not compiling them  for non-debug-build. we can probably call these debugger_xxx(...), but that would lengthen the names a bit. Comments?

ngraph_bridge/ngraph_utils.cc
bani-intelaipg: Yes

ngraph_bridge/ngraph_utils.cc
```diff
@@ -350,6 +350,84 @@ Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
...
+  for (const auto& node : func.get_ops()) {
+    // std::cout << "    trying fn(name) = " << node->get_friendly_name() << "("
+    // << node->get_name() << ", " << node->get_type_name() << ")\n";
+    auto outputs = node->outputs();  // std::vector<Output<Node>>
+    if (outputs.size() == 1) {
+      auto const& outHndl = outputs[0];
+      auto const& targetInputHndls =
+          outHndl.get_target_inputs();  // std::set<Input<Node>>
+                                        // get_target_inputs()
+      // std::cout << "        node's outHndl.get_target_inputs().size() = " <<
```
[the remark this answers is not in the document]
bani-intelaipg: Yes

ngraph_bridge/ngraph_utils.cc
adk9: If it only needs a cast, you can do that in gdb.

PR #604, opened 2020-07-31
Upgrade Transpose and Tile ops to opset3

adk9:
This PR upgrades translations for "Transpose" and "Tile" ops to their equivalent nGraph opset3 ops.


adk9:
> There's a disabled Tile test inn tf_exec, may be enable that. Also, did not find a test for Transpose specifically.

There are tests for tile and transpose in `ArrayOps.` amongst other places.

PR #605, opened 2020-07-31
Use "Clamp" op for ReLU6

adk9:
Use fused `ngraph::opset3::Clamp` op for relu6 translation.


adk9:
> Code format fails
> Do we have a test for this?

Yes, I think many of the tests use Relu6, for instance `test/python/test_fusedConv2d.py`.

PR #612, opened 2020-08-10
Add test for sinking through unary op + extra check

kanvi-nervana:
Add an extra check to make sure the output shape of Result is same as it's expected output shape after TS pass

Also, add a simple test for sinking through a unary op

### DIFF
--- a/ngraph_bridge/pass/transpose_sinking.cc
+++ b/ngraph_bridge/pass/transpose_sinking.cc
@@ -444,17 +444,15 @@ static void sink_concat(shared_ptr<ngraph::opset3::Concat> n,
 // materialize pending transposes if they can't be propagated through op
 bool TransposeSinking::run_on_function(shared_ptr<ngraph::Function> f) {
   TransposeMap reorders, reuse_map;
-  ngraph::NodeVector results;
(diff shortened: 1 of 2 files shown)


ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -481,12 +481,19 @@ bool TransposeSinking::run_on_function(shared_ptr<ngraph::Function> f) {
...
     NGRAPH_CHECK(
         r->get_shape() == r->get_input_shape(0) &&
             r->get_element_type() == r->get_argument(0)->get_element_type(),
         " op::Result = ", *r, ", Arg = ", *r->get_argument(0));
+
+    // make sure that after TransposeSinking pass the output_shape for Result
+    // does not change from the expected output_shape before the pass
+    NGRAPH_CHECK(r->get_output_shape(0) == orig_result_out_shape[r->get_name()],
+                 " op::Result = ", *r, " expected output shape = ",
+                 orig_result_out_shape[r->get_name()]);
```
kanvi-nervana, line 496: I can combine the checks, left those separate for ease of understanding.

adk9: I think it looks okay as is.

PR #613, opened 2020-08-11
Add a namespace alias for default opset in use

adk9:
Add tensorflow::ngraph_bridge::{opset/default_opset} to refer to the default nGraph opset
being used.


ngraph_bridge/default_opset.h
```diff
@@ -0,0 +1,32 @@
...
+#define NGRAPH_TF_BRIDGE_DEFAULT_OPSET_H_
+#pragma once
+
+#include "ngraph/opsets/opset3.hpp"
+
+namespace tensorflow {
+namespace ngraph_bridge {
+
+namespace opset = ngraph::opset3;
+namespace default_opset = ngraph::opset3;
```
bani-intelaipg, line 27: When to use which one?

adk9: Keeping both since one is just a shorter alias for the other; It's usually clear from the context that `opset::` refers to the default opset.

PR #616, opened 2020-08-12
Functionalize nullary functions for IE backend

adk9:
Check for nGraph functions passed to IE backend with no parameters. For
such zero arity functions, we try to find a constant (currently only at
the edge) that can be turned into a parameter and hoist it up.


ngraph_bridge/ie_executable.cc
```diff
@@ -40,9 +40,46 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
+            node->get_element_type() == ngraph::element::u64)) {
+        auto constant = ngraph::as_type_ptr<opset::Constant>(node);
+        auto element_type = constant->get_element_type();
+        auto shape = constant->get_shape();
+        auto param = std::make_shared<opset::Parameter>(element_type, shape);
+        ngraph::replace_node(node, param);
+        // nGraph doesn't provide a way to set a parameter to an existing
+        // function, so we clone the function here...
+        func = make_shared<Function>(func->get_results(),
+                                     ParameterVector{param}, func->get_name());
```
bani-intelaipg, line 63: Set friendly name

ngraph_bridge/ie_executable.cc
```diff
@@ -40,9 +40,46 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
     }
   }
 
+  NGRAPH_VLOG(2) << "Checking for function parameters in IE backend";
```
bani-intelaipg, line 43: consider moving this to a small function?

ngraph_bridge/ie_executable.cc
```diff
@@ -40,9 +40,46 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
+        auto param = std::make_shared<opset::Parameter>(element_type, shape);
+        ngraph::replace_node(node, param);
+        // nGraph doesn't provide a way to set a parameter to an existing
+        // function, so we clone the function here...
+        func = make_shared<Function>(func->get_results(),
+                                     ParameterVector{param}, func->get_name());
+        auto ie_tensor = make_shared<IETensor>(element_type, shape);
+        ie_tensor->write(constant->get_data_ptr(),
+                         shape_size(shape) * element_type.size());
+        m_params[param->get_name()] = ie_tensor;
```
bani-intelaipg, line 67: friendly_name?

ngraph_bridge/ie_executable.h
```diff
@@ -42,6 +42,7 @@ class IE_Executable final : public Executable {
  private:
   InferenceEngine::CNNNetwork m_network;
   InferenceEngine::InferRequest m_infer_req;
+  unordered_map<string, shared_ptr<ngraph::runtime::Tensor>> m_params;
```
bani-intelaipg, line 45: consider changing name to m_fake_params

ngraph_bridge/ie_executable.cc
adk9: Why?

ngraph_bridge/ie_executable.h
adk9: Thanks. Changed to `m_hoisted_params` which I think is more descriptive, and added a small note.

ngraph_bridge/ie_executable.cc
```diff
@@ -40,9 +40,46 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
+        ie_tensor->write(constant->get_data_ptr(),
+                         shape_size(shape) * element_type.size());
+        m_hoisted_params[param->get_name()] = ie_tensor;
+        NGRAPH_VLOG(1) << "Converted node " << constant << " to a parameter "
+                       << param;
+        param_replaced = true;
+        break;
+      }
+      if (!param_replaced) {
+        THROW_IE_EXCEPTION << "Found a function with no parameters!";
```
kanvi-nervana, line 74: May be change the error message here to say that "Unable to add a param to a func with no params"

ngraph_bridge/ie_executable.cc
```diff
@@ -63,14 +100,32 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
+    shared_ptr<IETensor> tv;
+    // First check if there were any constants we converted to parameters
+    if (m_hoisted_params.size() > 0) {
+      // We only support one parameter replacement for nullary functions
+      CHECK(m_hoisted_params.size() == 1)
+          << "Multiple input constants were converted to parameters.";
+      CHECK(input_info.size() == 1) << "Expecting one input that was converted "
+                                       "from constant to parameter.";
+      auto input = m_hoisted_params.find(it.first);
+      if (input != m_hoisted_params.end()) {
```
kanvi-nervana, line 120: This check means we were able to convert 1 const to param but the input value does not exist?

ngraph_bridge/ie_executable.cc
kanvi-nervana: Do we need to add a friendly name for a fake* param we add? Would IE need it ?

adk9: I misunderstood the first comment to mean use friendly name instead of name for the function.

IE doesn't seem to need a "friendly name" for parameters, but I'll add one just to be safe.

ngraph_bridge/ie_executable.cc
adk9: Done.

ngraph_bridge/ie_executable.cc
adk9: Yes.

kanvi-nervana: Can that happen?

adk9: Unlikely, but it could happen. The parameter to look up here is coming from IE.

ngraph_bridge/ie_executable.cc
bani-intelaipg: We will postpone this idea for now.

PR #618, opened 2020-08-13
Upgrade Slice and StridedSlice translation

adk9:
Upgrade StridedSlice translation to opset3


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -924,10 +924,7 @@ GetTFToNgOpMap() {
       {"Square", {std::make_shared<opset::Multiply>()}},
       {"SquaredDifference", {std::make_shared<opset::SquaredDifference>()}},
       {"Squeeze", {std::make_shared<opset::Squeeze>(), constant}},
-      {"StridedSlice",
-       {std::make_shared<ngraph::op::Reverse>(),
-        std::make_shared<ngraph::op::Slice>(),
-        std::make_shared<ngraph::op::Reshape>()}},
+      {"StridedSlice", {std::make_shared<opset::StridedSlice>()}},
```
kanvi-nervana, line 927: constant is missing

kanvi-nervana: Because the translation uses it and I thought this map should contain all the ops used for translation or else it complains which is why we mention Constant at other places too.

adk9: This map is only used to check if the op is supported by the backend, right? Besides, Constant is typically supported by all backends so checking it for each op translation is probably redundant.

I added it for the sake of completeness though.

PR #624, opened 2020-08-18
Remove quantization ops

adk9:
This op removes old opset0 translations associated with quantized ops


test/ci/buildkite/test_runner.py
```diff
@@ -88,8 +88,6 @@ def main():
     # Decide which tests to run
     if (arguments.test_cpp):
         test_filter = None
-        if 'INTERPRETER' in backend:
-            test_filter = str("-NNOps.QuantizedAvgPoolEvenInput:ArrayOps.Tile")
```
kanvi-nervana, line 92: Removes the Tile exclusion too

PR #625, opened 2020-08-18
fix const to param conversion

bani-intelaipg:
Fix for commit ea5ab70. The map key must be same as const's friendly name.


adk9:
Thanks, but I'm not sure I fully understand the fix. Does IE use friendly-name as an input's name when one is set?

bani-intelaipg:
The fix only takes care of the saved mapping. The other part of the question re: "Does IE use friendly-name as an input's name when one is set?", all we know is that within OPV/CNN/IE they identify each layer with the friendly_name cominbg from NGraph node - so as long as those are unique, we are fine.

PR #626, opened 2020-08-18
Enable C++ unit-tests for OPV IE:CPU via manifest

bani-intelaipg:
NEW: Specify test patterns using a manifest file via env var NGTF_GTEST_FILE.
To run C++ gtests locally in workspace: .../build_cmake$ NGTF_GTEST_FILE=../test/cpp_tests_list_IECPU.txt LD_LIBRARY_PATH="$LD_LIBRARY_PATH:./ngraph_bridge:./artifacts/lib" ./test/gtest_ngtf
If file is not found, it ignores and runs all/default testcases.


test/main.cpp
```diff
@@ -30,7 +34,51 @@
 
 using namespace std;
 
+static void set_filters_for_backend() {
+  std::string backend = "GENERIC";
```
adk9, line 38: Default to "CPU".

test/ci/buildkite/ngtf-ie-ubuntu.yml
```diff
@@ -12,6 +12,16 @@
       queue: ${BUILDKITE_AGENT_META_DATA_QUEUE}
       name: ${BUILDKITE_AGENT_META_DATA_NAME}
 
+  - wait
+  - command: |
+      NGRAPH_TF_BACKEND=IE:CPU PYTHONPATH=`pwd` python3 test/ci/buildkite/test_runner.py \
```
adk9, line 17: `NGRAPH_TF_BACKEND` should be set to CPU here. The format for backend string is "plugin:device".

test/main.cpp
```diff
@@ -30,7 +34,51 @@
...
+  fstream fs;
+  fs.open(filter_file, ios::in);
+  if (fs.is_open()) {
+    std::string filters = "";
+    std::string line;
+    bool excluded_section = false;
+    while (getline(fs, line)) {
+      line = std::regex_replace(line, std::regex("^\\s+"), "");
+      line = std::regex_replace(line, std::regex("#.*$"), "");
+      line = std::regex_replace(line, std::regex("\\s+$"), "");
```
adk9, line 59: Probably not a big deal, but can combine the whitespace removal into one call using `remove_if`.

test/main.cpp
```diff
@@ -30,7 +34,51 @@
...
+  std::string backend = "GENERIC";
+  char* pBE = getenv("NGRAPH_TF_BACKEND");
+  if (pBE) {
+    backend = std::string(pBE);
+    // backend.replace(backend.begin(), backend.end(), ":", "");
+    backend = std::regex_replace(backend, std::regex(":"), "");
+  }
+  std::string this_filepath = __FILE__;
+  this_filepath =
+      std::regex_replace(this_filepath, std::regex("^(.*)/.*$"), "$1");
```
adk9, line 47: Use `dirname` or something similar.

test/main.cpp
```diff
@@ -30,7 +34,51 @@
...
+    }
+    fs.close();
+    if (filters.back() == ':') filters.pop_back();  // remove last :
+
+    ::testing::GTEST_FLAG(filter) = filters;
+  }
+}
+
 int main(int argc, char** argv) {
+  set_filters_for_backend();
```
adk9, line 80: I think it's probably better to be explicit about the manifest file here, or perhaps find a way to distinguish between which "bridge backend" is active.

test/main.cpp
bani-intelaipg: So, do we want to create all flavors of the cpp_tests_list_<XXX>.txt file in this PR?

test/ci/buildkite/ngtf-ie-ubuntu.yml
bani-intelaipg: If we set it to CPU for IE build, some tests will fail. e.g. Device with "" name is not registered in the InferenceEngine

test/main.cpp
bani-intelaipg: First I strip off leading spaces, then all comments, then all remaining trailing spaces

bani-intelaipg:
NEW: Specify test patterns using a manifest file via env var NGTF_GTEST_FILE.
To run C++ gtests locally in workspace: `.../build_cmake$ NGTF_GTEST_FILE=../test/cpp_tests_list_IECPU.txt LD_LIBRARY_PATH="$LD_LIBRARY_PATH:./ngraph_bridge:./artifacts/lib" ./test/gtest_ngtf`
If file is not found, it ignores and runs all/default testcases.

test/opexecuter.cpp
```diff
@@ -116,8 +116,11 @@ void OpExecuter::ExecuteOnNGraph(vector<Tensor>& ngraph_outputs) {
   TF_CHECK_OK(BackendManager::GetCurrentlySetBackendName(&backend_name));
   tf::SessionOptions options = GetSessionOptions(backend_name);
   ClientSession session(tf_scope_, options);
-  TF_CHECK_OK(session.Run(sess_run_fetchoutputs_, &ngraph_outputs))
-      << "Failed to run opexecutor on NGTF";
+  try {
+    session.Run(sess_run_fetchoutputs_, &ngraph_outputs);
+  } catch (...) {
+    EXPECT_TRUE(false);
```
adk9, line 122: Why is this needed?

bani-intelaipg: If exception happens while running a test in IE, our gtest execution is aborted. This fix allows to amrk that as failure and move on to try next tests in the suite.

test/main.cpp
```diff
@@ -30,7 +34,42 @@
 
 using namespace std;
 
+static void set_filters_from_file() {
+  char* pEnv = getenv("NGTF_GTEST_FILE");
```
kanvi-nervana, line 38: Minor: Change to snake case for var's

PR #627, opened 2020-08-18
Fix path of downloaded pretrained model when running CI RN50 job

adk9:
* Download the pretrained model outside of the git working directory to avoid having to download it again for each job in the CI pipeline
* Remove duplicated unused method

### DIFF
--- a/tools/test_utils.py
+++ b/tools/test_utils.py
@@ -240,6 +240,7 @@ def run_resnet50_from_artifacts(ngraph_tf_src_dir, artifact_dir, batch_size,
 
     os.chdir(root_pwd)
 
+
 def run_resnet50_infer_from_artifacts(artifact_dir, batch_size, iterations):
     root_pwd = os.getcwd()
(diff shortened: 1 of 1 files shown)


tools/test_utils.py
```diff
@@ -348,7 +250,8 @@ def run_resnet50_infer_from_artifacts(artifact_dir, batch_size, iterations):
         install_ngraph_bridge(artifact_dir)
 
     # Check/download pretrained model
-    pretrained_models_dir = root_pwd + '/pretrained_models'
+    pretrained_models_dir = os.path.abspath(
```
kanvi-nervana, line 253: So you have the model downloaded once on each machine? In case some changes are made, we'll need to manually download or it'll happen automatically?

adk9: The script downloads the pretrained model required for the test(s) if it doesn't exist already. This PR only fixes the location of the download.

PR #629, opened 2020-08-18
Use nGraph function to determine I/O tensor order for IE backend

adk9:
The iteration order of inputs and outputs isn't necessarily guaranteed to be the same for a nGraph function and its corresponding IE CNN network. This PR uses the nGraph function order for parameters and results to set the inputs and outputs specified to the `call()` method of IE backend.


ngraph_bridge/ie_executable.h
```diff
@@ -42,9 +44,8 @@ class IE_Executable final : public Executable {
  private:
   InferenceEngine::CNNNetwork m_network;
   InferenceEngine::InferRequest m_infer_req;
-  // This map holds the parameters we insert for functions with no input
-  // parameters
-  unordered_map<string, shared_ptr<ngraph::runtime::Tensor>> m_hoisted_params;
+  TensorMap m_inputs;
```
bani-intelaipg, line 47: Add a comment on the purpose and content structure of this member

ngraph_bridge/ie_executable.cc
```diff
@@ -91,59 +91,56 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
-
-  // Create infer request
   m_infer_req = exe_network.CreateInferRequest();
 }
 
 bool IE_Executable::call(const vector<shared_ptr<runtime::Tensor>>& outputs,
                          const vector<shared_ptr<runtime::Tensor>>& inputs) {
   InferenceEngine::InputsDataMap input_info = m_network.getInputsInfo();
-  if (input_info.size() != (inputs.size() + m_hoisted_params.size())) {
+  if (input_info.size() != (inputs.size() + m_inputs.size())) {
```
bani-intelaipg, line 103: This check looks a bit confusing. May be some comment will help.

ngraph_bridge/ie_executable.h
```diff
@@ -33,6 +33,8 @@ namespace ngraph_bridge {
 // A Inference Engine executable object produced by compiling an nGraph
 // function.
 class IE_Executable final : public Executable {
+  using TensorMap = unordered_map<string, shared_ptr<ngraph::runtime::Tensor>>;
```
kanvi-nervana, line 36: May be add a comment specifying what the map is?

ngraph_bridge/ie_executable.cc
```diff
@@ -91,59 +91,56 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
-                           << it.first;
-      }
-    } else {
-      tv = static_pointer_cast<IETensor>(inputs[i]);
-    }
-    m_infer_req.SetBlob(it.first, tv->get_blob());
-    i++;
+  auto func = m_network.getFunction();
+  auto parameters = func->get_parameters();
+  for (int i = 0; i < inputs.size(); i++) {
```
bani-intelaipg, line 110: Why do we need to do this in every tensor execution/call?

ngraph_bridge/ie_executable.cc
```diff
@@ -91,59 +91,56 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
   if (output_info.size() != outputs.size()) {
     THROW_IE_EXCEPTION
         << "Function outputs number differ from number of given outputs";
   }
 
-  i = 0;
-  for (const auto& it : output_info) {
-    shared_ptr<IETensor> tv = static_pointer_cast<IETensor>(outputs[i]);
+  auto results = func->get_results();
+  for (int i = 0; i < outputs.size(); i++) {
```
bani-intelaipg, line 121: Why do we need this for every call?

ngraph_bridge/ie_executable.cc
```diff
@@ -91,59 +91,56 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
+      return static_pointer_cast<IETensor>((*val).second);
+    } else {
+      THROW_IE_EXCEPTION << "Tensor not found for i/o layer " << key;
+    }
+  };
+
+  //  Prepare input blobs
+  for (const auto& it : m_network.getInputsInfo()) {
+    shared_ptr<IETensor> tv = get_tensor(it.first, m_inputs);
+    m_infer_req.SetBlob(it.first, tv->get_blob());
```
bani-intelaipg, line 137: Let's measure and check if we see any performance hit before & after these changes.

ngraph_bridge/ie_executable.h
adk9: Isn't it self-explanatory? And this is just a shorthand alias, not even a class member.

ngraph_bridge/ie_executable.cc
adk9: There will be a different output vector for each call.

ngraph_bridge/ie_executable.cc
adk9: There will be a different input vector for each call.

ngraph_bridge/ie_executable.cc
adk9: Yeah, agreed. Will add a comment here.

ngraph_bridge/ie_executable.h
adk9: Isn't this self-explanatory too? These are inputs and outputs to the executable.

ngraph_bridge/ie_executable.cc
adk9: Won't we need an example with lots of inputs or outputs to see any significant difference in performance? In any case, this is on the critical path and should ideally avoid multiple traversals. Let me see if these loops can be fused with the ones above.

ngraph_bridge/ie_executable.cc
```diff
@@ -91,59 +92,47 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
   NGRAPH_VLOG(2) << "Loading IE CNN network to device " << m_device;
```
bani-intelaipg, line 93: Wed can possibly call a InfoSaveInOutIndexMaps() here to save the map

ngraph_bridge/ie_executable.cc
bani-intelaipg: True, but for a given IE_Executable object, the call param/result names are fixed always.

ngraph_bridge/ie_executable.cc
adk9: For what? There's no need for a map.

ngraph_bridge/ie_executable.cc
adk9: But the map values will change and are provided in this call? So what's the point of a map?

PR #630, opened 2020-08-18
Use nGraph ShapeOf op for translating TF Shape op

kanvi-nervana:


adk9:
Would this need a "static input"?

kanvi-nervana:
> Would this need a "static input"?

I don't think so

adk9:
> > Would this need a "static input"?
> 
> I don't think so

Ok, but it changes the semantics of this op from being static to dynamic. I'm thinking of a simple example with partial or dynamic shapes that could potentially fail because of this; something like:

```
input = tf.placeholder(tf.float32, shape=(2, 2, None, 2))
shape = tf.shape(input)[1]
```

PR #632, opened 2020-08-19
Remove quantization ops from IE tests and clustering

adk9:
This PR removes left-over quantization tests from the IE unit test manifest and quantization ops from the mark-for-clustering phases.


kanvi-nervana:
Does this PR remove the left over quantization code?

adk9:
> Does this PR remove the left over quantization code?

Yes. And it also removes quantization unit tests from the newly added IE unit test manifest.

PR #633, opened 2020-08-20
Grouped Sum Op-Tests into sets using GTest Fixture

bani-intelaipg:


test/cpp_tests_list_IECPU.txt
```diff
@@ -21,6 +21,7 @@ EncapsulateOp.AllocateNGInputTensors
 EncapsulateOp.AllocateNGOutputTensors
 OpByOpCapability.Backend
 MathOps.Sum
+MathOpsSumFixture.FullSet
```
adk9, line 24: I just noticed that this wasn't in the `MathOps` namespace. Can you make sure it belongs to the `MathOps.Sum.*` namespace, even temporarily?

bani-intelaipg: I don't know of a way. First arg of TEST_F has to be a class. The MathOpsSumFixture class is specific for Sum related tests, so can't really name it MathOps.

adk9: Ok, it's a little odd but I understand that it's just an example. We should use TEST_P for parameterized tests.

PR #635, opened 2020-08-21
Short-circuit trivial functions for IE backend

adk9:
* ~~Enable constant-folding pass for all backends~~
* Added short-circuiting for trivial functions in IE backend
  * Trivial functions include constant-folded (Const -> Result) and identity (Parameter -> Result) functions
* Enabled more C++ and Python unit tests


ngraph_bridge/ie_executable.cc
```diff
@@ -141,5 +154,36 @@ bool IE_Executable::call(const vector<shared_ptr<runtime::Tensor>>& outputs,
...
 }
+
+bool IE_Executable::call_trivial(
+    const vector<shared_ptr<runtime::Tensor>>& outputs,
+    const vector<shared_ptr<runtime::Tensor>>& inputs) {
+  // Do basic sanity checking first
+  if (!outputs.size()) {
+    THROW_IE_EXCEPTION << "Called trivial IE function with no outputs";
+  }
+  if (outputs.size() && outputs.size() != 1) {
```
kanvi-nervana, line 167: Shouldn't outputs.size() != 1 be enough?

ngraph_bridge/ie_executable.cc
```diff
@@ -141,5 +154,36 @@ bool IE_Executable::call(const vector<shared_ptr<runtime::Tensor>>& outputs,
...
+    THROW_IE_EXCEPTION << "Called trivial IE function with multiple outputs";
+  }
+  if (!inputs.size() && !m_hoisted_params.size()) {
+    THROW_IE_EXCEPTION
+        << "Called trivial IE function with no inputs or hoisted params";
+  }
+  if (inputs.size() && inputs.size() != 1) {
```
kanvi-nervana, line 174: Shouldn't inputs.size() != 1 be enough?

ngraph_bridge/ie_executable.cc
```diff
@@ -81,6 +81,15 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
 
   set_parameters_and_results(*func);
 
+  NGRAPH_VLOG(2) << "Checking for trivial functions in IE backend";
+  auto nodes = func->get_ordered_ops();
+  if (nodes.size() == 2 && ngraph::is_type<opset::Parameter>(nodes[0]) &&
+      ngraph::is_type<opset::Result>(nodes[1])) {
+    auto param = ngraph::as_type_ptr<opset::Parameter>(nodes[0]);
```
kanvi-nervana, line 88: Is the param used later?

ngraph_bridge/ie_executable.h
```diff
@@ -40,11 +40,18 @@ class IE_Executable final : public Executable {
...
+  bool call_trivial(const vector<shared_ptr<ngraph::runtime::Tensor>>& outputs,
+                    const vector<shared_ptr<ngraph::runtime::Tensor>>& inputs);
+
   InferenceEngine::CNNNetwork m_network;
   InferenceEngine::InferRequest m_infer_req;
+  string m_device;
   // This holds the parameters we insert for functions with no input parameters
   vector<pair<string, shared_ptr<ngraph::runtime::Tensor>>> m_hoisted_params;
-  string m_device;
+  // This keeps track of whether the originial function was either
```
kanvi-nervana, line 51:
```suggestion
  // This keeps track of whether the original function was either
```

ngraph_bridge/ie_executable.h
```diff
@@ -40,11 +40,18 @@ class IE_Executable final : public Executable {
...
+  // constant-folded
+  // or was simply a trivial identity function
```
kanvi-nervana, line 53: I am a little confused with the comment - won't constant folding lead to trivial functions?

ngraph_bridge/ie_executable.cc
adk9: Yes, fixed.

ngraph_bridge/ie_executable.cc
adk9: No, because there could be no inputs when there's a hoisted parameter.

ngraph_bridge/ie_executable.cc
adk9: No, deleted.

ngraph_bridge/ie_executable.h
adk9: Yes, but not vice versa. Identity functions won't be constant-folded.

kanvi-nervana: So this bool is for keeping track of the trivial identity functions?

adk9: It keeps track of both.

ngraph_bridge/ie_executable.h
```diff
@@ -40,11 +40,17 @@ class IE_Executable final : public Executable {
             const vector<shared_ptr<ngraph::runtime::Tensor>>& inputs) final;
 
  private:
+  bool call_trivial(const vector<shared_ptr<ngraph::runtime::Tensor>>& outputs,
```
bani-intelaipg, line 43: Great!

test/cpp_tests_list_IECPU.txt
```diff
@@ -7,85 +7,32 @@
...
-ArrayOps.Rank
-ArrayOps.ScatterNd1D
-ArrayOps.ScatterNdRepeatIndices
-ArrayOps.ScatterNdComplex
-ArrayOps.ScatterNd3D
+MathOps.Cumsum
+MathOps.SquaredDifference
+MathOps.SquaredDifferenceBroadcasting
+MathOps.Mod
+NNOps.Conv2DBackpropInputNCHWSame
```
bani-intelaipg, line 23: So, these NNOps.conv* tests were passing earlier. Are these failing now due to const folding?

test/cpp_tests_list_IECPU.txt
```diff
@@ -7,85 +7,32 @@
...
+NNOps.Conv2DBackpropInputNHWCWithDilation
+NNOps.Conv3DNDHWCSame
+ArrayOps.DepthToSpaceNHWC
+ArrayOps.Pad
 ArrayOps.Shape2D
-ArrayOps.Shape3D
 ArrayOps.SizeOpAttrsChange
-ArrayOps.SizeOpDefault
 ArrayOps.Slice
+ArrayOps.SpaceToDepthNHWC
```
bani-intelaipg, line 35: Is this a new failure too due to const folding?

ngraph_bridge/ie_executable.cc
```diff
@@ -141,5 +155,33 @@ bool IE_Executable::call(const vector<shared_ptr<runtime::Tensor>>& outputs,
...
+    THROW_IE_EXCEPTION << "Called trivial IE function with multiple inputs";
+  }
+  if (m_hoisted_params.size() && m_hoisted_params.size() != 1) {
+    THROW_IE_EXCEPTION
+        << "Called trivial IE function with multiple hoisted params";
+  }
+
+  auto input = m_hoisted_params.size() ? m_hoisted_params[0].second : inputs[0];
+  auto size = input->get_size_in_bytes();
+  unsigned char* buf_ptr = new unsigned char[size];
```
bani-intelaipg, line 180: Alloc/Dealloc in every iteration will slow down inference. Can we optimize with saving a map of name->buffer, along with sanity-check size every time and possibly some heuristics to purge LRU cache if we cross a threshold.

test/cpp_tests_list_IECPU.txt
adk9: Yes, but what's odd is that the failure is not seen when these tests are run individually.

test/cpp_tests_list_IECPU.txt
adk9: Yes.

test/cpp_tests_list_IECPU.txt
```diff
@@ -7,85 +7,32 @@
...
-NNOps.SoftmaxZeroDimTest2
-ArrayOps.Fill
-ArrayOps.GatherNd2D
-ArrayOps.GatherNd3D
-ArrayOps.Rank
-ArrayOps.ScatterNd1D
-ArrayOps.ScatterNdRepeatIndices
-ArrayOps.ScatterNdComplex
-ArrayOps.ScatterNd3D
+MathOps.Cumsum
```
bani-intelaipg, line 19: MathOps.Cumsum was passing earlier.

ngraph_bridge/ie_executable.cc
adk9: Um, what? We could eliminate the temporary but the input and output tensors could be allocated on different devices.

test/cpp_tests_list_IECPU.txt
bani-intelaipg: If that is reproducible, we should investigate (now or later).

test/cpp_tests_list_IECPU.txt
adk9: Yes.

ngraph_bridge/ie_executable.cc
bani-intelaipg: Yes my comment was about the temp buf

adk9: Ah, so some form of buffer caching? I think it's probably an overkill to consider it at this moment for short-circuiting trivial functions.

bani-intelaipg: Ok for now, as our target models won't probably have such Const->Result disjoint subgraphs in a ng-func. In GATK type model we saw plenty of these patterns. We can monitor performance and take action later.

ngraph_bridge/ie_executable.cc
```diff
@@ -40,6 +40,21 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
     }
   }
 
+  NGRAPH_VLOG(2) << "Checking for trivial functions in IE backend";
+  bool trivial_fn = true;
+  for (auto results : func->get_results()) {
```
bani-intelaipg, line 45: minor: result? or results?

adk9: Fixed.

ngraph_bridge/ie_executable.cc
```diff
@@ -141,5 +162,37 @@ bool IE_Executable::call(const vector<shared_ptr<runtime::Tensor>>& outputs,
...
+      inputs[index]->read(buf_ptr, size);
+      outputs[0]->write(buf_ptr, size);
+      delete buf_ptr;
+    } else if (ngraph::is_type<opset::Constant>(parent)) {
+      auto constant = ngraph::as_type_ptr<opset::Constant>(parent);
+      outputs[i]->write(constant->get_data_ptr(),
+                        shape_size(constant->get_shape()) *
+                            constant->get_element_type().size());
+    } else {
+      THROW_IE_EXCEPTION << "Expected constant or parameter feeding to a "
```
kanvi-nervana, line 191: This will not happen, right? The call_trivial will only be called if the input is a constant or a param to a result.

test/cpp_tests_list_IECPU.txt
```diff
@@ -7,92 +7,11 @@
...
-tf_exec.BatchMatMul
-EncapsulateOp.GetNgExecutable
+BackendManager.SetBackend
+BackendManager.GetCurrentlySetBackendName
+BackendManager.CanCreateBackend
 OpByOpCapability.Backend
-MathOps.Sum
-MathOpsSumFixture.FullSet
-MathOps.Mean
-MathOps.ArgMaxNeg
```
kanvi-nervana, line 24: How are ArgMax, ArgMin, Atan2, BatchMatmul working with IE, these aren't upgraded to opset3?

bani-intelaipg: These tests + CF-pass are creating C->R, so they are not actually going to IE

ngraph_bridge/ie_executable.cc
adk9: Yes, it won't happen – it's just overly defensive programming.

test/cpp_tests_list_IECPU.txt
kanvi-nervana: So there may be other tests doing the same thing but we actually want those to be tested on IE

adk9: IE seemingly supports some pre-opset3 ops using that same path.

adk9: Disabled the constant-folding pass for now.

PR #636, opened 2020-08-23
Enable Python unit tests for IE backend

bani-intelaipg:
* Added env var `PYTEST_SKIPFILTERS` to specify unit tests to skip
* Enabled pytests for IE backend

```
Example usage:
To run like CI: (venv-tf-py3) .../SRCROOT$ PYTEST_SKIPFILTERS=test/python/filters_IECPU.py NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS=1 NGRAPH_TF_LOG_PLACEMENT=1 PYTHONPATH=`pwd`:`pwd`/examples:`pwd`/tools:`pwd`/examples/mnist python3 test/ci/buildkite/test_runner.py --artifacts ./build_cmake/artifacts/ --test_python 

To run in local sandbox: (venv-tf-py3) .../SRCROOT/test/python$ PYTEST_SKIPFILTERS=filters_IECPU.py NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS=1 ROOT=../.. PYTHONPATH=$ROOT:$ROOT/test/python:$ROOT/tools:$ROOT/examples:$ROOT/examples/mnist python -m pytest
```


tools/test_utils.py
```diff
@@ -95,6 +95,11 @@ def run_ngtf_pytests_from_artifacts(artifacts_dir):
     if not os.path.isdir(test_dir):
```
adk9, line 95: Changes to this file shouldn't be needed.

test/python/filters_IECPU.py
```diff
@@ -0,0 +1,147 @@
+# ==============================================================================
```
adk9, line 1: Can we turn this into a .txt manifest file just for consistency's sake?

test/ci/buildkite/ngtf-ie-ubuntu.yml
```diff
@@ -34,6 +34,18 @@
       queue: ${BUILDKITE_AGENT_META_DATA_QUEUE}
       name: ${BUILDKITE_AGENT_META_DATA_NAME}
 
+  - wait
+  - command: |
+      source /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/venv/bin/activate 
+      PYTEST_SKIPFILTERS=`pwd`/test/python/filters_IECPU.py PYTHONPATH=`pwd`:`pwd`/tools:`pwd`/examples:`pwd`/examples/mnist python3 test/ci/buildkite/test_runner.py \
```
adk9, line 40: I'm wondering if it makes sense to use the same env-var for exclusion manifest for both gtests and pytests or at least normalize it so that it's easier to remember?

tools/test_utils.py
bani-intelaipg: There are chdir(...) in scripts, so need to convert to absolute path early in the script-chain, in case user specifies relative path.

test/ci/buildkite/ngtf-ie-ubuntu.yml
bani-intelaipg: In gtest case, the file allows both inclusion and exclusion patterns. In this case we just specify skipped ones. Also, mostly these will be used via CI which won't be changed frequently.

test/python/filters_IECPU.py
bani-intelaipg: With .py extension it also gets automatically copied to artifacts folder during install. Also, while editing/adding content the vscode-IDE will highlight any syntax errors quickly.

tools/test_utils.py
adk9: The test scripts should convert to absolute path if they need to. Please avoid adding such special handling in the CI test runner.

tools/test_utils.py
```diff
@@ -241,6 +246,104 @@ def run_resnet50_from_artifacts(ngraph_tf_src_dir, artifact_dir, batch_size,
     os.chdir(root_pwd)
```
adk9, line 248: Please remove the following changes.

test/python/filters_IECPU.py
adk9: Manifest files don't need syntax highlighting.

test/ci/buildkite/ngtf-ie-ubuntu.yml
adk9: Too many undocumented env vars to remember. Why don't we be consistent across the different tests we have?

bani-intelaipg: Named Bridge and TF pytest skip filter env vars similar: PYTEST_SKIPFILTERS and TFPYTEST_SKIPFILTERS

tools/test_utils.py
bani-intelaipg: OK

tools/test_utils.py
bani-intelaipg: ok

test/ci/buildkite/ngtf-ie-ubuntu.yml
adk9: Yeah, the naming can be improved but this is OK for now.

PR #638, opened 2020-08-24
Eliminate dummy shapes in transpose-sinking

adk9:
This PR eliminates the usage of fake shapes for input arguments when sinking concat and pad in transpose-sinking optimization pass.


adk9:
Duplicate of #637.

PR #640, opened 2020-08-25
Enable TF Python Tests for IE; Unify manifest filenames

bani-intelaipg:
* Enable TF Python Tests with NGRAPH_TF_TEST_MANIFEST which allows both included & excluded test name/patterns
* Also, now we will run/launch each test in separate process to handle possible crashes.

To mimic CI in local sandbox: (venv-tf-py3) .../SRCROOT$   NGRAPH_TF_TEST_MANIFEST=\`pwd\`/test/python/tensorflow/tests_linux_ie_cpu.txt  NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS=1 PYTHONPATH=\`pwd\` python3 test/ci/buildkite/test_runner.py --artifacts ./build_cmake/artifacts/ --test_tf_python


adk9:
Our TF Python unit tests already have a manifest. Can we please use that for IE?

bani-intelaipg:
> Our TF Python unit tests already have a manifest. Can we please use that for IE?
There is a python_tests_list.txt which is for generic/CPU targets. We read that one as we currently do, then apply any skip-filters on top (via the SKIP manifest). Let me know.

adk9:
The manifest for IE CPU shouldn't depend on the manifest for nGraph CPU, and it's confusing to have both a run and a skip manifest.

bani-intelaipg:
> The manifest for IE CPU shouldn't depend on the manifest for nGraph CPU, and it's confusing to have both a run and a skip manifest.

We could use an independent manifest for IECPU, but then we have to ensure that we add newly enabled tf-py-tests  in all the necessary manifest. Is that OK?

adk9:
> > The manifest for IE CPU shouldn't depend on the manifest for nGraph CPU, and it's confusing to have both a run and a skip manifest.
> 
> We could use an independent manifest for IECPU, but then we have to ensure that we add newly enabled tf-py-tests in all other manifests. Is that OK? Having a common generic list allows that list to be enabled for all with optional separate skiplists for different purpose.

Having a common run manifest is OK, but can we create one that's shared by all backend tests rather than using the nGraph CPU manifest as the common manifest?

bani-intelaipg:
> > > The manifest for IE CPU shouldn't depend on the manifest for nGraph CPU, and it's confusing to have both a run and a skip manifest.
> > 
> > 
> > We could use an independent manifest for IECPU, but then we have to ensure that we add newly enabled tf-py-tests in all other manifests. Is that OK? Having a common generic list allows that list to be enabled for all with optional separate skiplists for different purpose.
> 
> Having a common run manifest is OK, but can we create one that's shared by all backend tests rather than using the nGraph CPU manifest as the common manifest?

We already have a common one: python_tests_list.txt. It has no mention of ng-cpu. (But folks also later created _int and _mac flavors for different build platforms/backends) Do you want a new common for all IE, say, python_tests_list_IE.txt? And then have skip_IECPU.txt, skip_IEGPU.txt etc?

adk9:
> > > > The manifest for IE CPU shouldn't depend on the manifest for nGraph CPU, and it's confusing to have both a run and a skip manifest.
> > > 
> > > 
> > > We could use an independent manifest for IECPU, but then we have to ensure that we add newly enabled tf-py-tests in all other manifests. Is that OK? Having a common generic list allows that list to be enabled for all with optional separate skiplists for different purpose.
> > 
> > 
> > Having a common run manifest is OK, but can we create one that's shared by all backend tests rather than using the nGraph CPU manifest as the common manifest?
> 
> We already have a common one: python_tests_list.txt. It has no mention of ng-cpu. (But folks also later created _int and _mac flavors for different build platforms/backends) Do you want a new common for all IE, say, python_tests_list_IE.txt? And then have skip_IECPU.txt, skip_IEGPU.txt etc?

I was thinking of a naming scheme like the following for all tests – C++ tests, pytests and TF unit tests:

```
tests_linux_ngraph_cpu.run
tests_linux_ngraph_interpreter.run
tests_linux_ie_cpu.run
tests_linux_ie_cpu.skip
tests_linux_ie_myriad.skip
tests_osx_ngraph_cpu.run
tests_common.run
tests_linux_common.run
tests_linux_ie_common.run
```

Of course, we don't need all of these but it's the convention we should follow. What do you think? Pretty soon we're going to want to run tests on IE that are not enabled on any of the nGraph backends which is why we definitely want a run manifest for it that's different than the nGraph CPU manifest.

bani-intelaipg:
Ok cool. I would like to make those consolidation naming changes together in a separate PR. For now, let's use the current PR to get started on TF-pytests.

tools/test_utils.py
```diff
@@ -113,6 +113,12 @@ def run_tensorflow_pytests_from_artifacts(backend, ngraph_tf_src_dir,
                                           tf_src_dir, xml_output):
     root_pwd = os.getcwd()
     ngraph_tf_src_dir = os.path.abspath(ngraph_tf_src_dir)
+    tf_src_dir = os.path.abspath(tf_src_dir)
+
+    if ('TFPYTEST_SKIPFILTERS' in os.environ):
+        # Use abs path. Do this early, before we change dir
+        filterfile = os.path.abspath(os.environ['TFPYTEST_SKIPFILTERS'])
+        os.environ['TFPYTEST_SKIPFILTERS'] = filterfile
```
adk9, line 121: Not needed.

adk9:
> Ok cool. I would like to make those consolidation naming changes together in a separate PR. For now, let's use the current PR to get started on TF-pytests. Let us meanwhile think about how do we make it easy for Devs to update all the relevant manifests when we enable/add/edit/deactivate a test.

Yeah, I wasn't suggesting we do it in this PR. Can you at least add a separate run manifest for IE CPU?

kanvi-nervana:
> > > > > The manifest for IE CPU shouldn't depend on the manifest for nGraph CPU, and it's confusing to have both a run and a skip manifest.
> > > > 
> > > > 
> > > > We could use an independent manifest for IECPU, but then we have to ensure that we add newly enabled tf-py-tests in all other manifests. Is that OK? Having a common generic list allows that list to be enabled for all with optional separate skiplists for different purpose.
> > > 
> > > 
> > > Having a common run manifest is OK, but can we create one that's shared by all backend tests rather than using the nGraph CPU manifest as the common manifest?
> > 
> > 
> > We already have a common one: python_tests_list.txt. It has no mention of ng-cpu. (But folks also later created _int and _mac flavors for different build platforms/backends) Do you want a new common for all IE, say, python_tests_list_IE.txt? And then have skip_IECPU.txt, skip_IEGPU.txt etc?
> 
> I was thinking of a naming scheme like the following for all tests – C++ tests, pytests and TF unit tests:
> 
> ```
> tests_linux_ngraph_cpu.run
> tests_linux_ngraph_interpreter.run
> tests_linux_ie_cpu.run
> tests_linux_ie_cpu.skip
> tests_linux_ie_myriad.skip
> tests_osx_ngraph_cpu.run
> tests_common.run
> tests_linux_common.run
> tests_linux_ie_common.run
> ```
> 
> Of course, we don't need all of these but it's the convention we should follow. What do you think? Pretty soon we're going to want to run tests on IE that are not enabled on any of the nGraph backends which is why we definitely want a run manifest for it that's different than the nGraph CPU manifest.

Why not have a single file like we currently do and just comment the tests we don't want to run instead have 2 separate files for each. The naming convention could be what's suggested.
**test_linux_ie_cpu.txt** and so forth

bani-intelaipg:
Re: manifest and/or skip names... There will be a very high overlap of tf-py-unitests across configurations (build-type, backend, plugin-type). So, with the proposed approaches, how do we ensure a desired test is not missing inadvertently in one of the manifests? For the first time, it's OK. As we enable more tests, this problem will worsen.

bani-intelaipg:
Changed to use a single manifest

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -87,41 +93,45 @@ def main():
...
+        #     for test in test_list[1]:
+        #         if test is not None:
+        #             invalid_list.append(test_list[1])
+        #             result_str = "\033[91m INVALID \033[0m " + test + \
+        #             '\033[91m' + '\033[0m'
+        #             print('TEST:', result_str)
+        #     test_list = list(set(test_list[0]))
+        #     for test_name in test_list:
+        #         if test_name not in all_test_list:
+        #             all_test_list.append(test_name)
```
adk9, line 121: Remove commented out code.

tools/test_utils.py
adk9: Is this still needed?

test/ci/buildkite/test_runner.py
```diff
@@ -97,7 +97,8 @@ def main():
     elif (arguments.test_tf_python):
         os.environ['NGRAPH_TF_LOG_0_DISABLED'] = '1'
         run_tensorflow_pytests_from_artifacts(
-            backend, './', arguments.artifacts_dir + '/tensorflow/python',
+            backend, os.path.abspath('./'),
+            os.path.abspath(arguments.artifacts_dir + '/tensorflow/python'),
```
adk9, line 101: Not needed.

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -87,41 +93,45 @@ def main():
...
 
 
 def get_test_list(tf_path, test_regex):
     accepted_formats = [
-        "math_ops_test.DivNoNanTest.testBasic", "math_ops_test.DivNoNanTest.*",
-        "math_ops_test.D*", "math_ops_test.*", "math_*_test", "math_*_*_test",
-        "math*_test"
+        "*test*", "math_ops_test.DivNoNanTest.testBasic",
+        "math_ops_test.DivNoNanTest.*", "math_ops_test.D*", "math_ops_test.*",
+        "math_*_test", "math_*_*_test", "math*_test"
```
adk9, line 134: This is just a standard regex; no need to be explicit about it.

kanvi-nervana:
So with these changes the TF py tests for other backends will run using the older files where we comment the tests we don't want to run and for IE we have an exclusion list?

test/python/tensorflow/tests_linux_ie_cpu.txt
```diff
@@ -0,0 +1,401 @@
...
+#    math_ops_test.* --> module.class
+#    math_*_test --> module
+#    math_*_*_test --> module
+#    math*_test --> module
+
+###################################################
+# First specify the included tests/patterns/regex
+# or, just include a filename relative to this file's path
+
+include python_tests_list.txt # --> include a filename relative to this file's path
```
kanvi-nervana, line 21: python_tests_list.txt has some tests commented, so those won't be run for IE anyway - correct?

bani-intelaipg:
> So with these changes the TF py tests for other backends will run using the older files where we comment the tests we don't want to run and for IE we have an exclusion list?
As of current implementation, for a given CI, we are free to choose any manifest we like, no dependency on any other manifest. The manifest has both inclusion and exclusion lists within. We can (if we want) include another manifest in the inclusion part. What I understood from Abhishek's requirements is: allow a generic tests_linux_common.txt manifest, which should contain all possible tests (that way we don't have to repeat those in other individual manifets).

tools/test_utils.py
bani-intelaipg: No

test/python/tensorflow/tf_unittest_runner.py
bani-intelaipg: Dots are treated literally, not like python regex

test/python/tensorflow/tests_common.txt
```diff
@@ -1,3 +1,31 @@
...
+# Read in one/more external manifest file(s)
+# Path specified is relative to this file's path
+
+# None
+
+###################################################
+[RUN]
+# Specify tests/patterns/regex that should be included
+
 #argmax_op_test.ArgMaxTest.testDefaultAxis
```
kanvi-nervana, line 29: Shouldn't the disabled tests be in the next section?

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -166,20 +176,24 @@ def regex_walk(dirname, regex_input):
...
+                if path not in sys.path:
+                    sys.path.append(os.path.abspath(path))
                 name = os.path.splitext(name)[0]
                 module_list.append(name)
     if not module_list:
-        print("Test name does not exist")
-        sys.exit(1)
+        print("Test pattern/name does not exist:", regex_input, "dirname",
+              dirname)
+        #sys.exit(1)
```
kanvi-nervana, line 196: commented code
Shouldn't the script exit f the test pattern is not found?

test/python/tensorflow/tests_common.txt
bani-intelaipg: Yes, they are blank. See line 655

test/python/tensorflow/tf_unittest_runner.py
bani-intelaipg: No, it's going to excuse and move on and parse more lines from the manifest. This will happen if user gives a pattern which doesn't match with any test.

kanvi-nervana: So if the pattern given by user does not match, the script continues with parsing the manifest - Shouldn't it exit because the user specified a certain pattern which wasn't found?

bani-intelaipg: I am leaning with the design that should continue parsing. The test result will display all the tests being executed, so user will know if the desired test is being run.

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -209,35 +223,136 @@ def list_tests(module_list, regex_input):
...
-            if line[0] != '#' and line != '\n'
-        ]
+def read_tests_from_manifest(manifestfile, tensorflow_path):
+    """
+    Reads a file that has include & exclude patterns,
+    Returns a list of leaf-level single testcase, no duplicates
+    """
+    list_of_tests = []
+    with open(manifestfile) as fh:
+        includefile_pattern = r'^\s*include\s+(\S+)\s*$'
```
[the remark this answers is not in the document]
bani-intelaipg: Sure. Thanks.

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -209,35 +223,136 @@ def list_tests(module_list, regex_input):
...
+        excluded_items = []  # this is temporary to each manifest
+        for line in fh.readlines():
+            line = line.split('#')[0].rstrip('\n').strip(' ')
+            if line == '':
+                continue
+            if re.search(r'\[IMPORT\]', line):
+                curr_section = 'import_section'
+                continue
+            if re.search(r'\[RUN\]', line):
+                curr_section = 'included_section'
```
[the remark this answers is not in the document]
bani-intelaipg: ok, thanks

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -209,35 +223,136 @@ def list_tests(module_list, regex_input):
...
+            if re.search(r'\[SKIP\]', line):
+                curr_section = 'excluded_section'
+                continue
+            if curr_section == 'import_section':
+                if not os.path.isabs(line):
+                    line = os.path.abspath(
+                        os.path.dirname(manifestfile) + '/' + line)
+                list_of_tests.extend(
+                    read_tests_from_manifest(line, tensorflow_path))
+                continue
```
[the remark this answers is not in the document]
bani-intelaipg: Once within an import, it's a nested recursion and comes back to caller/parent stack with a list of 'run' cases. Did I understand your comment correctly?

test/python/tensorflow/tf_unittest_runner.py
kanvi-nervana: As a user I would expect the script to only execute the tests that match the pattern I provide and if none match then not execute anything unless I do not provide a pattern at all, in which case it executes all the tests. It should continue parsing and executing if a test fails in-between in both cases.

bani-intelaipg: Let's discuss this over a call

kanvi-nervana (approved):

test/python/tensorflow/tf_unittest_runner.py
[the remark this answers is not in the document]
adk9: Ah, ok. I misread the code to think that `read_tests_from_manifest()` returns both run and skip tests. Here, you make an assumption that the skip tests are always a strict subset of run tests which may not necessarily be true.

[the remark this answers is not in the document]
adk9: Just to elaborate, I was thinking of a scenario like the following. Consider the manifest files:
**skip_mathops_tests.txt**
```
[RUN]
ArrayOps.*
[SKIP]
MathOps.*
```

**main_tests.txt**
```
[IMPORT]
skip_mathops_tests.txt
[RUN]
*
```

Would you expect the skip section to be imported?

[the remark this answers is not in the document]
bani-intelaipg: I see. No, the skip-items won't be satisfied with current implementation. I will put in an enhancement later.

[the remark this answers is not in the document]
adk9: Yeah, I think it's okay as long as we understand the behavior. That kind of a use-case won't be common for us.

The other thing you need to watch out for are circular imports. Some check to break the recursion may be needed.

PR #641, opened 2020-08-25
Minor refactor of tensor allocation in executor

adk9:
* Unify methods to allocate nGraph input and output tensors
* Avoid the use of internal TF DMA helper method

### DIFF
--- a/test/cpp_tests_list_IECPU.txt
+++ b/test/cpp_tests_list_IECPU.txt
@@ -17,8 +17,6 @@ TFExec.SingleGraphOn2Threads
 TFExec.axpy
 tf_exec.BatchMatMul
 EncapsulateOp.GetNgExecutable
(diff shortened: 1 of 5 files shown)


test/cpp_tests_list_IECPU.txt
```diff
@@ -17,8 +17,6 @@ TFExec.SingleGraphOn2Threads
 TFExec.axpy
 tf_exec.BatchMatMul
 EncapsulateOp.GetNgExecutable
-EncapsulateOp.AllocateNGInputTensors
-EncapsulateOp.AllocateNGOutputTensors
```
kanvi-nervana, line 21: Should the new test be added here?

adk9: No, because it seems to pass.

PR #650, opened 2020-09-03
enable passes via env vars

bani-intelaipg:
Default pass enable/disable settings are applied for all unspecified options via command line env-var.
For example, NGRAPH_PASS_ENABLES="TransposeSinking:0" will set TransposeSinking off, and use the deafult for ConstantFolding and others. Deafult values are specified in TranslateGraph()


adk9 (changes requested):

ngraph_bridge/ngraph_builder.cc
```diff
@@ -3347,7 +3347,8 @@ Status Builder::TranslateGraph(
   // Apply additional passes on the nGraph function here.
   //
   ngraph::pass::Manager passes;
-  // passes.register_pass<ngraph::pass::ConstantFolding>();
+  if (PassEnabled(NGraphPassType::ConstFold))
+    passes.register_pass<ngraph::pass::ConstantFolding>();
   passes.register_pass<pass::TransposeFolding>();
   passes.register_pass<pass::TransposeSinking>();
```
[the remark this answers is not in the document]
bani-intelaipg: Are these are the only 3 as of now?

ngraph_bridge/ngraph_utils.cc
```diff
@@ -32,6 +32,8 @@
 #include "ngraph_bridge/ngraph_utils.h"
```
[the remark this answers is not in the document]
adk9: `ngraph_utils.*` is a kitchen sink already. I think that it's preferable to have helper methods such as these in the proper ngraph_bridge namespace instead of putting them all in utils.

ngraph_bridge/ngraph_utils.cc
```diff
@@ -586,6 +589,30 @@ void ClearAttribute(Graph* g,
   }
 }
 
+static ngraph::pass::PassConfig* _p_pass_config = nullptr;
```
adk9, line 592: Singleton config is not required. You should check for enabled passes and attributes everytime `TranslateGraph()` is called.

ngraph_bridge/ngraph_utils.cc
```diff
@@ -586,6 +589,30 @@ void ClearAttribute(Graph* g,
...
+
+static void InitPassConfig() {
+  // use a singleton config object
+  if (_p_pass_config) return;
+  _p_pass_config = new ngraph::pass::PassConfig();
+  // Set the defaults, unless already specified via env var
+  if (ngraph::getenv_string("NGRAPH_PASS_ENABLES").empty()) {
```
adk9, line 599: This is error-prone. Use `PassConfig` to find out whether a pass was enabled or not.

adk9 (changes requested):

ngraph_bridge/ngraph_utils.cc
bani-intelaipg: We can/may call PassEnabled() many times as needed. The singleton is to just read the env-var once at the very beginning. If we had a system-init kind of func, we could init it there too.

adk9 (approved):

ngraph_bridge/ngraph_builder.cc
```diff
@@ -3257,11 +3258,24 @@ Status Builder::TranslateGraph(
...
-  passes.run_passes(ng_function);
+  {
+    ngraph::pass::Manager passes;
+    ngraph::pass::PassConfig pass_config;
+    // set/honor the defaults, unless specified via env var
+    auto set_default = [&pass_config](std::string pass, bool enable) {
+      auto enables_map = pass_config.get_enables();
+      if (enables_map.find(pass) == enables_map.end())
+        pass_config.set_pass_enable(pass, enable);
+    };
```
adk9, line 3269: Minor comment: This is fine, but if you change it to `get_enabled(std::string, bool default)` then there's only one change required to enable/disable a pass.

test/test_ngraph_exec.cpp
```diff
@@ -36,6 +41,14 @@ namespace tensorflow {
...
 namespace testing {
 
+string THISDIR =
+#if defined(WIN32) || defined(_WIN32) || \
+    defined(__WIN32) && !defined(__CYGWIN__)
+    std::regex_replace(__FILE__, std::regex("^(.*)\\[^\\]+$"), "$1\\");
+#else
+    std::regex_replace(__FILE__, std::regex("^(.*)/[^/]+$"), "$1/");
+#endif
+
```
adk9, line 51: Why is this required?

bani-intelaipg: I removed path dependency of running gtest from some other folder e.g. source root. Now, it will always find the associated pbtxt files in the same dir as the test script.

adk9: This creates a dependency on the source. If you want to not use a path relative to CWD, please use a path relative to the test executable.

adk9 (changes requested):

test/test_ngraph_exec.cpp
bani-intelaipg: I will revert to older approach of copying pbtxt files to build output.

adk9 (approved):

kanvi-nervana (approved):

PR #652, opened 2020-09-04
Improve test selection for TF unittest runner

bani-intelaipg:
- handle exact test name match
- The later RUN inavlidates previous SKIP, and vice-versa

### DIFF
--- a/test/python/tensorflow/tf_unittest_runner.py
+++ b/test/python/tensorflow/tf_unittest_runner.py
@@ -267,7 +267,13 @@ def read_tests_from_manifest(manifestfile, tensorflow_path):
     Reads a file that has include & exclude patterns,
     Returns a list of leaf-level single testcase, no duplicates
     """
-    list_of_tests = []
+
+    def invalidate_skips_after_runs(skips, runs):
+        for aTest in runs:
+            if aTest in skips:
+                skips.remove(aTest)
+
+    run_items = []
     skipped_items = []
     g_imported_files.append(manifestfile)
     with open(manifestfile) as fh:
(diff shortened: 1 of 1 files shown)


test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -243,10 +244,10 @@ def list_tests(module_list, regex_input):
                 alltests.append(aTestCase.id())
 
     for aTestCaseID in alltests:
```
adk9, line 246: Can you please follow the style here: https://google.github.io/styleguide/pyguide.html?

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -267,7 +268,13 @@ def read_tests_from_manifest(manifestfile, tensorflow_path):
     Reads a file that has include & exclude patterns,
     Returns a list of leaf-level single testcase, no duplicates
     """
-    list_of_tests = []
+
+    def invalidate_A_after_B(AList, BList):
+        for aTest in BList:
+            if aTest in AList:
+                AList.remove(aTest)
+
```
adk9, line 276: Use "sets" for tests and you can simply take a set difference for this. You'll also be able to eliminate any duplicated tests as a result of using sets.

adk9 (changes requested):

test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -243,10 +244,10 @@ def list_tests(module_list, regex_input):
                 alltests.append(aTestCase.id())
 
     for aTestCaseID in alltests:
-        if regex_input in aTestCaseID:  # substring match
+        if regex_input == aTestCaseID:  # exact match
             listtests.append(aTestCaseID)
-        elif regex_input.count('*') > 0:
-            regex_pattern = regex_input
+        else:
+            regex_pattern = '^' + regex_input + '$'
```
bani-intelaipg, line 250: This was buggy earlier.

adk9 (approved):
Looks good, thanks.

PR #656, opened 2020-09-10
Short-circuit functions with zero dim results for IE backend

adk9:
OV Inference Engine backends do not support zero dim tensors or results. In this PR, we mark such zero-dim result functions as trivial functions and short-circuit the whole cluster by turning into a NOP. This works because all clusters encapsulated by the bridge, at present, are strictly stateless.


bani-intelaipg (approved):
Some tests are failing, which need investigation. Code changes look good.

test/python/tensorflow/tests_linux_ie_cpu.txt
```diff
@@ -44,6 +44,10 @@ argmax_op_test.ArgMaxTest.testOutputEmpty
 argmax_op_test.ArgMaxTest.testEmpty
 array_ops_test.SnapshotOpTest.testInvertPermutation
 array_ops_test.StridedSliceGradTest.testInt64Indices
+array_ops_test.StridedSliceTest.testExpand
```
bani-intelaipg, line 47: Can we put some comments as to why these are skipped?

bani-intelaipg (changes requested):

test/python/tensorflow/tests_linux_ie_cpu.txt
kanvi-nervana: I think these are enabled now with the change

ngraph_bridge/ie_executable.cc
```diff
@@ -40,12 +40,18 @@ IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
...
+  //  3. zero function (* -> Zero)
   NGRAPH_VLOG(2) << "Checking for trivial functions in IE backend";
   bool trivial_fn = true;
   for (auto result : func->get_results()) {
     auto parent = result->input_value(0).get_node_shared_ptr();
+    auto& shape = result->get_shape();
     trivial_fn &= ngraph::is_type<opset::Parameter>(parent) ||
-                  ngraph::is_type<opset::Constant>(parent);
+                  ngraph::is_type<opset::Constant>(parent) ||
+                  count(shape.begin(), shape.end(), 0);
```
kanvi-nervana, line 54: Why do we have the zero dim check at 2 places?

test/python/tensorflow/tests_linux_ie_cpu.txt
adk9: That's correct.

ngraph_bridge/ie_executable.cc
adk9: This instance marks the function as a trivial function. When the trivial function is being executed, we handle all three cases separately and have to check what type of trivial function it is.

kanvi-nervana (approved):

PR #657, opened 2020-09-10
Upgrade LogSoftmax to opset3

kanvi-nervana:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -1898,28 +1898,14 @@ static Status TranslateLog1pOp(
...
-  auto ng_output = ConstructNgNode<opset::Subtract>(
-      op->name(), ng_inp_minus_max, ng_broadcast);
-  SaveNgOp(ng_op_map, op->name(), ng_output);
+  ng::Output<ng::Node> ng_input;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, ng_input));
+  auto input_shape = ng_input.get_shape();
+  auto rank = input_shape.size();
+  auto ng_log_softmax = ConstructNgNode<opset::Log>(
+      op->name(),
+      ConstructNgNode<opset::Softmax>(op->name(), ng_input, rank - 1));
```
adk9, line 1907: Softmax isn't numerically stable especially against under- and overflow; you need to normalize the input here (`logits - max(logits)`) to avoid stability errors.

See https://stackoverflow.com/a/52132033.

adk9 (changes requested):

ngraph_bridge/ngraph_builder.cc
bani-intelaipg: Yes, foir numerical stability we should calculate like: logsoftmax[i, j] = logits[i, j] - max(logits[i]) -	  SaveNgOp(ng_op_map, op->name(), ng_log_softmax);
  // log(sum(exp(logits[i] - max(logits[i]))))
where, we use  a multiplier C such that log(C) = - max(logits)

adk9 (approved):

PR #660, opened 2020-09-18
Fix Pad op translation

kanvi-nervana:
* Use helper `ConstructNgNode` for constructing nodes in the Pad translation
* Use `vector<int64>` to represent paddings as nGraph can internally convert to CoordinateDiff if necessary


bani-intelaipg:
Which tests were failing before this fix?

kanvi-nervana:
> Which tests were failing before this fix?

The tests weren't failing but the translation was using shared_ptr to node at a few places and CoordinateDiff which is not required anymore

adk9 (approved):

bani-intelaipg (approved):

PR #663, opened 2020-09-24
Revert "Eliminate use of pattern label for Concat in transpose-sinking"

adk9:
In some cases where a transpose is sunk through a Concat op, the shapes of the input arguments need to be transposed. In such cases, pattern labels are necessary to create dummy shapes for the new Concat op with transposed inputs.

* This reverts commit 8f95c9bd97ac08cbdd733c14f3e3c7aee2de9105.
* Also added a test case where dummy shapes are required in "concat sinking"
* Fixed a bug in pattern labels where the inputs to Concat come from a multi-output node such as a Split


kanvi-nervana (approved):

kanvi-nervana:
May be mention in the documentation, what the issue was which lead to the revert.

adk9:
> May be mention in the documentation, what the issue was which lead to the revert.

Documented the issue briefly in the PR description and the squashed commit.

PR #665, opened 2020-09-28
Increase minimum number of trivial nodes to 6

adk9:
Set the default minimum size of clusters to be at least 6 non-trivial
nodes (excluding Const and Identity).


test/python/test_ngraph_serialize_flag.py
```diff
@@ -46,13 +46,10 @@ def test_ng_serialize_to_json(self):
         out = tf.nn.l2_loss(tf.abs(x))
         values = np.random.rand(*xshape)
 
-        config = ngraph_bridge.update_config(tf.compat.v1.ConfigProto())
         ngraph_enable_serialize = os.environ.pop('NGRAPH_ENABLE_SERIALIZE',
                                                  None)
         os.environ['NGRAPH_ENABLE_SERIALIZE'] = '1'
-        ngraph_bridge.enable()
```
kanvi-nervana, line 53: How are these changes related to increasing minimum #of trivial nodes?

ngraph_bridge/ngraph_deassign_clusters.cc
```diff
@@ -51,7 +51,7 @@ namespace ngraph_bridge {
 // NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS=1.
 //
 
-static const int MIN_NONTRIVIAL_NODES = 2;
+static const int MIN_NONTRIVIAL_NODES = 6;
```
kanvi-nervana, line 54: Curious, how did you decide with the number? Is it related to the model you were running?

test/python/test_ngraph_serialize_flag.py
adk9: This test wasn't written to run with cluster de-assignment disabled, and hence depends on the parameters used for cluster deassignment.

ngraph_bridge/ngraph_deassign_clusters.cc
adk9: Not really. Generally speaking, we'd need some sort of symbolic analysis/execution coupled with a cost model to come up with a good heuristic for cluster triviality. Nevertheless, 2 seems like a really low number for a trivial cluster. Given our current op coverage and the average number of nodes/cluster for the standard set of models we test against, 6 seems like a safe minimum?

bani-intelaipg (approved):

PR #666, opened 2020-09-28
Remove ScatterNd op and tests

kanvi-nervana:
Remove ScatterNd as it does not have an opset3 equivalent


adk9:
Can't `opset::ScatterUpdate` be used for scatter_nd?

adk9 (approved):

kanvi-nervana:
> Can't `opset::ScatterUpdate` be used for scatter_nd?

They are a bit different in semantics and as discussed we'll revisit it with opset5 upgrade.

PR #667, opened 2020-09-29
Fix FusedBatchNorm translation

kanvi-nervana:
Fix argument order for `opset::BatchNormInference` in _FusedConv2D translation


adk9 (approved):
I think both constructors work but it probably makes sense to use this order since the other one is deprecated.

PR #668, opened 2020-09-29
Handle multi-output results nodes in IE executor

adk9:
Handle cases where the node feeding into a result node has multiple
outputs possibly leading to multiple results as well. In cases where the
parent of the result node has multiple outputs, we use the correct
output index to set the blob corresponding to the result's output.


test/python/tensorflow/tests_linux_ie_cpu.txt
```diff
@@ -28,38 +28,30 @@ tests_common.txt
 
 ###################################################
 [SKIP]
-# Specify tests/patterns/regex that should be excluded/skipped
+
```
kanvi-nervana, line 31: Are there tests enabled or just moved as I see additions as well?

adk9: There aren't supposed to be any additions. The tests are just moved as this change enabled many of them.

kanvi-nervana (approved):

PR #669, opened 2020-09-29
Extend type constraint to exclude double-precision data types

adk9:
* Do not mark ops with double-precision data types for clustering
* Change a few unit tests to use single-precision data type for testing


adk9:
This PR depends on #668.

kanvi-nervana (approved):

PR #670, opened 2020-09-29
Upgrade OV to 2021.1 pre-release

adk9:
This PR upgrades the version of OpenVINO to 2021.1 pre-release. Changes include:
* Wrap node check calls to support both nGraph master and OV nGraph
* Replace calls to deprecated `get_argument()`
* Comment out reduction tests because of regression introduced in reduction operations possibly for rank 1 tensors
* Use signed integer for reshape pattern
* Minor debug/error formatting fixes
  * Print more node details in log-placement debug prints
  * Add missing space in compilation failure error message
* Fix parameter replacement check in IE executor
  * Check for all constants to replace to parameters until one is found
  * Dump optimized nGraph function in IE executor


adk9:
@googlebot I consent.

kanvi-nervana:
@googlebot I consent.

bani-intelaipg:
@googlebot I consent

PR #671, opened 2020-09-29
Remove nGraph function serialization support

adk9:
Since OV 2021.1 nGraph deprecated and removed support for nGraph
JSON serializer (https://github.com/openvinotoolkit/openvino/pull/1638),
remove usage of `ngraph::serialize()` and use visualize-tree pass
(`ngraph::plot_graph()`) instead to dump graphs.


ngraph_bridge/ngraph_encapsulate_impl.cc
```diff
@@ -135,8 +134,6 @@ Status NGraphEncapsulateImpl::GetNgExecutable(
     try {
       ng_exec = backend->compile(ng_function);
     } catch (const std::exception& ex) {
-      string fn_name = ng_function->get_friendly_name();
-      NgraphSerialize("tf_function_" + fn_name + ".json", ng_function);
```
kanvi-nervana, line 139: plot_graph in case of failure?

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -292,17 +292,11 @@ void NGraphEncapsulateOp::Compute(OpKernelContext* ctx) {
       try {
         ng_exec->call(ng_outputs, ng_inputs);
       } catch (const std::exception& exp) {
-        NgraphSerialize(
```
kanvi-nervana, line 295: same as above

ngraph_bridge/ngraph_encapsulate_op.cc
```diff
@@ -292,17 +292,11 @@ void NGraphEncapsulateOp::Compute(OpKernelContext* ctx) {
...
-            "tf_function_error_" + ctx->op_kernel().name() + ".json",
-            ng_function);
         string status_string =
             "Caught exception while executing nGraph computation: " +
             string(exp.what());
         OP_REQUIRES(ctx, false, errors::Internal(status_string));
       } catch (...) {
-        NgraphSerialize(
```
kanvi-nervana, line 303: same as above

kanvi-nervana (changes requested):

bani-intelaipg:
@googlebot I consent

ngraph_bridge/ngraph_encapsulate_impl.cc
kanvi-nervana: May be add fn_name to the error message, that way we'll know which file to look at.

adk9: > plot_graph in case of failure?

I think it's of limited use to dump the graph on error with there being limited debug info in the graphs at the moment. For now, dumping graphs is explicit through `NGRAPH_TF_DUMP_GRAPHS`.

> May be add fn_name to the error message, that way we'll know which file to look at.

Good suggestion. I've added function name to the error message.

PR #674, opened 2020-10-06
Use manifest for all CPP and PY tests

kanvi-nervana:
Use manifest for all tests - enable the individually disabled tests.
Add Select op
Add FloorMod op
Disable tests using FloorMod for CPU and Interpreter as it is not supported on either
Remove cnms test


test/tests_linux_ie_cpu.txt
```diff
@@ -17,14 +17,18 @@ tests_common.txt
...
 # Specify tests/patterns/regex that should be excluded/skipped
 
 # [PARAMETER_MISMATCH] Failed to set Blob with precision not corresponding to user output precision
 NGraphExecTest.Axpy8bit
+MathOps.AnyKeepDims
+MathOps.AnyNegativeAxis
+MathOps.AnyPositiveAxis
+MathOps.AllKeepDims
+MathOps.AllNegativeAxis
+MathOps.AllPositiveAxis
```
adk9, line 31: Why do these get added? Are the failures indirectly related to FloorMod?

kanvi-nervana: No, these tests were individually disabled and not being run on IE or CPU or Interpreter. I moved them to the manifest after enabling and checking the status.

bani-intelaipg (approved):

adk9 (approved):

PR #679, opened 2020-10-09
Remove option to build with standalone nGraph

adk9:
This PR removes options to build and test against standalone nGraph. The default build option now builds with nGraph distributed by OpenVINO.


python/ngraph_bridge/__init__.in.py
```diff
@@ -223,5 +218,4 @@ def get_disabled_ops():
     "nGraph version used for this build: " + str(ngraph_bridge_lib.ngraph_lib_version()) + "\n" + \
```
kanvi-nervana, line 218: would this change?

adk9: Change to what? This is just printing out the version of nGraph in use.

kanvi-nervana: nGraph version won't change anymore, right?
So we don't need to display it.

kanvi-nervana (commented):
Remove these files as well
https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/tensorflow/tests_linux_ngraph_cpu.txt
https://github.com/tensorflow/ngraph-bridge/blob/master/test/python/tensorflow/tests_linux_ngraph_interpreter.txt

python/ngraph_bridge/__init__.in.py
adk9: It's not under our control, but I agree that it isn't particularly useful to show.

kanvi-nervana: when building with OV I have noticed that the version is 0.0.0+<some-commit-id>
eg: nGraph version used for this build: b'0.0.0+a5f5384' which is why I assumed it wouldn't change

adk9: It identifies the git commit which changes with every version.

But if you notice carefully, our integration (build scripts, source, etc.) are still nGraph-specific; this PR only changes where the nGraph we build against comes from. Should we revisit the suggested change above as part of the OV rebranding?

kanvi-nervana: Sure, we can do that.

bani-intelaipg (approved):

build_ngtf.py
```diff
@@ -334,104 +308,54 @@ def main():
...
+
+        cmake_build(build_dir, openvino_src_dir, openvino_cmake_flags,
+                    verbosity)
 
     # Next build CMAKE options for the bridge
     ngraph_tf_cmake_flags = [
         "-DNGRAPH_TF_INSTALL_PREFIX=" + artifacts_location,
-        "-DUSE_PRE_BUILT_NGRAPH=ON",
         "-DNGRAPH_TARGET_ARCH=" + target_arch,
         "-DNGRAPH_TUNE_ARCH=" + target_arch,
```
kanvi-nervana, line 346: Are these flags required for the bridge?

adk9: No they're not. Good catch, I'll remove them from the build script too.

kanvi-nervana (approved):

build_ngtf.py
adk9: Actually, I added it to the bridge build just in case, but most importantly also pass it to OV's build (d0d573494)

PR #683, opened 2020-10-13
Add prebuilt intel tf build option

kanvi-nervana:
depends on #684


adk9:
Can we just have `--use-prebuilt-tf` take a package/version defaulting to `tensorflow==2.2.0`?

kanvi-nervana:
> Can we just have `--use-prebuilt-tf` take a package/version defaulting to `tensorflow==2.2.0`?

We already have the option to take a version, in addition to that an option to differentiate between native or intel tf? Isn't having a separate build flag more straight forward?

adk9:
> > Can we just have `--use-prebuilt-tf` take a package/version defaulting to `tensorflow==2.2.0`?
> 
> We already have the option to take a version, in addition to that an option to differentiate between native and intel tf? Isn't having a separate build flag more straight forward?

The `--use-prebuilt-tf` flag simply instructs the script to build against a prebuilt TF wheel that comes from PyPI. It doesn't strictly matter where it comes from or what version it is. (It may matter now if you need to fetch the source to run TF unit tests but it's possible to extract the version of the prebuilt package).

This is okay but I was thinking of something more general that allows us to choose packages or optionally specify version numbers: 
* `--use-prebuilt-tf=tensorflow`
* `--use-prebuilt-tf=intel-tensorflow`
* `--use-prebuilt-tf=intel-tensorflow==2.3.0`

so on and so forth.

bani-intelaipg:
I have similar comment like adk9

build_ngtf.py
```diff
@@ -29,7 +29,8 @@ def version_check(use_prebuilt_tensorflow, use_tensorflow_from_location,
             gcc_ver = get_gcc_version()
             if gcc_ver < '5.3.0':
                 raise Exception(
-                    "Need GCC 5.3.0 or newer to build using prebuilt TensorFlow\n"
+                    "Need GCC 5.3.0 or newer to build using prebuilt TensorFlow"
+                    " or Intel-Tensorflow\n"
```
adk9, line 33:
```suggestion
                    " or Intel TensorFlow\n"
```

build_ngtf.py
```diff
@@ -49,11 +50,10 @@ def version_check(use_prebuilt_tensorflow, use_tensorflow_from_location,
 
 def main():
     '''
-    Builds TensorFlow, ngraph, and ngraph-tf for python 3
+    Builds TensorFlow, OpenVino, and ngraph-tf for python 3
```
adk9, line 53:
```suggestion
    Builds TensorFlow, OpenVINO, and ngraph-tf for python 3
```

build_ngtf.py
```diff
@@ -216,6 +218,9 @@ def main():
     if arguments.use_prebuilt_tensorflow != '':
         tf_version = arguments.use_prebuilt_tensorflow
 
+    if arguments.use_intel_tensorflow != '':
+        use_intel_tf = True
```
adk9, line 222: No need for this variable, use `arguments.use_intel_tensorflow` directly.

build_ngtf.py
```diff
@@ -144,8 +146,8 @@ def main():
 
     assert not (
         arguments.use_tensorflow_from_location != '' and
-        arguments.use_prebuilt_tensorflow != ''
-    ), "\"use_tensorflow_from_location\" and \"use_prebuilt_tensorflow\" "
+        (arguments.use_prebuilt_tensorflow != '')
```
adk9, line 149: Is this change needed?

build_ngtf.py
```diff
@@ -256,10 +261,15 @@ def main():
         os.chdir(cwd)
     else:
         if arguments.use_prebuilt_tensorflow != '':
-            print("Using existing TensorFlow version", tf_version)
-            # Install TensorFlow
-            command_executor(
-                ["pip", "install", "-U", "tensorflow==" + tf_version])
+            print("Using TensorFlow version", tf_version)
+            if use_intel_tf:
+                print("Install Intel-Tensorflow")
```
adk9, line 266:
```suggestion
                print("Install Intel TensorFlow")
```

build_ngtf.py
```diff
@@ -256,10 +261,15 @@ def main():
...
+                command_executor(
+                    ["pip", "install", "-U", "intel-tensorflow==" + tf_version])
+            else:
+                print("Install native Tensorflow")
```
adk9, line 270:
```suggestion
                print("Install native TensorFlow")
```

build_ngtf.py
```diff
@@ -272,10 +282,6 @@ def main():
                           "https://github.com/tensorflow/tensorflow.git",
                           tf_version)
             os.chdir(pwd_now)
-            if not arguments.disable_cpp_api:
-                # Now build the libtensorflow_cc.so - the C++ library
-                build_tensorflow_cc(tf_version, tf_src_dir, artifacts_location,
-                                    target_arch, verbosity)
         else:
```
adk9, line 285: Is this still needed to build CC API? How would someone build the CC API with prebuilt TF now?

adk9 (changes requested):

build_ngtf.py
kanvi-nervana: This variable is needed for  build_tensorflow and build_tensorflow_cc functions to build with config = mkl option.

build_ngtf.py
kanvi-nervana: As discussed, I removed the dependence for prebuilt option on this flag to reduce the number of build flags we need to pass.

build_ngtf.py
kanvi-nervana: I think it's accidental change and not needed.

adk9 (approved):

PR #685, opened 2020-10-13
Upgrade to opset4

kanvi-nervana:
Upgrade to opset4
Update Softplus translation to use opset4 nGraph op SoftPlus
Add translations for unary ops Acosh, Asinh, Atanh


adk9 (approved):
Nice, thanks!

bani-intelaipg (changes requested):
Should we enable/run some related tf-py-tests possibly?

kanvi-nervana:
> Should we enable/run some related tf-py-tests possibly?

Yes, we can check if there are any old/new tests that can be enabled.

PR #696, opened 2020-11-05
Add use_openvino_from_location option

kanvi-nervana:
Rename use_prebuilt_openvino --> use_openvino_from_location to be consistent with TF build options at the bridge
Move build steps to a separate file build_ov.py


build_ngtf.py
```diff
@@ -351,9 +351,13 @@ def main():
             "-DENABLE_SPEECH_DEMO=FALSE",
             "-DCMAKE_INSTALL_RPATH=\"$ORIGIN\"",
             "-DCMAKE_INSTALL_PREFIX=" + os.path.join(artifacts_location,
-                                                     "openvino")
+                                                     "openvino"),
+            "-DENABLE_PYTHON=ON",
+            "-DPYTHON_EXECUTABLE=" + os.path.join(build_dir_abs,
+                                                  "venv-tf-py3/bin/python3"),
```
bani-intelaipg, line 357: venv dir is defined in line 191, we could use that

build_ngtf.py
```diff
@@ -351,9 +351,13 @@ def main():
             "-DENABLE_SPEECH_DEMO=FALSE",
             "-DCMAKE_INSTALL_RPATH=\"$ORIGIN\"",
             "-DCMAKE_INSTALL_PREFIX=" + os.path.join(artifacts_location,
-                                                     "openvino")
+                                                     "openvino"),
+            "-DENABLE_PYTHON=ON",
+            "-DPYTHON_EXECUTABLE=" + os.path.join(build_dir_abs,
```
bani-intelaipg, line 356: Instead of forcing Python3, can we get/read it from current venv?

bani-intelaipg (changes requested):

build_ngtf.py
kanvi-nervana: Our scripts assume python3 https://github.com/tensorflow/ngraph-bridge/blob/master/tools/build_utils.py#L125 so we can probably stick to it.

bani-intelaipg (approved):

build_ngtf.py
```diff
@@ -328,39 +329,31 @@ def main():
...
-        openvino_src_dir = "./openvino"
+    if arguments.use_openvino_from_location != "":
+        print("Using OpenVINO from " + arguments.use_openvino_from_location)
+    else:
+        print("Building OpenVINO from source")
+        print(
+            "NOTE: OpenVINO python module is not built when building from source."
+        )
+        python_module_enabled = False
+
```
adk9, line 340: Remove building of Python module. We do not use the OpenVINO Python API.

test/ci/buildkite/setup.yml
```diff
@@ -6,7 +6,7 @@ steps:
       rm -rf /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID
       virtualenv -p python3 /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/venv 
       source /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/venv/bin/activate 
-      pip install -U pip yapf==0.26.0 pytest psutil keras==2.3.1
+      pip install -U pip yapf==0.26.0 pytest psutil keras==2.3.1 cython
```
adk9, line 9: Not needed.

build_ngtf.py
```diff
@@ -328,39 +329,31 @@ def main():
...
-            "-DENABLE_SPEECH_DEMO=FALSE",
-            "-DCMAKE_INSTALL_RPATH=\"$ORIGIN\"",
-            "-DCMAKE_INSTALL_PREFIX=" + os.path.join(artifacts_location,
-                                                     "openvino")
-        ]
-
+        debug_enabled = False
         if arguments.debug_build:
-            openvino_cmake_flags.extend(["-DCMAKE_BUILD_TYPE=Debug"])
+            debug_enabled = True
```
adk9, line 352: `arguments.debug_build` can be used directly below.

adk9 (changes requested):

adk9 (approved):

PR #697, opened 2020-11-06
Refactor logging

adk9:
* Remove separate "ngraph_logger" library
* Remove TF graph serialization to dot
* Use TF logging for verbose logging


adk9:
Closing for now in favor of #786 which removes the separate `ngraph_logger` library while keeping the `NGRAPH_VLOG` logging macros.

PR #698, opened 2020-11-10
Remove old obselete model-level tests

adk9:
Move new model-zoo scripts to tests/models/tf_ov_model_zoo


kanvi-nervana (approved):
Does it get rid of the nightly pipeline as well?

bani-intelaipg (approved):

adk9:
> Does it get rid of the nightly pipeline as well?

No, but I've deactivated the pipeline now.

PR #699, opened 2020-11-11
Translation for LRN

kanvi-nervana:
This adds a translation for TensorFlow's local_response_normalization op.

https://www.tensorflow.org/api_docs/python/tf/nn/local_response_normalization
https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/l-r-n


adk9 (approved):

test/test_nn_ops.cpp
```diff
@@ -554,6 +554,49 @@ TEST(NNOps, LogSoftmax) {
...
 }
 
+// Test Op :"LRN"
+TEST(NNOps, LRN) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 2;
+  int dim3 = 2;
+  int dim4 = 2;
+
```
adk9, line 564: Style comment: This could be inlined if it isn't being used anywhere else.

bani-intelaipg (commented):
LGTM

PR #701, opened 2020-11-12
Handle multi-output nodes in TS

kanvi-nervana:
Delete transposes instead of replacing
Add a new TS test for multi-output nodes


ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -105,11 +105,15 @@ static void write_transposemap(TransposeMap& reorders,
 
 static shared_ptr<opset::Transpose> read_transposemap(
     TransposeMap& reorders, ngraph::Output<ngraph::Node> target) {
+  // if(reorders.find(target.get_node_shared_ptr()) != reorders.end()) {
```
adk9, line 108: Is this needed? Do we ever try to read transposes that are not in the transposemap?

ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -174,8 +178,13 @@ static void insert_transpose(shared_ptr<ngraph::Node> target,
 static void delete_transpose(shared_ptr<ngraph::Node> transpose) {
   NGRAPH_VLOG(4) << "Removing transpose " << transpose->get_name();
   if (!transpose->get_users().empty()) {
-    ngraph::replace_node(transpose,
-                         transpose->input_value(0).get_node_shared_ptr());
+    ngraph::Output<ngraph::Node> output = transpose->output(0);
+    NGRAPH_VLOG(4) << "output " << output.get_node_shared_ptr()->get_name();
```
adk9, line 182: Make these newly added vlogs here level 5 if you absolutely must have them.

ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -461,14 +469,18 @@ bool TransposeSinking::run_on_function(shared_ptr<ngraph::Function> f) {
   TransposeMap reorders, reuse_map;
   set<shared_ptr<ngraph::Node>> transposes_to_delete;
   unordered_map<std::string, ngraph::Shape> orig_result_out_shape;
+  if (std::getenv("NGRAPH_TF_DUMP_GRAPHS") != nullptr) {
+    NGRAPH_VLOG(0) << "Dumping ng_function before TS to Before_TS_"
+                   << f->get_name() << ".dot";
+    ngraph::plot_graph(f, "Before_TS_" + f->get_name() + ".dot");
+  }
```
adk9, line 476: Should we dump these graphs in between the passes in the builder? It would be nice to have the function dumped after each pass à-la TF.

ngraph_bridge/pass/transpose_sinking.cc
kanvi-nervana: No, this was done while debugging the model and there was an issue with reading the map for which there wasn't an entry. I'll remove, don't think we'll hit this issue.

ngraph_bridge/pass/transpose_sinking.cc
kanvi-nervana: We can do this in the builder.

kanvi-nervana: On second thought, I think it's better this way. In the builder we can do the "before_running_all_passes" but in between might not be possible since we run them all together.

adk9 (approved):

PR #703, opened 2020-11-13
Cavusmustafa/tf tensor without blob

cavusmustafa:
This is the first PR for the modular backend.
IE Blob is removed from IE_Tensor. Blobs are created during execution (in call() function of Executable). This modification is required to enable batch parallelism for HDDL later.
A new file, "ie_utils.hpp" is created for generic ie related operations. It only includes a function to create ie blobs for now. It will include additional functions in next PRs.

### DIFF
--- a/ngraph_bridge/executable.cc
+++ b/ngraph_bridge/executable.cc
@@ -142,7 +142,7 @@ bool Executable::call(const vector<shared_ptr<runtime::Tensor>>& inputs,
   auto func = m_network.getFunction();
   auto parameters = func->get_parameters();
   for (int i = 0; i < inputs.size(); i++) {
-    shared_ptr<IETensor> tv = static_pointer_cast<IETensor>(inputs[i]); 
+    shared_ptr<IETensor> tv = static_pointer_cast<IETensor>(inputs[i]);
     InferenceEngine::SizeVector input_shape = tv->get_dims();
     InferenceEngine::Precision input_precision = tv->get_precision();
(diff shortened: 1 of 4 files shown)


adk9:
@cavusmustafa Sorry for the delay, but I'm not sure I understand the intent behind this change. Would looking at #702 help?

Is IETensor an nGraph tensor or an IE Blob? Would we need an IETensor at all if the backend were to directly operate on IE blobs?

cavusmustafa:
This is a part of changes from #702 . IE Blobs are created during the inference execution instead of IETensor's constructor. IETensor does not own a IE Blob variable anymore. Instead, it owns the information (like description, shape, etc.) needed to create a blob.

IETensor is still the parameter to the call() function. During the execution inside the call() function, IE Blobs are created using the information from IETensor.

The reason of this change is to be able to handle batch parallelism for HDDL. #702  has the full implementation which might explain why this is needed. Multiple number of IE Blobs should be created from a single TF Input Tensor based on the number of requests. Batch paralellism can be enabled/disabled or the number of requests might be changed after IETensors are created. Moving the blob creation into inference execution stage provides flexibility in this case.

Alternatively, we can have a IE Blob vector inside IETEnsor but in this case we have to make sure if batch paralellism is enabled and we now the total number of requests when IETensor is created.

IETensor is still an nGraph tensor (#702  doesn't need it, but some other parts of NGraph-Bridge still consider it as an nGraph tensor).

Also, any change made inside the call() function in this PR is temporary. The final implementation should look like #702 .

PR #706, opened 2020-11-19
Simplify layout conversion and padding helpers

adk9:
* Remove unnecessary conversion helpers
* Rename conversion helpers to be explict about layouts, e.g. `NCHWtoNHWC`
* Remove duplicate padding computations and use nGraph API for padding
* Remove duplicated Maxpool3D translation and use same Maxpool translation for 2D/3D

### DIFF
--- a/ngraph_bridge/ngraph_conversions.cc
+++ b/ngraph_bridge/ngraph_conversions.cc
@@ -19,49 +19,31 @@
 
 namespace tensorflow {
 namespace ngraph_bridge {
-namespace detail {
 
-void NhwcToNGraph(ngraph::Output<ngraph::Node>& node) {
-  Transpose<0, 3, 1, 2>(node);
-}
-
-void NdhwcToNGraph(ngraph::Output<ngraph::Node>& node) {
-  Transpose3D<0, 4, 1, 2, 3>(node);
-}
-}  // namespace detail
-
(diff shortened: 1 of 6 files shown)


ngraph_bridge/ngraph_conversions.cc
```diff
@@ -19,49 +19,31 @@
...
-  Builder::SetTracingInfo(op_name, node);
-}
-
-void BatchToTensorflow3D(const string& op_name, bool is_ndhwc,
-                         ngraph::Output<ngraph::Node>& node) {
-  if (!is_ndhwc) {
-    return;
+void NCHWtoNHWC(const string& op_name, bool is_nhwc,
+                ngraph::Output<ngraph::Node>& node) {
+  if (is_nhwc) {
```
bani-intelaipg, line 38: Should this be is_nchw?

kanvi-nervana: Or (!is_nhwc)

adk9: Actually, this is somewhat unintuitive but I've maintained the original behavior here, i.e. if the original inputs were in NHWC then we add transposes in the beginning (`NHWCoNCHW`) and at the end (`NCHWtoNHWC`) by looking at the `is_nhwc` flag.

PR #707, opened 2020-11-20
Remove unused "data cache"

adk9:
* Remove unused ngraph_data_cache.h and tests
* Move thread_safe_queue.h to examples/cpp where it's used
* Move cpp inference utilities one level up to examples/cpp


examples/cpp/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # ******************************************************************************
-# Copyright 2019-2020 Intel Corporation
+# Copyright 2018-2020 Intel Corporation
```
kanvi-nervana, line 2: Do all files need this, then?

test/test_thread_safe_queue.cc
```diff
@@ -1,131 +0,0 @@
-/*******************************************************************************
- * Copyright 2019-2020 Intel Corporation
- *
```
kanvi-nervana, line 3: thread_safe_queue has been moved to a new location but not removed, shouldn't the tests still apply?

examples/cpp/tf_label_image_utils.cc
```diff
@@ -300,6 +300,8 @@ Status GetTopLabels(const std::vector<Tensor>& outputs, int how_many_labels,
 //-----------------------------------------------------------------------------
 Status PrintTopLabels(const std::vector<Tensor>& outputs,
                       const string& labels_file_name) {
+  // Disable nGraph so that we run these using TensorFlow on CPU
+  tensorflow::ngraph_bridge::config::ngraph_disable();
```
kanvi-nervana, line 304: Why disable the bridge?

examples/cpp/CMakeLists.txt
adk9: No.

examples/cpp/tf_label_image_utils.cc
adk9: The bridge is disabled here to run validation using vanilla TF (CPU).

test/test_thread_safe_queue.cc
adk9: Yes, I moved it to `examples/cpp` since `infer_multiple_networks` was the only consumer of it. I'm not sure if we need either that utility or another implementation of a thread-safe queue, but I've added it back anyway.

PR #708, opened 2020-11-20
Rename encapsulate op to _nGraphEncapsulate

adk9:
* Prefix encapsulation op introduced by our JIT passes with an underscore to _nGraphEncapsulate to
  indicate that it's not a public TF op
* Remove unused function in ngraph_utils.{c,h}


ngraph_bridge/ops/ngraph_ops.cc
```diff
@@ -1,34 +0,0 @@
-/*******************************************************************************
```
kanvi-nervana, line 1: This is a TF convention to keep Op registrations in a separate file from the kernel. Intel-TF also follows the same and I think we should keep this file even though there is only 1 op that we register.

ngraph_bridge/ngraph_utils.cc
```diff
@@ -469,7 +463,7 @@ bool IsProcessedByNgraphPass(Graph* g) {
   // Current method may fail when graph has no encapsulates after first pass
   // Also variable/optimizer change introduces other types of ng nodes
   for (Node* node : g->nodes()) {
-    if (node->type_string() == "NGraphEncapsulate") return true;
+    if (node->type_string() == "_nGraphEncapsulate") return true;
```
kanvi-nervana, line 466: Comment at line 464 can be removed, we don't have variables/optimizers anymore

bani-intelaipg (approved):

ngraph_bridge/ngraph_utils.cc
adk9: Removed.

ngraph_bridge/ops/ngraph_ops.cc
adk9: Ok, I'll do the refactoring to use canonical TF convention in a follow-up PR.

adk9: See #710

PR #715, opened 2020-12-02
Upgrade OV to pre-release 2021.2

adk9:
* Upgrade OV to pre-release 2021.2
* Add env vars `TF_OV_CONSTANT_FOLDING` and `TF_OV_TRANSPOSE_SINKING` to control behavior of passes
* Fix incorrect int FloorDiv translation by restricting translation to only real types for now
* Allow all types for FloorMod


ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -452,9 +452,8 @@ const TypeConstraintMap& GetTypeConstraintMap() {
     type_constraint_map["Exp"]["T"] = NGraphNumericDTypes();
     type_constraint_map["ExpandDims"]["T"] = NGraphDTypes();
     type_constraint_map["Floor"]["T"] = NGraphNumericDTypes();
-    type_constraint_map["FloorDiv"]["T"] = NGraphNumericDTypes();
-    type_constraint_map["FloorMod"]["T"] = {
-        DT_INT32};  // IE only supports i32 for input
+    type_constraint_map["FloorDiv"]["T"] = NGraphRealDTypes();
```
bani-intelaipg, line 455: Are we restricting type due to OV-IE limitation?

bani-intelaipg (changes requested):

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2893,20 +2878,13 @@ Status Builder::TranslateGraph(
...
-    // set/honor the defaults, unless specified via env var
-    auto set_default = [&pass_config](std::string pass, bool enable) {
-      auto enables_map = pass_config.get_enables();
-      if (enables_map.find(pass) == enables_map.end())
-        pass_config.set_pass_enable(pass, enable);
-    };
-    set_default("ConstantFolding", false);
-    set_default("TransposeSinking", true);
-    if (pass_config.get_pass_enable("ConstantFolding"))
+    passes.register_pass<pass::TransposeSinking>();
```
kanvi-nervana, line 2881: Should the pass be registered even if the user does not want to run it?

test/python/tensorflow/tests_linux_cpu.txt
```diff
@@ -54,12 +54,12 @@ scan_ops_test.CumsumTest.test1D
 scan_ops_test.CumsumTest.test2D
 scan_ops_test.CumsumTest.test3D
 
-
 #Unable to add a parameter to a function with no parameters!
 argmax_op_test.ArgMaxTest.testInt64
 array_ops_test.StridedSliceGradTest.testInt64Indices
 cast_op_test.CastOpTest.testIntToFloatBoundary
 math_ops_test.AddNTest.testInt
+math_ops_test.ModTest.testFixed
```
kanvi-nervana, line 62: Is this a new test?

ngraph_bridge/ngraph_builder.cc
adk9: Done.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -2878,13 +2878,13 @@ Status Builder::TranslateGraph(
   //
   {
     ngraph::pass::Manager passes;
-    passes.register_pass<pass::TransposeSinking>();
     if (GetEnv("TF_OV_CONSTANT_FOLDING") == "1") {
       passes.register_pass<ngraph::pass::ConstantFolding>();
     }
     if (GetEnv("TF_OV_TRANSPOSE_SINKING") != "0") {
-      passes.run_passes(ng_function);
+      passes.register_pass<pass::TransposeSinking>();
```
kanvi-nervana, line 2885: Will TS be enabled by default, even if the user does not pass any env var?

ngraph_bridge/ngraph_mark_for_clustering.cc
adk9: In a way, yes. If IE supported a `FloorDiv` op, we could support all numeric types here. But since it doesn't, we need to handle additional cases for integer types - especially the ones with negative integer types. This will require static input constraints on the FloorDiv op which could be added in a follow-up PR.

test/python/tensorflow/tests_linux_cpu.txt
adk9: I re-enabled this test since we're restricting FloorMod only to float types now.

ngraph_bridge/ngraph_builder.cc
adk9: I think so. It will be disabled only when the user sets `TF_OV_TRANSPOSE_SINKING=0`.

PR #721, opened 2020-12-03
Update pack/unpack translations

adk9:
Update pack/unpack translations to be more optimized


ngraph_bridge/ngraph_builder.cc
```diff
@@ -2531,58 +2496,33 @@ static Status TranslateUnpackOp(const Node* op,
...
-        op->name(), ng_input, lower_bound, upper_bound, std::vector<int64_t>{},
-        std::vector<int64_t>{});
-    auto ng_shape = ConstructNgNode<opset::Constant>(
-        op->name(), ng::element::u64, ng::Shape{output_shape.size()},
-        output_shape);
-    auto reshaped =
-        ConstructNgNode<opset::Reshape>(op->name(), slice, ng_shape, false);
-    SaveNgOp(ng_op_map, op->name(), reshaped);
+        op->name(), ng_input, ng_begin, ng_end, begin_mask, end_mask,
+        new_axis_mask, shrink_axis_mask);
```
kanvi-nervana, line 2524: For Unpack, is there no need to use Squeeze as was suggested?

adk9: No.

PR #725, opened 2020-12-03
Rename CI jobs/pipelines

adk9:
Rename CI jobs to account for different backend by adopting the following convention:

cpu (Basic sanity tests)
models-cpu (Model tests)
vpu
models-vpu
igpu
models-igpu
cpu-grappler
cpu-prebuilt
cpu-intel-tf


bani-intelaipg:
May be we keep a README in ci/buildkite folder with the convention info and some examples

adk9:
> May be we keep a README in ci/buildkite folder with the convention info and some examples

I think it's somewhat self-explanatory from the directory structure but I went ahead and added a README anyway.

bani-intelaipg (approved):

PR #727, opened 2020-12-04
Range op translation

kanvi-nervana:
Add translation and tests for Range op
The tests need to be run with CS enabled as Range Op with param input leads to a dynamic output not supported by CPU backend. The test although having constant inputs is converted to param. With CS=1, this does not happen.
Skip some TF py tests for the same reason.


ngraph_bridge/ngraph_builder.cc
```diff
@@ -2015,6 +2015,19 @@ static Status TranslatePadOp(const Node* op,
   return Status::OK();
 }
 
+static Status TranslateRangeOp(const Node* op,
+                               const std::vector<const Tensor*>&,
+                               Builder::OpMap& ng_op_map) {
+  ng::Output<ng::Node> ng_start, ng_stop, ng_step;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, ng_start, ng_stop, ng_step));
```
adk9, line 2022: start, stop and step should be static inputs.

kanvi-nervana: Why? Those values are not needed to be known before hand

ngraph_bridge/ngraph_mark_for_clustering.cc
```diff
@@ -535,6 +536,8 @@ const TypeConstraintMap& GetTypeConstraintMap() {
     type_constraint_map["PreventGradient"]["T"] = NGraphDTypes();
     type_constraint_map["Prod"]["T"] = NGraphNumericDTypes();
     type_constraint_map["Prod"]["Tidx"] = NGraphIndexDTypes();
+    type_constraint_map["Range"]["Tidx"] =
+        NGraphNumericDTypes();  //{DT_INT32, DT_INT64}; // IE expects int types
```
kanvi-nervana, line 540:
```suggestion
        NGraphIndexDTypes();  //{DT_INT32, DT_INT64}; // IE expects int types
```

ngraph_bridge/ngraph_builder.cc
kanvi-nervana: The inputs probably need to be static.

ngraph_bridge/ngraph_builder.cc
```diff
@@ -310,6 +310,47 @@ static Status GetStaticInputVector(
...
+    } break;
+    case DataType::DT_INT64: {
+      std::vector<int64> vec_i64;
+      TF_RETURN_IF_ERROR(
+          GetStaticInputVector(op, input_index, static_input_map, &vec_i64));
+      node_ = ConstructNgNode<opset::Constant>(op->name(), type, ng::Shape{},
+                                               vec_i64[0]);
+    } break;
+    default:
+      break;
```
kanvi-nervana, line 349: Need to add the error message here

adk9 (approved):

PR #732, opened 2020-12-04
add OV GPU pipeline to BuildKite CI

bani-intelaipg:


adk9:
@googlebot I consent.

test/ci/buildkite/igpu.yml
```diff
@@ -0,0 +1,89 @@
...
+      queue: ${BUILDKITE_AGENT_META_DATA_QUEUE}
+      name: ${BUILDKITE_AGENT_META_DATA_NAME}
+
+  - wait
+  - command: |
+      rm -rf /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID
+    label: ":wastebasket: Cleanup"
+    agents:
+      queue: ${BUILDKITE_AGENT_META_DATA_QUEUE}
+      name: ${BUILDKITE_AGENT_META_DATA_NAME}
```
adk9, line 89: Is this file needed if all the steps are now enabled? Please use cpu.yml with the plugin appropriately set.

test/ci/buildkite/run_pipeline.sh
```diff
@@ -22,6 +22,10 @@ elif [ "${BUILDKITE_PIPELINE_NAME}" == "cpu-intel-tf" ]; then
    export TF_LOCATION=/localdisk/buildkite-agent/prebuilt_intel_tensorflow
    export NGRAPH_TF_BACKEND=CPU
    PIPELINE_STEPS+=" ${SCRIPT_DIR}/cpu.yml "
+elif [ "${BUILDKITE_PIPELINE_NAME}" == "igpu" ]; then
+   export NGRAPH_TF_BACKEND=GPU
+   export TF_LOCATION=/localdisk/buildkite-agent/prebuilt_tensorflow_2_2_0
```
adk9, line 27: If `TF_LOCATION` is the same for all pipelines, can you please hoist it out of the conditional?

adk9 (changes requested):
Please rename the pipeline to gpu instead of igpu. Our convention is to use the OV plugin name for the pipeline.

Secondly, please add a pipeline to BK so that it shows up for this PR.

test/ci/buildkite/run_pipeline.sh
kanvi-nervana: It differs in case of Intel-TF.

test/python/tests_linux_gpu.txt
```diff
@@ -0,0 +1,42 @@
...
+# This manifest is consumed by nGraph-Bridge PyTest framework via .../test/python/conftest.py
+
+# Read comments in tests_common.txt for allowed regex/patterns
+
+###################################################
+[IMPORT]
+# Read in one/more external manifest file(s)
+# Path specified is relative to this file's path
+
+tests_linux_cpu.txt
```
kanvi-nervana, line 12: Shouldn't the import be for the common.txt?

test/tests_linux_gpu.txt
```diff
@@ -0,0 +1,23 @@
...
+#  Follows --gtest_filter patterns:
+#    * --> all 
+#    MathOps.* --> all tests under MathOps class
+
+###################################################
+[IMPORT]
+# Read in one/more external manifest file(s)
+# Path specified is relative to this file's path
+
+tests_linux_cpu.txt
```
kanvi-nervana, line 14: Same here

test/python/tensorflow/tests_linux_gpu.txt
```diff
@@ -0,0 +1,71 @@
...
+#    math_*_test --> module
+#    math_*_*_test --> module
+#    math*_test --> module
+
+###################################################
+[IMPORT]
+# Read in one/more external manifest file(s)
+# Path specified is relative to this file's path
+
+tests_linux_cpu.txt
```
kanvi-nervana, line 23: Shouldn't the import be for the common.txt?

test/ci/buildkite/igpu.yml
kanvi-nervana: If using cpu.yml for all pipelines then please change the name accordingly.

kanvi-nervana (changes requested):

test/ci/buildkite/run_pipeline.sh
bani-intelaipg: Hoisted all but intel-tf

test/python/tests_linux_gpu.txt
bani-intelaipg: Reusing rules as in CPU, and then adjust the extra skips

adk9: No, why? The same inclusions/exclusions as for CPU apply in the case of GPU.

test/ci/buildkite/run_pipeline.sh
adk9: Then it should only be set for the Intel TF case.

test/python/tests_linux_gpu.txt
adk9: Actually I agree with @kanvi-nervana. In both of these cases there are too many exclusions that it may be better to base this off of `common.txt`.

kanvi-nervana: If you import CPU.txt then it's exclusions will also be carried forward, right? Those exclusions might not be needed for GPU and you'll then explicitly put them in the RUN list which is not required.

bani-intelaipg: Correct. Can we enable more tests as part of a separate PR? As of this PR, we will have a baseline of equivalent to CPU minus few.

test/ci/buildkite/cpu.yml
```diff
@@ -13,6 +13,8 @@
 
   - wait
   - command: |
+      source /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/venv/bin/activate 
+      LD_LIBRARY_PATH=${TF_LOCATION}/artifacts/tensorflow:/localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/lib \
```
adk9, line 17: This shouldn't be needed.

test/python/tests_linux_gpu.txt
kanvi-nervana: It's up to you, but we are in agreement that the base for every pipeline should be common.txt

adk9: More tests can be enabled as a part of a separate PR (since it's a new change) but if we agree that GPU tests shouldn't depend on the CPU tests then that change should be made as a part of this PR.

I understand that the CPU pipeline is the most permissive but at the risk of being slightly repetitive, I think it's generally a better idea to keep them separate from the outset?

test/ci/buildkite/cpu.yml
adk9: Will be removed in a follow-up PR.

test/python/tensorflow/tests_linux_gpu.txt
adk9: Will be fixed in a follow-up PR.

test/tests_linux_gpu.txt
adk9: Will be fixed in a follow-up PR.

adk9 (approved):

PR #733, opened 2020-12-06
Update DepthwiseConv2DNative translation to use GroupConvolution

adk9:


ngraph_bridge/ngraph_builder.cc
```diff
@@ -1065,55 +1063,26 @@ static Status TranslateDepthwiseConv2dNativeOp(
...
-        op->name(), ng_sliced_input, ng_sliced_filter, ng_strides,
-        ng_padding_below, ng_padding_above, ng_dilations);
-
-    ng_args.push_back(ng_conv);
-  }
-
-  size_t ng_concatenation_axis = 1;  // channel axis
-  auto ng_concat = ConstructNgNode<opset::Concat>(op->name(), ng_args,
-                                                  ng_concatenation_axis);
+  // H W I M -> H W I 1 M
```
kanvi-nervana, line 1066: Is there an existing test, if not can you please add a test?

adk9: Yes, there are tests in `test/python/test_depthwiseconv2d.py` for both single and multiple groups that I used to debug the translation. Besides, few of the models enabled in the `models-cpu` pipeline also cover the translation now.

kanvi-nervana (approved):

PR #735, opened 2020-12-06
Add translation for Where op

adk9:
Add translation for [tf.where](https://www.tensorflow.org/api_docs/python/tf/where).


ngraph_bridge/ngraph_builder.cc
```diff
@@ -2675,6 +2675,7 @@ const static std::map<
         {"TopKV2", TranslateTopKV2Op},
         {"Transpose", TranslateTransposeOp},
         {"Unpack", TranslateUnpackOp},
+        {"Where", TranslateSelectOp},
```
kanvi-nervana, line 2678: Please add a test for where op or enable TF test.

adk9: I assumed that `tests/python/test_select.py` is actually testing for Where since it uses `tf.where`. Let me check again and add explicit Where tests if that's not the case.

adk9: Added and enabled tests for where op.

test/python/test_select.py
```diff
@@ -104,3 +104,58 @@ def run_test(sess):
 
         assert (
             self.with_ngraph(run_test) == self.without_ngraph(run_test)).all()
+
+
+class TestWhere(NgraphTest):
+    env_map = None
+
+    def __init__(self):
+        self.store_env_variables(['TF_OV_CONSTANT_FOLDING'])
```
bani-intelaipg, line 113: env_map = self.store_env_variables(['TF_OV_CONSTANT_FOLDING'])

test/python/test_select.py
```diff
@@ -104,3 +104,58 @@ def run_test(sess):
...
+        self.set_env_variable('TF_OV_CONSTANT_FOLDING', '1')
+
+    def __del__(self):
+        self.restore_env_variables(env_map)
+
+    def test_where(self):
+        a = np.random.random(size=[2, 2]).astype(np.float32)
```
bani-intelaipg, line 120: May be better to not randomize the test data, to ensure there is a predictable value for the number of '3'

test/python/test_select.py
adk9: Fixed in #742

test/python/test_select.py
adk9: Fixed in #742

PR #741, opened 2020-12-09
update CI test manifests for OV GPU plugin

bani-intelaipg:
* Update GPU manifests to be independent from CPU manifests
* In build_tf.py make intel-mkl-tf as NOT default


build_tf.py
```diff
@@ -71,7 +71,7 @@ def main():
         os.chdir(pwd)
 
     use_intel_tf = False
-    if arguments.use_intel_tensorflow != '':
+    if arguments.use_intel_tensorflow:
         use_intel_tf = True
```
adk9, line 75: This doesn't change the default. Please change the default to false.

examples/cpp/infer_single_network.cc
```diff
@@ -59,15 +59,6 @@ void PrintAvailableBackends() {
...
 
-// Sets the specified backend. This backend must be set BEFORE running
-// the computation
-tf::Status SetNGraphBackend(const string& backend_name) {
-  // Select a backend
-  tf::Status status =
-      tf::ngraph_bridge::BackendManager::SetBackend(backend_name);
-  return status;
-}
-
```
adk9, line 70: Why do you have these changes again?

adk9 (changes requested):

build_tf.py
bani-intelaipg: The action="strore_true" will set it as True only if the arg is specified

examples/cpp/infer_single_network.cc
bani-intelaipg: merge issues, resolved.

bani-intelaipg:
@googlebot I consent

adk9:
@googlebot I consent

adk9 (approved):
LGTM

PR #751, opened 2020-12-15
set plugin/device for run_infer script used in CI

bani-intelaipg:


adk9 (changes requested):
LGTM. Please comment out the failing models in `models-myriad` pipeline.

adk9 (approved):

test/ci/buildkite/models-cpu.yml
```diff
@@ -16,9 +16,12 @@
   - wait
   - command: |
       source /localdisk/buildkite/artifacts/$BUILDKITE_BUILD_ID/venv/bin/activate
+      device=${NGRAPH_TF_BACKEND:-"CPU"}
+      device=$(echo "\$device" | tr '[:upper:]' '[:lower:]')
+      MODELFILENAME=models_\${device}.txt
```
adk9, line 21: Not needed since run_infer_multiple.sh knows how to use the correct device manifest?

bani-intelaipg: In local dev/testing, we might want to pass a custom manifest to the run_infer_multiple.sh for targeted testing - so need a way to specify the param

adk9: Sure, but the correct one will be used if one isn't specified? Why do you need to explictly specify a manifest here?

adk9: I'm marking this as resolved in the interest of merging this PR ASAP to enable testing of other outstanding PRs.

Please remove the explicit setting of model manifest above in a follow-up PR.

PR #752, opened 2020-12-15
Fix multi-output handling bug in TS

adk9:
This fixes an issue in transpose sinking where transposes where being
combined for ops with multiple outputs. `ngraph::Output<ngraph::Node>` is
used to track edges instead of using `ngraph::Node`.


ngraph_bridge/pass/transpose_sinking.cc
```diff
@@ -79,38 +78,41 @@ static string describe(shared_ptr<ngraph::Node> node) {
...
 
 static shared_ptr<opset::Transpose> read_transposemap(
     TransposeMap& reorders, ngraph::Output<ngraph::Node> target) {
-  auto transpose = reorders.at(target.get_node_shared_ptr());
-  NGRAPH_VLOG(4) << "Read TransposeMap[" << target.get_node()->get_name()
-                 << "]  -> "
-                 << describe<opset::Transpose>(transpose[target.get_index()]);
-  return transpose[target.get_index()];
+  auto name =
+      target.get_node()->get_name() + "." + to_string(target.get_index());
```
kanvi-nervana, line 111: This would be an issue if the same output from the node goes to two different nodes

adk9: Not sure I understand. Can you elaborate?

kanvi-nervana: Node n 
                              /\
                             /. \
                   Dst x     Dst y
In this case the key in the map for output x would be n.0 and for y would also be n.0, right?

adk9: No? The index is considered when generating the name.

kanvi-nervana: Yes, and name is the key so when writing in the map this could be a problem

adk9: Ok, so `n` in the example above is not a multi-output op but a single-output op whose inputs are fed to more than one user?

kanvi-nervana: Yes, n is a single output op whose output is fed to more than 1 user

kanvi-nervana: Based on offline discussion, looks ok for now. We can look for a corner case where this might be an issue.

kanvi-nervana (approved):

PR #758, opened 2020-12-16
Use env var to set min non-trivial nodes

kanvi-nervana:
Add support to use the environment variable `TF_OV_MIN_NONTRIVIAL_NODES` to set the cluster threshold for an encapsulated nGraph cluster. A cluster is busted if the number of non-trivial nodes (excluding Const and Identity) exceed the specified minimum number of non-trivial nodes.

### DIFF
--- a/ngraph_bridge/ngraph_deassign_clusters.cc
+++ b/ngraph_bridge/ngraph_deassign_clusters.cc
@@ -190,9 +190,9 @@ Status DeassignClusters(Graph* graph) {
     int min_non_trivial_nodes = 6;
     if (std::getenv("TF_OV_MIN_NONTRIVIAL_NODES") != nullptr) {
       min_non_trivial_nodes =
-          std::atof(std::getenv("TF_OV_MIN_NONTRIVIAL_NODES"));
+          std::stoi(std::getenv("TF_OV_MIN_NONTRIVIAL_NODES"));
     }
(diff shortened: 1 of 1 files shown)


ngraph_bridge/ngraph_deassign_clusters.cc
```diff
@@ -189,7 +187,14 @@ Status DeassignClusters(Graph* graph) {
       }
     }
 
-    if (non_trivial_count < MIN_NONTRIVIAL_NODES) {
+    int min_non_trivial_nodes = 6;
+    if (std::getenv("TF_OV_MIN_NONTRIVIAL_NODES") != nullptr) {
+      min_non_trivial_nodes =
+          std::atof(std::getenv("TF_OV_MIN_NONTRIVIAL_NODES"));
```
adk9, line 193:
```suggestion
          std::stoi(std::getenv("TF_OV_MIN_NONTRIVIAL_NODES"));
```

ngraph_bridge/ngraph_deassign_clusters.cc
```diff
@@ -189,7 +187,14 @@ Status DeassignClusters(Graph* graph) {
...
+    }
+    NGRAPH_VLOG(0) << "MIN_NONTRIVIAL_NODES set to " << min_non_trivial_nodes;
```
adk9, line 195: Maybe VLOG(1) since this will be printed everytime?

adk9 (approved):

PR #762, opened 2020-12-17
Cavusmustafa/modular backend review v6

cavusmustafa:


ngraph_bridge/ie_basic_engine.cc
```diff
@@ -0,0 +1,79 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+#include "ngraph_bridge/ie_basic_engine.h"
+#include <iostream>
+#include "logging/ngraph_log.h"
+#include "ngraph_bridge/ie_utils.h"
```
kanvi-nervana, line 20: General guideline followed in the bridge
`standard includes`
`OV`
`nGraph`
`bridge`

ngraph_bridge/executable.cc
```diff
@@ -213,22 +225,29 @@ bool Executable::call(const vector<shared_ptr<runtime::Tensor>>& inputs,
...
-      m_infer_req.SetBlob(get_output_name(results[i]), tv->get_blob());
+      ie_outputs[i] = static_pointer_cast<IETensor>(outputs[i]);
     }
+    output_names[i] = get_output_name(results[i]);
+  }
+
+  if (multi_req_execution) {
+    m_ie_engine->enable_multi_req_execution();
+  } else {
+    m_ie_engine->disable_multi_req_execution();
```
kanvi-nervana, line 240: Is this needed, isn't it disabled by default?

kanvi-nervana (changes requested):
Requesting some cosmetic changes
Please remove commented code, use CamelCase for function names

ngraph_bridge/executable.cc
cavusmustafa: It's added in case batching might be disabled during consecutive executions. But we don't have advanced checks for now and there is no scenario to disable batching if it's already enabled. I updated it accordingly.

cavusmustafa:
> Requesting some cosmetic changes
> Please remove commented code, use CamelCase for function names

I'm confused about using CamelCase. Function names does not use it in bridge now. Is there a plan to update them all?

kanvi-nervana:
> > Requesting some cosmetic changes
> > Please remove commented code, use CamelCase for function names
> 
> I'm confused about using CamelCase. Some function names does not use it in bridge now. Is there a plan to update them all?

Yes, we will update the function names that don't follow it

adk9 (changes requested):
Thanks for the revised PR. I have a few high-level comments:

1. Please break this PR into individual self-contained PRs. The changes here not only introduce a new backend infrastructure (with possibly redundant interfaces to those that currently exist) but also add a new HDDL/VADM backend. I'd propose that we add a modular backend first, followed by a separate PR to add a new HDDL/VADM backend with the tests enabled for it.

2. Secondly, please take a look at the TensorFlow [Contribution guidelines and standards](https://github.com/tensorflow/tensorflow/blob/master/CONTRIBUTING.md#contribution-guidelines-and-standards) for the general contribution guidelines that we use. Please use a short explanatory title for the PR and add a description to your PR if the title is not self-explanatory. The guide also points to the Google C++ style guide that's being used for the code in here.

PR #767, opened 2020-12-17
Remove Const->Result pattern check

adk9:


ngraph_bridge/ngraph_assign_clusters.cc
```diff
@@ -554,37 +554,6 @@ Status AssignClusters(Graph* graph) {
 
   NGRAPH_VLOG(2) << "Contraction done";
 
-  // Remove Const with outputs to another cluster from being clustered
```
kanvi-nervana, line 557: Can OV now handle this?

adk9: Yes.

kanvi-nervana (approved):

PR #769, opened 2020-12-17
Update Readme

kanvi-nervana:


README.md
```diff
@@ -151,39 +149,9 @@ See [this README](/tools) if you want to use Docker.
...
-    sess = tf.compat.v1.Session(config=config_ngraph_enabled)
-
-Run the classification:
-
-    python label_image.py --graph inception_v3_2016_08_28_frozen.pb \
-        --image grace_hopper.jpg --input_layer=input \
-        --output_layer=InceptionV3/Predictions/Reshape_1 \
-        --input_height=299 --input_width=299 \
-        --labels imagenet_slim_labels.txt 
+    python infer_image.py
```
adk9, line 154:
```suggestion
    python examples/infer_image.py
```

README.md
```diff
@@ -151,39 +149,9 @@ See [this README](/tools) if you want to use Docker.
...
```
adk9, line 155: Maybe add a note here saying that this will download the Inception v3 model and a sample image before printing the results.

One thing that'd be nice to call out is that the only change required to the sample script is
```
import ngraph_bridge
```

adk9 (approved):

PR #774, opened 2021-01-06
Set KEY_DUMP_EXEC_GRAPH_AS_DOT only for CPU plugin

adk9:
Set plugin config KEY_DUMP_EXEC_GRAPH_AS_DOT only for CPU plugin as it is not supported by other plugins.

### DIFF
--- a/ngraph_bridge/executable.cc
+++ b/ngraph_bridge/executable.cc
@@ -135,8 +135,10 @@ Executable::Executable(shared_ptr<Function> func, string device)
     auto& name = m_function->get_friendly_name();
     m_network.serialize(name + ".xml", name + ".bin");
     util::DumpNGGraph(func, name + "_executable");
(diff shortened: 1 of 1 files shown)


ngraph_bridge/executable.cc
```diff
@@ -135,8 +135,10 @@ Executable::Executable(shared_ptr<Function> func, string device)
     auto& name = m_function->get_friendly_name();
     m_network.serialize(name + ".xml", name + ".bin");
     util::DumpNGGraph(func, name + "_executable");
-    options[InferenceEngine::PluginConfigParams::KEY_DUMP_EXEC_GRAPH_AS_DOT] =
-        name + "_IE_" + m_device;
+    if (m_device == "CPU") {
```
kanvi-nervana, line 138: Do the other plugins not support this?

adk9: That's correct. Each plugin has a different way to dump graphs if that.

kanvi-nervana (approved):

PR #777, opened 2021-01-08
Enable performance benchmarking CI for CPU plugin

bani-intelaipg:
See CI pipeline: https://buildkite.com/ngraph/bench-cpu


test/ci/buildkite/models-cpu.yml
```diff
@@ -23,7 +23,7 @@
       cd /localdisk/buildkite/artifacts && rm -rf ./$BUILDKITE_BUILD_ID
       if (( \$retcode > 0 )); then echo "Some model testing failed!"; exit 1; fi
     label: ":bar_chart: Model Inference"
-    timeout_in_minutes: 30
+    timeout_in_minutes: 60
```
bani-intelaipg, line 26: Sometimes if we re-download models the GPU pipeline needs more time

test/models/tf_ov_model_zoo/run_infer_single.sh
```diff
@@ -19,11 +19,17 @@ INFER_PATTERN=$3
...
+if [ "${BENCHMARK}" == "" ]; then
+    echo "Error: benchmark flag (YES/NO) not specified!" && exit 1
+fi
 
-echo MODEL=$MODEL IMAGE=$IMAGE INFER_PATTERN=$INFER_PATTERN
+echo MODEL=$MODEL IMAGE=$IMAGE INFER_PATTERN=$INFER_PATTERN BENCHMARK=$BENCHMARK
 
 REPO=https://gitlab.devtools.intel.com/mcavus/tensorflow_openvino_models_public
-COMMIT=14eff8ce # 2020-Dec-02
+COMMIT=d12f2d57 # 2021-Jan-06
```
adk9, line 30: Is this necessary?

test/models/tf_ov_model_zoo/run_infer_single.sh
```diff
@@ -57,83 +77,174 @@ function get_model_repo {
...
+
+    cd ${LOCALSTORE} || exit 1
+    commit_check=$(git cat-file -t $COMMIT 2>/dev/null)
+    [ "$commit_check" == "commit" ] || ( git fetch; git checkout ${COMMIT} || exit 1; )
+    
+    if [ -d "temp_build" ]; then rm -rf temp_build; fi
+
+    if [ ! -f "${LOCALSTORE}/frozen/${MODEL}.pb" ] || \
+    [ ! -f "${LOCALSTORE}/frozen/${MODEL}.txt" ] || \
+    [ "$(file_newer_than_commit ${LOCALSTORE}/frozen/${MODEL}.pb)" == "0" ]; then
```
adk9, line 90: Wouldn't this always be true? Also, it's checking the timestamp of the file with the timestamp of the latest commit of the repo?

test/models/tf_ov_model_zoo/run_infer_single.sh
```diff
@@ -57,83 +77,174 @@ function get_model_repo {
...
 if [ ! -f "${IMGFILE}" ]; then echo "Cannot find image ${IMGFILE} !"; exit 1; fi
-cd ${LOCALSTORE}/demo
-NUM_ITER=20
-[ -z "$NGRAPH_TF_LOG_PLACEMENT" ] && export NGRAPH_TF_LOG_PLACEMENT=1
-[ -z "$NGRAPH_TF_VLOG_LEVEL" ] && export NGRAPH_TF_VLOG_LEVEL=-1
 device=${NGRAPH_TF_BACKEND:-"CPU"}
+
+if [ "${BENCHMARK}" == "YES" ]; then
+    NUM_ITER=150
+    WARMUP_ITERS=50
```
adk9, line 209: 100 and 5 is enough here.

test/models/tf_ov_model_zoo/run_infer_single.sh
```diff
@@ -57,83 +77,174 @@ function get_model_repo {
...
+    export NGRAPH_TF_VLOG_LEVEL=-1
+else
+    NUM_ITER=20
+    WARMUP_ITERS=1
```
adk9, line 213: 1 and 0 may be enough here since we simply want to check if it works?

test/models/tf_ov_model_zoo/run_infer_single.sh
```diff
@@ -57,83 +77,174 @@ function get_model_repo {
...
 ################################################################################
 
-pip list | grep 'Pillow' 2>&1 >/dev/null; found=$(( ! $? ));
-if (( ! $found )); then pip install Pillow; fi
+pip list | grep 'Pillow' 2>&1 >/dev/null; (($?==0)) || pip install Pillow;
+if [ "${BENCHMARK}" == "YES" ]; then
+    pip list | grep 'networkx' 2>&1 >/dev/null; (($?==0)) || pip install networkx;
+    pip list | grep 'defusedxml' 2>&1 >/dev/null; (($?==0)) || pip install defusedxml;
+    pip list | grep 'test-generator' 2>&1 >/dev/null; (($?==0)) || pip install test-generator==0.1.1;
+fi
```
adk9, line 198: Are these OV dependencies? I think that it's better to explicitly install them up-front in the manifest.

Also, please use something like `grep -E "^pkg-name "` for a proper match.

test/models/tf_ov_model_zoo/run_infer_single.sh
adk9: Actually, it's better to fix this than use the latest. Never mind my question above.

bani-intelaipg: Yes, we don't want a moving baseline from other sources, as it might break our CI. With known baseline, we know it works.

test/models/tf_ov_model_zoo/run_infer_single.sh
bani-intelaipg: Will update.

test/models/tf_ov_model_zoo/run_infer_single.sh
bani-intelaipg: This way, if we update the model-zoo commit-id, it will re-create all models in CI agents. This might be needed, as the zoo-scripts sometimes create additional config-files.

test/models/tf_ov_model_zoo/run_infer_single.sh
bani-intelaipg: Not sure, how to install in manifest. Please suggest. Will update the grep pattern.

test/models/tf_ov_model_zoo/run_infer_single.sh
bani-intelaipg: This code is also used for non-benchmark pipeline, where we print average speed. That way we can check for regression on the average inference speed.

test/models/tf_ov_model_zoo/run_infer_single.sh
adk9: In that case, we should check if the commit-id declared is newer than the commit-id of the HEAD? The timestamp of a file is a local attribute and it corresponds to the time the file was checked out/cloned and it has nothing to do with the git timestamp of that file or HEAD.

test/models/tf_ov_model_zoo/run_infer_single.sh
adk9: The non-benchmark CI pipeline do not run exclusively and we shouldn't rely on them for average inference speed.

test/models/tf_ov_model_zoo/run_infer_single.sh
adk9: I meant that the install should be done in the "build" step of the `bench-cpu.yml` file. But it's not a big deal to have it here too as a sanity check. Your call.

test/models/tf_ov_model_zoo/run_infer_single.sh
bani-intelaipg: Updated

test/models/tf_ov_model_zoo/run_infer_single.sh
```diff
@@ -45,8 +52,21 @@ function gen_frozen_models {
...
+    filepath=$1
+    [ -e "$filepath" ] || ( echo "0"; return; )
+    file_ts=$(stat -c %Y $filepath)
+    # must be already under the git repo
+    commit_ts=$(git show -s --format=%at $COMMIT 2>/dev/null)
+    (( $? == 0 )) || ( echo "0"; return; )
+    (( file_ts > commit_ts )) && ( echo "1"; return; )
+    echo "0"
+}
+
```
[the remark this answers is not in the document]
bani-intelaipg: Thanks. Removed.

test/models/tf_ov_model_zoo/run_infer_single.sh
bani-intelaipg: Once kicked off we should monitor few runs and check the performacne.

adk9 (approved):

PR #779, opened 2021-01-09
Enable py tests for VPU pipeline

kanvi-nervana:


adk9 (approved):
How many tests are now enabled for the VPU pipeline?

kanvi-nervana:
> How many tests are now enabled for the VPU pipeline?

This PR only adds more bridge py tests, about 57 of those.

PR #783, opened 2021-01-15
build - make cxx_abi=0 default

kanvi-nervana:
Build options: Build with CXX_ABI=0 by default

### DIFF
--- a/build_ngtf.py
+++ b/build_ngtf.py
@@ -232,8 +232,9 @@ def main():
     # this is set to 1
     # The specific value is determined from the TensorFlow build
     # Normally the shipped TensorFlow going forward is built with gcc 7.3
-    # and thus this flag is set to 1
-    cxx_abi = "1"
+    # and thus this flag should be set to 1
+    # But, tensorflow still uses cxx_abi=0 for backward compatibility
+    cxx_abi = "0"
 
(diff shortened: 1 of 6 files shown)


adk9 (changes requested):
Why can't one use prebuilt-tensorflow with OV built with _GLIBCXX_USE_CXX11_ABI=0? Perhaps we should default to building everything with _GLIBCXX_USE_CXX11_ABI=0?

kanvi-nervana:
> Why can't one use prebuilt-tensorflow with OV built with _GLIBCXX_USE_CXX11_ABI=0? Perhaps we should default to building everything with _GLIBCXX_USE_CXX11_ABI=0?

You could, you would need to make sure that when you use these 2 options together then you provide the right location of OV built with ABI=0
Fixing the build to use ABI=0 always is a better option.

adk9 (approved):

build_ngtf.py
```diff
@@ -229,8 +228,9 @@ def main():
     # this is set to 1
```
bani-intelaipg, line 228: This comment should be updated?

bani-intelaipg (approved):

build_ngtf.py
kanvi-nervana: It is comment in general

PR #793, opened 2021-01-27
Use --single-branch when cloning git repos

adk9:


tools/build_utils.py
```diff
@@ -511,7 +511,10 @@ def install_ngraph_tf(tf_version, venv_dir, ngtf_pip_whl):
 
 def download_repo(target_name, repo, version, submodule_update=False):
     # First download to a temp folder
-    call(["git", "clone", repo, target_name])
+    call([
```
bani-intelaipg, line 514: This might not work where "version" is passed as a commit# or label

adk9: Good point. I'll close it for now.

PR #799, opened 2021-02-01
Remove use-grappler-optimizer compile option

adk9:
* Unconditionally enable "ngraph-optimizer" grappler pass
* Remove is_grappler_enabled() API call
* Run tests using the grappler pass

### DIFF
--- a/test/graph_rewrites/config_for_grappler_test.cc
+++ b/test/graph_rewrites/config_for_grappler_test.cc
@@ -41,6 +41,7 @@ namespace testing {
 // etc.,etc.
 
 TEST(GrapplerConfig, RConfig1) {
+  ActivateNGraph();
   // Create Graph
   Scope root = Scope::NewRootScope();
   auto A = ops::Const(root.WithOpName("A"), {3.f, 2.f});
@@ -72,7 +73,6 @@ TEST(GrapplerConfig, RConfig1) {
   // Run grappler
   tensorflow::grappler::MetaOptimizer optimizer(nullptr, config_proto);
   GraphDef output;
-  // const Status status = optimizer.Optimize(nullptr, item, &output);
   ASSERT_OK(optimizer.Optimize(nullptr, item, &output));
 
   // GraphDef to Graph
@@ -91,9 +91,11 @@ TEST(GrapplerConfig, RConfig1) {
     ng_encap = node;
   }
   ASSERT_NE(ng_encap, nullptr);
+  DeactivateNGraph();
 }
 
 TEST(GrapplerConfig, RConfig4) {
+  ActivateNGraph();
   // Create Graph
   Scope root = Scope::NewRootScope();
   auto A = ops::Const(root.WithOpName("A"), {3.f, 2.f});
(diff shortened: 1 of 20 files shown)


test/graph_rewrites/config_for_grappler_test.cc
```diff
@@ -41,6 +41,7 @@ namespace testing {
 // etc.,etc.
 
 TEST(GrapplerConfig, RConfig1) {
+  ActivateNGraph();
```
kanvi-nervana, line 44: Why is the explicit enabling/disabling of nGraph required?

test/python/common.py
```diff
@@ -53,9 +53,8 @@ def with_ngraph(self, l, config=None):
         # because mutable objects should not be used as defaults in python
         if config is None:
             config = tf.compat.v1.ConfigProto()
-        # TODO: Stop grappler on failure (Add fail_on_optimizer_errors=True)
-        config = ngraph_bridge.update_config(config)
 
+        # config = ngraph_bridge.update_config(config)
```
kanvi-nervana, line 57: Disabling the config update; tests don't run with the optimizer?

test/test_utilities.cpp
```diff
@@ -344,20 +344,18 @@ tf::SessionOptions GetSessionOptions() {
...
-    custom_config->set_name("ngraph-optimizer");
-    options.config.mutable_graph_options()
-        ->mutable_rewrite_options()
-        ->set_min_graph_nodes(-1);
-
-    options.config.mutable_graph_options()
-        ->mutable_rewrite_options()
-        ->set_meta_optimizer_iterations(tf::RewriterConfig::ONE);
-  }
+  auto* custom_config = options.config.mutable_graph_options()
```
kanvi-nervana, line 347: Won't the CPP tests now run with the optimizer, all the time?

tools/test_utils.py
```diff
@@ -167,9 +167,7 @@ def run_tensorflow_pytests_from_artifacts(ngraph_tf_src_dir, tf_src_dir,
 
     # Check to see if we need to apply the patch for Grappler
     import ngraph_bridge
-    patch_file_name = "test/python/tensorflow/tf_unittest_ngraph" + (
-        "_with_grappler"
-        if ngraph_bridge.is_grappler_enabled() else "") + ".patch"
+    patch_file_name = "test/python/tensorflow/tf_unittest_ngraph.patch"
```
kanvi-nervana, line 170: The TF py tests will also run only with the graph pass?

kanvi-nervana (changes requested):

tools/test_utils.py
adk9: Yes, bridge C++ unit tests run with the optimizer, bridge Python and TF Python unit tests run without the optimizer. It seems a little excessive to test all the combinations here since the optimizer and non-optimizer paths are shared. Which configurations would we prefer testing?

test/test_utilities.cpp
adk9: See below.

test/python/common.py
adk9: See below.

test/graph_rewrites/config_for_grappler_test.cc
adk9: It's to disable de-assigning of clusters.

tools/test_utils.py
kanvi-nervana: Which one is going to be the default/promoted more than the other, we should make sure all tests pass with that. If there isn't any then probably the above configuration is fine.

test/graph_rewrites/config_for_grappler_test.cc
kanvi-nervana: You can use `NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS` ?

adk9: Where? To run the entire unit test suite? I'm not sure why this isn't being done already.

tools/test_utils.py
adk9: The current default is non-optimizer. We could use this for all three but that'd leave the optimizer path untested.

PR #800, opened 2021-02-01
Add py test for Range op with constant inputs

kanvi-nervana:


test/python/test_range.py
```diff
@@ -0,0 +1,44 @@
+# ==============================================================================
+#  Copyright 2018-2020 Intel Corporation
```
adk9, line 2:
```suggestion
#  Copyright 2021 Intel Corporation
```

adk9 (approved):

